PR #270, opened 2014-12-24
Refactor Client

nateberkopec:
+ remove redundant excluded environment check
+ cleanup methods, esp. send
+ remove redundant private
+ remove unnecessary return statements, variable assignments

### DIFF
--- a/lib/raven/client.rb
+++ b/lib/raven/client.rb
@@ -7,9 +7,8 @@ require 'raven/transports/http'
 require 'raven/transports/udp'
 
 module Raven
-
+  # Encodes events and sends them to the Sentry server.
   class Client
-
     PROTOCOL_VERSION = '5'
     USER_AGENT = "raven-ruby/#{Raven::VERSION}"
     CONTENT_TYPE = 'application/json'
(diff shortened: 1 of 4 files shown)


lib/raven/client.rb
```diff
@@ -23,14 +22,6 @@ def initialize(configuration)
...
 
     def send(event)
-      unless configuration.send_in_current_environment?
-        configuration.log_excluded_environment_message
-        return
-      end
-
-      # Set the project ID correctly
-      event.project = self.configuration.project_id
-
```
nateberkopec: I couldve sworn I left a comment on this!

* [setting the project ID is now the responsibility of the Event](https://github.com/getsentry/raven-ruby/pull/270/files#diff-8338fea26aee624929fed65e2d301325R44)
* Excluding due to the environment is already happening in Raven::Base, although after I saw how people use Raven::Client.send in background jobs, I think I may need to reconsider this.

nateberkopec:
Renaming should_send to should_capture is a fairly major API change. We may need to get serious about deprecation warnings. I know it's 0.X and all but I dunno.

nateberkopec:
Deciding I'm going to roll with it for now. I'm going to try to keep patchlevel releases non-breaking, but minors may break things for as long as we're at 0.x.

PR #271, opened 2014-12-24
Use an openstruct for a config object

nateberkopec:
Not mergeable - just an idea for now. Will need to replace all the docs I destroyed with this.


nateberkopec:
Yeah, I think you're right. Interesting idea but [the big boys do it with attr_accessor](https://github.com/rails/rails/blob/0082f02d1efcfd65b4c7f0f80db4356f82271d70/railties/lib/rails/application/configuration.rb), so that's good enough for us.

PR #275, opened 2015-01-05
Adding suppport for sanitizing query parameters.

pcorliss:
Paired with @kjperry

For GET requests and POST requests with query parameters the query_string remains unsanitized. While GET requests with sensitive data is rare for our application we wanted to be sure that bad requests would be sanitized as expected.

screen shot 2015-01-05 at 9 21 25 am

### DIFF
--- a/lib/raven/processor/sanitizedata.rb
+++ b/lib/raven/processor/sanitizedata.rb
@@ -14,6 +14,8 @@ module Raven
         process(v)
       elsif v.is_a?(Array)
         v.map{|a| sanitize(nil, a)}
+      elsif k == 'query_string'
+        sanitize_query_string(v)
       elsif v.is_a?(String) && (json = parse_json_or_nil(v))
         #if this string is actually a json obj, convert and sanitize
         json.is_a?(Hash) ? process(json).to_json : v
@@ -27,6 +29,13 @@ module Raven
     end
 
     private
+
+    def sanitize_query_string(query_string)
+      query_string.split('&').map do |key_val_pairs|
+        key, val = key_val_pairs.split('=')
+        "#{key}=#{sanitize(key, val)}"
+      end.join('&')
+    end
 
     def fields_re
       @fields_re ||= /(#{(DEFAULT_FIELDS + @sanitize_fields).join("|")})/i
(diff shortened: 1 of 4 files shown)


lib/raven/processor/sanitizedata.rb
```diff
@@ -28,6 +30,13 @@ def sanitize(k,v)
 
     private
 
+    def sanitize_query_string(query_string)
+      query_string.split('&').map do |key_val_pairs|
```
nateberkopec: Can you kindly use `CGI::parse` here? I think the code as written will break with keys that have several values + url encoding.

pcorliss:
Modified to use standard library functions. Please note that I had to modify the array sanitizer to pass the key down. Otherwise elements in arrays wouldn't be sanitized as desired.

pcorliss:
Phew, okay now that the builds are passing...

I added in a backport of URI from the rack gem to support `URI::encode_www_form_component` which allows us to convert back from `CGI::parse`. This solution involves a fair amount of extra stuff, but does result in an encoding safe sanitization.

If the above changes seem excessive we could always drop back to my original solution. I believe it will handle odd encodings and several values but I'm not 100% certain. I attempted several odd encodings and special character combinations and it seemed to work fine.

nateberkopec:
Wow! That's some awesome work! Thank you so much for taking the time to do all of this - I was considering just doing a "if >1.8 do" sort of branch, but you've made it work. I'm sure the people running 1.8 will appreciate it.

@pcorliss Can you write a test for the case of multiple values to a key and for url encoding? Just two more test assertions is all, nothing crazy. I know both of these cases are common in Rails so I want make sure we get that right.

pcorliss:
Tests added. A few notes, because of the way this is implemented the query string can mutate a bit. This is due to the decoding and subsequent re-encoding of the content of the strings. Simple stuff like `%20` being converted to `+`, and `[` converting to `%5B`. This is unavoidable as far as I can tell. And shouldn't overtly affect clients.

nateberkopec:
@pcorliss thanks so much again. This is some very classy work.

I think the gain here (sanitizing query strings) is better than the loss (query strings which were not url encoded in the original request will appear url encoded in Sentry). Ideally you will always have your actual application logs to check against any discrepancy.

PR #277, opened 2015-01-06
Add proxy support

dcramer:


dcramer:
fwiw I spent 30 seconds on this and just implemented what I assumed would work based on Google:

http://www.intridea.com/blog/2012/3/12/faraday-one-http-client-to-rule-them-all

PR #278, opened 2015-01-07
Truncate project_root prefix from filename

eagletmt:
Especially in Rails project, when an exception is raised in view template, the filename is always full path.
With capistrano (de-facto standard for Ruby application deployment), the full path differs on each release.
It causes Sentry to recognize the same error differently.

### DIFF
--- a/lib/raven/interfaces/stack_trace.rb
+++ b/lib/raven/interfaces/stack_trace.rb
@@ -36,6 +36,12 @@ module Raven
         return nil if self.abs_path.nil?
 
         prefix = $LOAD_PATH.select { |s| self.abs_path.start_with?(s.to_s) }.sort_by { |s| s.to_s.length }.last
+        if prefix.nil? && Raven.configuration.project_root
+          project_root = Raven.configuration.project_root.to_s
+          if self.abs_path.start_with?(project_root)
+            prefix = project_root
+          end
+        end
         prefix ? self.abs_path[prefix.to_s.chomp(File::SEPARATOR).length+1..-1] : self.abs_path
       end
 
--- a/spec/raven/event_spec.rb
+++ b/spec/raven/event_spec.rb
@@ -401,11 +401,11 @@ describe Raven::Event do
             expect(frames[1][:in_app]).to eq(false)
             expect(frames[2][:filename]).to eq("/gem/lib/path")
             expect(frames[2][:in_app]).to eq(false)
-            expect(frames[3][:filename]).to eq("/rails/root/app/foobar")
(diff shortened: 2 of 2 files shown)


seanlinsley:
Hmm, my Rails application uses Capistrano and doesn't have an issue with Sentry failing to group events.

eagletmt:
Hmm... my Rails application always says the filename is `/path/to/deploy/directory/releases/$TIMESTAMP/app/views/foo/bar.html.erb` .

Even if the filename was `/path/to/deploy/directory/current`, truncating meaningless part would be better.

seanlinsley:
I'm using Rails 4.1 with Capistrano 2.15.5

eagletmt:
Again, even if the filename was `/path/to/deploy/directory/current/app/views/foo/bar.html.erb` , truncating meaningless part would be better.
`app/views/foo/bar.html.erb` is a familiar form to Rails users.

Model's or controller's filename is `book.rb` or `books_controller.rb` in Rails project.
Only view templates have full path currently.

nateberkopec:
@dcramer Should Sentry be able to correctly group the case described [here](https://github.com/getsentry/raven-ruby/pull/278#issuecomment-69123573)?

nateberkopec:
Sorry, but I think we have to keep abs_path as the actual absolute path to [stay on spec](http://sentry.readthedocs.org/en/latest/developer/interfaces/index.html). If you think there's an issue with how Sentry groups events together, please open that issue on the main Sentry server repository.

nateberkopec:
Hmm not sure I *quite* understood this PR, you weren't actually changing abs_path, you were changing filename!

dannyfallon:
:+1: @eagletmt, this is precisely what I was getting at in the issue

lib/raven/interfaces/stack_trace.rb
```diff
@@ -36,6 +36,12 @@ def filename
         return nil if self.abs_path.nil?
 
         prefix = $LOAD_PATH.select { |s| self.abs_path.start_with?(s.to_s) }.sort_by { |s| s.to_s.length }.last
+        if prefix.nil? && Raven.configuration.project_root
+          project_root = Raven.configuration.project_root.to_s
+          if self.abs_path.start_with?(project_root)
```
nateberkopec: This ``self`` is unnecessary.

nateberkopec: Confusing I know since we use it everywhere else in this method.

nateberkopec:
@eagletmt If you could please rebase against master, I can merge.

eagletmt:
@nateberkopec rebased.

Issue #282, opened 2015-01-12
How to serialize Raven::Event?


marclennox:
I'm trying to use a Sidekiq job to send the raven events, which means I need to serialize the Raven::Event and deserialize it before sending.

I see there's a "to_hash" method but no "from_hash".

I tried to use YAML::load and YAML::dump but it yacks on procs from the config.

Any ideas how I can accomplish this?

dcramer:
Ideally we'd improve the API so you'd just serialize a hash, and the core send() method just took a simple mapping. This is how the Python API works which makes it fairly flexible.

We could do like we do in Python and ensure compatibility by having send(event) => send_remote(hash) and then let you override send_remote or similar.

That said I'm not sure we really need to ensure backwards compat on send

marclennox:
Agreed if I could simply send the hash, this would be ideal.  Is there a work-around with which I can achieve this now?

nateberkopec:
@marclennox Is there any way you could share a gist of the Sidekiq code you use to accomplish this? We need to improve our docs around pushing your Sidekiq sending to a job processor + I want to improve our API around this.

marclennox:
For sure, here you go.

https://gist.github.com/marclennox/b2dc84cc449164ff1397

marclennox:
Note that I'm using ActiveJob with Sidekiq as the provider.

nateberkopec:
ker-pow! addressed by the above.

PR #283, opened 2015-01-13
Modified the client send to accept both a Raven::Event and an already serialized "to_hash" event.

marclennox:
Addresses issue #282

### DIFF
--- a/lib/raven/client.rb
+++ b/lib/raven/client.rb
@@ -29,7 +29,14 @@ module Raven
         return
       end
 
-      Raven.logger.debug "Sending event #{event.id} to Sentry"
+      event_id = if event.respond_to?(:id)
+                   event.id
+                 elsif event.respond_to?(:[])
+                   event['id']
+                 else
+                   '<no event id>'
+                 end
+      Raven.logger.debug "Sending event #{event_id} to Sentry"
 
       content_type, encoded_data = encode(event)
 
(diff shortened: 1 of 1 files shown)


nateberkopec:
I like the principle, but not psyched about `is_a?`. Is there any way we can do this with some more Rubylike duck-typing?

marclennox:
I'll resubmit in a bit

—

marclennox:
I’ve rebased and re-pushed the commit.


—

marclennox:
I fixed a few issues with this PR, rebased and re-pushed.  Do you need anything else to accept it?

nateberkopec:
Yeah, I'm still not stoked about the code (respond_to? is still a bandaid IMO), but there's not really a fix for that without getting kind of crazy. 

That said, how about a test for sending a Hash?

nateberkopec:
@marclennox Also, I'll merge without this, but two suggestions for alternative implementations (again, if you just give me a test, I can do these things myself, so NBD):

* The #send method should try to convert whatever it receives into a Hash
* The caller of #send (Raven::Base I think?) should be responsible for calling #send with a Hash argument, and #send should assume that it's called with a hash-like object.

marclennox:
Thanks @nateberkopec, would you suggest I add a test to raven_spec.rb?

marclennox:
Also one thing to note, I didn't have to make any change to the send method because a hash responds to to_hash and returns itself.

nateberkopec:
@marclennox Yeah thats fine

marclennox:
This pull request is replaced by https://github.com/getsentry/raven-ruby/pull/300.

PR #290, opened 2015-01-19
fix argument name missmatch

poiati:
Related to issue: https://github.com/getsentry/raven-ruby/issues/289.


nateberkopec:
Thanks, and thanks for using "raven-head" I suppose via Github. It really helps us when people use the bleeding edge and fix stuff like this.

nateberkopec:
This is now covered by the integration testing in the Rails spec.

Issue #288, opened 2015-01-17
Stracktrace frames should strip the project_dir from abs_path for filenames


dannyfallon:
In the [filename method](https://github.com/getsentry/raven-ruby/blob/master/lib/raven/interfaces/stack_trace.rb#L35-L40) you're currently using the $LOAD_PATH array to determine if you can strip the start of the path for the current stacktrace frame. In Rails applications this presents some disadvantages, all related to grouping errors in Sentry

By not including the Rails root or project_dir when attempting to strip prefixes from the absolute path you end up using the full path for errors in places such as views, initialisers etc. This presents a problem if your path contains a variable, for example the timestamp or the SHA for the current version. You essentially end up polluting your sentry list with errors that _should_ be grouped together. Right now we're monkeypatching the filename method to work around this issue

In a bit of a tangent, perhaps using the full $LOAD_PATH is also causing some unintended problems for Rails apps. The Rails autoloader will add some paths such as `/app/models`, `/app/controllers` and `/lib` by default. Imagine a scenario where you've a model named `conversation.rb` and you've a lib of the same name. The filename for an error that happens in either will be the same, so the culprit will be the same. This presents issues in the Sentry UI listing, both errors have the same culprit (`conversation.rb`). It also makes things difficult when viewing the backtrace, which doesn't appear to expose the `abs_path` in an easy way.

nateberkopec:
Related: #278

dannyfallon:
Cheers Nate. We've been running the monkeypatch in production the last few days and it has certainly cleared up the grouping of errors. I'll try submit a PR this week for that behaviour for review

nateberkopec:
@dcramer Can you chime in on what parts of this abs_path issue are something we should solve in the client vs on the Sentry server?

dcramer:
Just to copy convo from IRC:

- absolute path is designed to be the absolute path on disk
- filename is designed to be some kind of normalized or relative path (i.e. you can trim project root)

dannyfallon:
Thanks @dcramer, #278 looks to solve the first issue where the Rails root remains on the filename.

How about your thoughts on the second one? Rails' autoloader adds some directories to the $LOAD_PATH, it might confuse some users by removing valuable path information

dcramer:
Assuming $LOAD_PATH is the same as PYTHONPATH, I think it's fine to only store the full path in abspath and truncate at $LOAD_PATH, but I'm not really sure here.

nateberkopec:
[...]

PR #291, opened 2015-01-24
Always preserve filename paths under project_root

eugeneius:
This addresses the second point from #288.

When a stack trace frame is from a path inside the project, the filename should be normalized relative to the project root. Only paths outside the project should be normalized based on `$LOAD_PATH`.

/cc @dannyfallon @nateberkopec


nateberkopec:
Thanks!

eagletmt:
:+1:

PR #293, opened 2015-01-31
Alias Raven.send to Rave.send_object_message

mattwalters:


nateberkopec:
*headdesk*. Yeah, we probably shouldn't be overriding send.

mattwalters:
Yeah, probably not. I'm happy to just rename the method as well. Although, I'm not sure if you want to do that since it is likely to break existing implementations.

nateberkopec:
Yeah. A lot of people are using Raven.send in their job processors are well. We'll definitely need a deprecation cycle.

nateberkopec:
Closing in favor of #326 for now. Thanks for pointing this out.

PR #294, opened 2015-02-03
union instead of concatenation, because duplicate fields

wyattisimo:


nateberkopec:
Oh hi Jared!

Do I get to bother you to add a test case?

wyattisimo:
yes

sigh

nateberkopec:
I like #295. I think it would be best if sanitize_fields was just a simple array. By default, the array is filled with the default list, but you can manipulate it as you would an array to modify.

wyattisimo:
@nateberkopec added test case

spec/raven/configuration_spec.rb
```diff
@@ -130,4 +130,16 @@
     end
   end
 
+  context 'configuration for sanitize fields' do
+    it 'should union default sanitize fields with user-defined sanitize fields' do
+      fields = Raven::Processor::SanitizeData::DEFAULT_FIELDS | %w(test monkeybutt)
```
nateberkopec: Shouldn't we use a field like "password" or something? Neither of the fields you're adding here are in DEFAULT_FIELDS

wyattisimo: Right, but I'm adding those to the list of default fields and assigning all of that to `sanitize_fields`.

nateberkopec: Right you are!

PR #297, opened 2015-02-04
Proposed: a new interface for sanitize_fields config

nateberkopec:
Old busted:
```ruby
Raven.configure do |config|
  config.sanitize_fields = %w(host User-Agent)
end
# We'll sanitize the defaults (password, ssn, etc), PLUS host and User-agent
```

New hotness:
```ruby
Raven.configure do |config|
  config.sanitize_fields = %w(host User-Agent)
end
# We sanitize ONLY host and User-Agent. If you want the old behavior, you must...
Raven.configure do |config|
  config.sanitize_fields << %w(host User-Agent)
end
```

After writing this all out, it's obvious that such a change should have some kind of deprecation cycle, otherwise we'd have a bunch of people not sanitizing things they thought they were! 

Also, after writing this change out, I'm not sure what the benefit is, since most of the default fields are "sanitized" anyway by the upstream Sentry server. It seems like an area where we should take convention over configuration. Or at least, the interface should make it a lot more difficult to *not* sanitize fields like "password" and "ssn". This change, I think, introduces *too much* configurability.

### DIFF
--- a/lib/raven/processor.rb
+++ b/lib/raven/processor.rb
@@ -2,11 +2,8 @@ require 'json'
 
 module Raven
   class Processor
-    attr_accessor :sanitize_fields
-
     def initialize(client)
       @client = client
-      @sanitize_fields = client.configuration.sanitize_fields
     end
 
     def process(data)
@@ -22,6 +19,5 @@ module Raven
         nil
       end
     end
-
   end
 end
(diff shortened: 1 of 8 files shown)


lib/raven/processor.rb
```diff
@@ -2,11 +2,8 @@
 
 module Raven
   class Processor
-    attr_accessor :sanitize_fields
```
nateberkopec: whether or not I merge this, I'm going to do this change in a separate PR. this stuff should not be in the processor base class.

nateberkopec: perhaps I could generalize processor config, such that you can provide configuration keys to your own custom processor chain. I doubt anyone *really* writes their own processors though.

PR #301, opened 2015-02-14
Remove BetterAttrAccessor

nateberkopec:
We're gonna learn @dcramer some Ruby


dcramer:
Impossible :)

seanlinsley:
Well `BetterAttrAccessor` is something I added (a professional Ruby dev) in #201 to remove the dependence on Hashie. I still think improving on `attr_accessor` to provide default values is better than overriding `initialize` like you do here.

Not a huge difference either way though ¯\\\_(ツ)\_/¯

lib/raven/interfaces/http.rb
```diff
@@ -15,6 +15,7 @@ class HttpInterface < Interface
     def initialize(*arguments)
       self.headers = {}
       self.env = {}
+      self.cookies = nil
       super(*arguments)
```
seanlinsley: This change seems unrelated to the rest of the PR. Perhaps this was accidentally added?

nateberkopec: I don't think so. I can't recall immediately but removing better_attr_accessor did cause this test to fail.

nateberkopec: Er, rather - cause a test related to this to fail.

nateberkopec: I remember now - without this, cookies is not included in `instance_variables`, so it's not included in the to_hash method (unless you set it to something yourself). [That caused this test to fail](https://github.com/getsentry/raven-ruby/pull/201/files#diff-2e67fc5726ac8c5cab5523f71896e49cR175).

nateberkopec:
@seanlinsley Thanks - you and I are on the same page with removing dependencies. I agree that the default syntax is nice, but I just don't think it's worth ~40 LOC. Overriding init is common enough that I'd rather just run with that approach, unless I'm missing something here that makes that approach behave significantly differently than BetterAttrAccessor. I think we can make the Interface objects even simpler, this was just one step towards that.

Issue #296, opened 2015-02-03
Trim default Rack HTTP data


dcramer:
It's possible for the default http context (from HttpInterface.from_rack) to contain data that is too large for servers to except.

We should trim (at least) req.POST/req.body to something like 16kb.

nateberkopec:
Folding into #281

PR #303, opened 2015-02-14
Try to simplify Rack HTTP interface a little

nateberkopec:
+ Also move it to a place where it's only required when needed
+ Add some more CGI variables to env


nateberkopec:
Going to try to solve #296 while I'm at it here

Issue #170, opened 2014-04-04
Raven::Processor::SanitizeData does not sanitize posted JSON data


Timo614:
JSON data isn't sanitized so it leaks sensitive parameters for APIs that post their data as JSON.

Any data passed in will appear in plaintext in sentry.

```
curl -X POST -H "Content-Type: application/json" -d "{\"email\":\"email@example.com\",\"password\":\"ok\"}" http://0:0:0:0:5000/api/login
```

If there's an exception it appears in plaintext in the sentry app parameters list.

I've patched it here:
https://github.com/SparkartGroupInc/raven-ruby/commit/19f28287172025270f9285758d0d1a13f39b06f9

I'm not sure on the efficiency of the json parsing before it throws the error but it probably would be more efficient if it checked to see if it passed some sort of simple smell test with regex before it runs since any string would go through it during the sanitation.

It's fine for my purposes though -- let me know if you want me to file a PR for it.

PR #306, opened 2015-02-14
Don't decode and sanitize JSON unless it matches our regex

nateberkopec:
This conditional is getting atrocious.

Inspired by #170. This should considerably speed up Raven capture for anyone working with JSON.


nateberkopec:
This conditional is so bad, but I'm going to leave refactoring it for another day. More important that we stop parsing JSON unless absolutely necessary.

PR #309, opened 2015-03-12
More detailed status report upon configuration

faber:
I was getting thrown off by the "ready to catch errors" message that was popping up even in environments where it was configured not to send.  My use case is that I have a Raven.configure block in my code all the time, but whether or not I actually enable capture is controlled the presence/absence of the SENTRY_DSN env var.  Since calling Raven.capture triggers the message, it came up every time regardless of whether I was actually capturing errors.

(I know this is a somewhat superficial PR, sorry!)

Two points open for debate (besides whether or not this PR is desired/worthwhile at all) are the actual text used to indicate that errors will not be sent, and whether or not to alias report_status with report_ready, for backwards compatibility purposes.  Technically report_ready is part of the public API, so even tho its doubtful that anyone was using it directly, better safe than sorry I suppose.  Another option is just leaving it named report_ready, but that seems marginally less correct.

Cheers!

### DIFF
--- a/lib/raven/base.rb
+++ b/lib/raven/base.rb
@@ -49,9 +49,15 @@ module Raven
     end
 
     # Tell the log that the client is good to go
-    def report_ready
-      logger.info "Raven #{VERSION} ready to catch errors"
+    def report_status
+      return if client.configuration.silence_ready
+      if client.configuration.send_in_current_environment?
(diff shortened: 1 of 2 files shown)


dcramer:
+1 from my end -- anything that makes it more obvious whats going on, without being intrusive is generally useful

nateberkopec:
Thanks!

PR #310, opened 2015-03-21
Add support for sending release version

eugeneius:
http://sentry.readthedocs.org/en/latest/developer/client/#release

The [Python](https://github.com/getsentry/raven-python/commit/332e05a8621cb170985ef2feb855810e022b2ac8) and [Go](https://github.com/getsentry/raven-go/commit/82e450d4d524d18d05dc3a90ecd8055f349d3dd2) clients can already send this parameter, so I thought I'd add it to the Ruby client too. As I understand it, `release` is basically just a tag for now, but will tie into https://github.com/getsentry/sentry/issues/1187 later (@dcramer, am I right about this?).

### DIFF
--- a/lib/raven/configuration.rb
+++ b/lib/raven/configuration.rb
@@ -71,6 +71,8 @@ module Raven
 
     attr_accessor :server_name
 
+    attr_accessor :release
+
     # DEPRECATED: This option is now ignored as we use our own adapter.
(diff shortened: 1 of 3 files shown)


nateberkopec:
Thank you!

eugeneius:
Thanks for the quick review @nateberkopec :+1:

PR #312, opened 2015-04-03
Support exception chaining

### LINKED ISSUE #190: Support chained exceptions
In Ruby 2.1.0, [`Exception#cause`](http://www.ruby-doc.org/core-2.1.0/Exception.html#method-i-cause) was added. While @dcramer informed me in #sentry on irc.freenode.net that the Sentry server already supports chained exceptions, the Ruby client does not. I propose that we add support for chained exceptions to the Ruby client. We could just detect whether an exception responds to `#cause` and, if so, add the cause's message and backtrace to the exception report.

I'm just opening this for discussion now. I may propose this work as a "10% time" project at my current client or try to work on it myself as early as this weekend.

javawizard:
Fixes #190


dcramer:
Unrelated to the code, but any idea which version of Ruby started offering chaining? I didn't actually realize it was even a feature.

dcramer:
Ignore that comment, I'm an idiot. The ticket actually points it out :)

nateberkopec:
Thanks! Will take a look.

javawizard:
Awesome. Just pushed up a fix to not check reporting when causes aren't supported.

lib/raven/event.rb
```diff
@@ -159,14 +159,25 @@ def self.from_message(message, options = {})
...
+      evt.interface(:exception) do |exc_int|
+        exceptions = [exc]
+        while exc.respond_to?(:cause) && exc.cause
+          exceptions << exc.cause
+          exc = exc.cause
+        end
+        exceptions.reverse!
+
+        exc_int.values = exceptions.map do |exc|
+          SingleExceptionInterface.new do |int|
```
nateberkopec: Hmmm, I'm not really a fan of this name/a new class. Why can't you just use multiple instances of ExceptionInterface?

lib/raven/interfaces/exception.rb
```diff
@@ -4,15 +4,12 @@ module Raven
   class ExceptionInterface < Interface
 
     name 'exception'
-    attr_accessor :type
-    attr_accessor :value
-    attr_accessor :module
-    attr_accessor :stacktrace
+    attr_accessor :values
```
nateberkopec: Now I can't find the old documentation for how Exception interfaces are supposed to work with Sentry. I thought that info was here: http://sentry.readthedocs.org/en/latest/developer/interfaces/ but it isn't anymore. @dcramer where'd that go?

dcramer: Docs should be fixed, but this looks correct to me (values is a list of exception instances)

PR #317, opened 2015-04-13
Raven::Transports::HTTP#send should return response

eagletmt:
I'm writing another transport class which uses Raven::Transports::HTTP internally.
I'd like to know what is the response of HTTP#send, especially whether it returns 429 or not.


nateberkopec:
:+1:

PR #318, opened 2015-04-14
Fix raven:test NoMethodError: undefined method `id` for Hash error

weynsee:
When running `rake raven:test`, the task exits with NoMethodError: undefined method `id` for #<Hash:...> error. This is because `Raven.capture_exception(e)` returns a hash representation of `Raven:Event`, so `event.id` leads to an error. This PR checks whether the returned event is a Hash, and if so, retrieves the id from it.

### DIFF
--- a/lib/raven/cli.rb
+++ b/lib/raven/cli.rb
@@ -47,7 +47,11 @@ module Raven
       end
 
       if evt && !(evt.is_a? Thread)
-        puts "-> event ID: #{evt.id}"
+        if evt.is_a? Hash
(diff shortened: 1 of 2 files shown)


nateberkopec:
Ack! Need to improve our test coverage here too if that passed anyway.

weynsee:
I've added specs for CLI (which is largely copied from integration_spec.rb) that would have caught that error. Hope this helps!

nateberkopec:
Awesome! Should get a release for this out.

PR #300, opened 2015-02-11
Addresses issue #282. Modified the client send method to accept an event in object or hash format.

marclennox:
This pull request replaces https://github.com/getsentry/raven-ruby/pull/283.

PR #319, opened 2015-04-14
Fix event hash lookup

eagletmt:
Event keys are symbolized at #256, but they're looked up by String
at #300.


eagletmt:
Ugh, old Ruby doesn't support `\h` regexp :imp:

PR #320, opened 2015-04-15
Fixed blockless Raven#capture

dinosaurjr10:
#272 introduced bugs into the blockless semantics of #capture.
* Fixed #capture to not yield when no block given
* Fixed #capture to pass options to #install_at_exit_hook
* Fixed #install_at_exit_block to check for the existence of $ERROR_INFO

### DIFF
--- a/lib/raven/base.rb
+++ b/lib/raven/base.rb
@@ -91,14 +91,17 @@ module Raven
     #     MyApp.run
     #   end
     def capture(options = {})
-      install_at_exit_hook unless block_given?
-      begin
-        yield
-      rescue Error
-        raise # Don't capture Raven errors
-      rescue Exception => e
-        capture_exception(e, options)
-        raise
+      if block_given?
+        begin
+          yield
+        rescue Error
+          raise # Don't capture Raven errors
+        rescue Exception => e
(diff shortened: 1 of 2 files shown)


nateberkopec:
Hmm, obviously we never considered that people would use #capture this way.

Can you give me an example of how/why you're using #capture without a block?

dinosaurjr10:
[#capture](http://www.rubydoc.info/gems/sentry-raven/0.12.3/Raven.capture) is documented as processing exceptions globally if no block is given. Isn't that why an at_exit handler is registered when no block is given?

We want to capture exceptions globally, so Instead of repeating Raven configuration and startup code in all of our applications' start scripts, they require a sharable raven_init.rb file which performs these functions. The global exception processing makes this pretty slick.
```
require 'raven'
Raven.configure { |config| config.should_send = ->(_) { !!ENV['SENTRY_DSN'] } }
Raven.capture
```

nateberkopec:
Thanks! I'm just unfamiliar with this usage because I'm always using Sentry with Rack middleware.

dinosaurjr10:
Thanks, @nateberkopec!

PR #322, opened 2015-04-24
Ensure that the tags hierarchy is respected where event > context > configuration for tags using the same key

JonathanBatten:
In the wiki, when talking about [setting tags in the default configuration](https://github.com/getsentry/raven-ruby/wiki/Advanced-Configuration#tags) it sounded to me like you could override the values for each of the keys set in `tags`.

`You can configure default tags to be sent with every event. These can be overridden in the context or event.`

With the way the tags are merged together currently, the priority of *tags with the same key* is context > configuration > event.

This seems wrong to me as I would think that if you set a tag explicitly in the event that is being sent to Sentry that it should take precedence over context and configuration tag values.

I've made a change to work the way I would imagine it should work.

### DIFF
--- a/lib/raven/event.rb
+++ b/lib/raven/event.rb
@@ -56,8 +56,12 @@ module Raven
 
       @user.merge!(@context.user)
       @extra.merge!(@context.extra)
+
+      tags = @tags.dup
+      @tags = {}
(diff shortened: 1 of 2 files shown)


nateberkopec:
Thanks! Good catch.

PR #323, opened 2015-04-30
Change `Rails.inject` to look for the `railties` gem, not `rails`.

lucasmazza:
`rails` is a meta gem that can be skipped in a Rails app if we don't want to install gems like `actionmailer`, `activerecord` or `activejob`, while the `railties` gem is always present in a Rails app.

To ensure that Raven is properly loaded when the `rails` gem is not present we look for the `railties` gem, not `rails` in the `Gem.loaded_specs` Hash.

### DIFF
--- a/lib/raven/base.rb
+++ b/lib/raven/base.rb
@@ -195,7 +195,7 @@ module Raven
 
     # Injects various integrations
     def inject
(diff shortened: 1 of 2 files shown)


nateberkopec:
:+1: Great catch.

PR #327, opened 2015-05-04
Add support for ActiveJob.

lucasmazza:
This PR has some of the initial code to add support for Active Job subclasses regardless of the queue backend that is being used - Raven currently supports Sidekiq on it's own but would be nice to have a broader support for other backends like SuckerPunch (the one I'm using), Resque, DelayedJob, etc.

I couldn't find any tests for the Sidekiq support to use as a base, but I'm take this to a spin in our app to see how it goes. Also, I fear that loading both the ActiveJob support + Sidekiq might generate duplicate reports since both integrations would be executed together, but this is something that I need to test it out to be sure about it.

### DIFF
--- a/lib/raven/integrations/rails.rb
+++ b/lib/raven/integrations/rails.rb
@@ -14,6 +14,13 @@ module Raven
       end
     end
 
+    initializer 'raven.active_job' do
+      ActiveSupport.on_load :active_job do
+        require 'raven/integrations/rails/active_job'
+        include Raven::Rails::ActiveJob
+      end
+    end
+
     config.after_initialize do
       Raven.configure do |config|
         config.logger ||= ::Rails.logger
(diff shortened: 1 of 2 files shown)


nateberkopec:
(thumbsup) We should probably only carry ActiveJob integration in the base gem, and spin out Sidekiq integration to a separate gem.

lucasmazza:
@nateberkopec one alternative that I can think of is make it disabled by default - the deprecation cycle/upgrade path could be simpler.

lucasmazza:
@nateberkopec sorry for the huge delay on this one. I've tested this branch against a fresh Rails app using Sidekiq and added a safeguard to the `ActiveJob` plugin to let the Sidekiq handle the exceptions instead of collecting the same error twice.

dcramer:
I've reverted this for now until we sort out #368.

It might be that as a quick workaround we remove the clause to prevent the duplicate and sort out those problems later.

PR #329, opened 2015-05-07
config to allow disabling credit card number sanitization

codekitchen:
This allows for still using SanitizeData, while disabling this specific value
sanitizer, which is broad enough to capture any other 13-16 digit numbers as well.

As discussed in https://github.com/getsentry/raven-ruby/issues/328

Overall I kinda liked my second suggestion better, splitting the credit card sanitizer into a separate processor, but decided against it for backwards compatibility reasons. It'd be sad if people suddenly lost that sanitization because they were doing something like

```ruby
config.processors = [Raven::Processor::SanitizeData, MyCustomProcessor]
```

### DIFF
--- a/lib/raven/configuration.rb
+++ b/lib/raven/configuration.rb
@@ -95,6 +95,9 @@ module Raven
     # additional fields to sanitize
     attr_accessor :sanitize_fields
 
+    # Sanitize values that look like credit card numbers
+    attr_accessor :sanitize_credit_cards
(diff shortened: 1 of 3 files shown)


nateberkopec:
> It'd be sad if people suddenly lost that sanitization because they were doing something like

Great point!

codekitchen:
thanks for the snappy response!

PR #330, opened 2015-05-09
Make user and extra context merge order match tags

eugeneius:
Followup to #322.

Tags passed when capturing an event override tags in the request context. For consistency, user and extra context should behave the same way.


nateberkopec:
Thanks!

PR #332, opened 2015-05-12
Use send_event when capturing an exception

ripta:
`Raven.send`, which happens to be used from `Raven.capture_type` is deprecated, and causes deprecation warnings. I would count this as a bug, because the end user cannot directly avoid the warning if they use `Raven.capture_type` directly—or in our case using it through the Rack integration middleware.

This PR makes `capture_type` use `send_event` and avoid the warning altogether, and changes the specs to match. 

All green: 170 examples, 0 failures.

Thanks!

### DIFF
--- a/lib/raven/base.rb
+++ b/lib/raven/base.rb
@@ -78,7 +78,7 @@ module Raven
     #
     # @example
     #   evt = Raven::Event.new(:message => "An error")
-    #   Raven.send(evt)
+    #   Raven.send_event(evt)
     def send_event(event)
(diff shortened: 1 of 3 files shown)


nateberkopec:
Derp! Sorry, I thought I caught all of these. Thanks.

nateberkopec:
I'll push out a tiny version release tonight to fix.

ripta:
\o/

nateberkopec:
0.13.3 has been pushed to rubygems.

PR #334, opened 2015-05-12
Correct handling of JRuby stacktraces

dcramer:
Fixes GH-333

PR #335, opened 2015-05-12
Better handling of unreachable file context

dcramer:
This will improve things like JRuby where the context is unavailable

Refs GH-334


lib/raven/linecache.rb
```diff
@@ -5,17 +5,24 @@ class LineCache
     class << self
       CACHE = {}
 
+      def isValidFile(path)
```
dcramer: realizing that this is probably not the way to name things

Issue #261, opened 2014-12-08
View local variables


ineu:
How do I view local variables when exception happens in Rails? According to https://www.getsentry.com/for/ruby/ is should be possible in Ruby (there's "Expand your scope with context locals" section), but #7 is there for 2.5 years and is not yet closed. So is it not possible? If not, then is it possible to do it manually by using `extra` as described in https://github.com/getsentry/raven-ruby/wiki/Context ?
Thanks in advance.

nateberkopec:
We've fixed the problem on the marketing page. For ongoing discussion of local variable capture, see #7

PR #337, opened 2015-05-13
Collect local variables

if1live:
## Overview
I add new option, capture_locals. (https://github.com/getsentry/raven-python/blob/133fcdb74869bb270f75642889b17b0752bf0b6e/raven/conf/defaults.py#L54-L55)

If config.capture_locals is true, raven capture local variables and send to sentry.
See #7. [binding_of_caller](https://github.com/banister/binding_of_caller) gem. 

If config.capture_locals is false, use previuse behavior.

Default value of capture_locals is false.
Because I can't believe that code run in all environments. (too many ruby implements.)
We can disable capture_locals if not working.

## Related Issue
* #7
* #261

### DIFF
--- a/lib/raven/base.rb
+++ b/lib/raven/base.rb
@@ -15,6 +15,7 @@ require 'raven/processor/sanitizedata'
 require 'raven/processor/removecircularreferences'
 require 'raven/processor/utf8conversion'
 require 'binding_of_caller'
+require 'thread'
 
 major, minor, patch = RUBY_VERSION.split('.').map(&:to_i)
 if (major == 1 && minor < 9) || (major == 1 && minor == 9 && patch < 2)
@@ -26,16 +27,30 @@ module Kernel
     if @@monkey_patch_raise_occur == true
       return
     end
-    apply_monkey_patch_raise
-    @@monkey_patch_raise_occur = true
+
+    @@monkey_patch_raise_mutex.synchronize {
+      return if @@monkey_patch_raise_occur == true
+
+      apply_monkey_patch_raise
+      @@monkey_patch_raise_occur = true
+    }
+  end
+
+  def self.monkey_patch_raise_occur
+    @@monkey_patch_raise_occur
+  end
+
+  def self.monkey_patch_original_raise
+    @@original_raise
   end
 
   private
   def self.apply_monkey_patch_raise
-    original_raise = method(:raise)
+    @@original_raise = method(:raise)
+
     define_method(:raise) do |*args|
       begin
-        original_raise.call(*args)
+        @@original_raise.call(*args)
       rescue Exception => e
         prev_stack_info = e.instance_variable_get(:@stack_info)
         if prev_stack_info.nil?
@@ -49,12 +64,14 @@ module Kernel
             e.backtrace[idx].instance_variable_set(:@stack_info, caller_obj)
           end
         end
-        original_raise.call(e)
+        @@original_raise.call(e)
       end
     end
   end
 
   @@monkey_patch_raise_occur = false
+  @@monkey_patch_raise_mutex = Mutex.new
+  @@original_raise = nil
 end
 
 module Raven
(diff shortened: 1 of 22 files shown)


lib/raven/base.rb
```diff
@@ -14,12 +14,49 @@
...
+          callers.each_with_index do |caller_obj, idx|
+            e.backtrace[idx].instance_variable_set(:@stack_info, caller_obj)
+          end
+        end
+        original_raise.call(e)
+      end
+    end
+  end
+
+  @@monkey_patch_raise_occur = false
```
nateberkopec: Are there threading issues with using a class variable here?

lib/raven/base.rb
```diff
@@ -14,12 +14,49 @@
...
+require 'binding_of_caller'
 
 major, minor, patch = RUBY_VERSION.split('.').map(&:to_i)
 if (major == 1 && minor < 9) || (major == 1 && minor == 9 && patch < 2)
   require 'raven/backports/uri'
 end
 
+module Kernel
+  def self.monkey_patch_raise
+    if @@monkey_patch_raise_occur == true
```
nateberkopec: Hmmm... I see why you're doing this, but I wonder if there's a better way.

lib/raven/base.rb
if1live: 1. add @@monkey_patch_raise_mutex.synchronize { ... } 
2. check flag twice (outside of critical section, inside of critical section)

sentry-raven.gemspec
```diff
@@ -17,6 +17,7 @@ Gem::Specification.new do |gem|
 
   gem.add_dependency "faraday", ">= 0.7.6"
   gem.add_dependency "certifi"
+  gem.add_dependency "binding_of_caller" if RUBY_VERSION >= '1.9.2' && !defined?(JRUBY_VERSION)
```
eugeneius: The gemspec is evaluated when the gem is built, not when it's installed - `RUBY_VERSION` here will return the version of Ruby that is being used to build the gem. RubyGems doesn't support conditional dependencies.

I don't know if there's a way to maintain compatibility with Ruby 1.8.7 and also optionally depend on binding_of_caller.

nateberkopec: I think we should remove this line, and instead, people should be instructed to add "binding_of_caller" to their Gemfile if they want to use this feature. Similar to how `dalli` uses `kgio`.

eugeneius: That sounds good!

Dalli tries to require `kgio` at startup, but rescues the `LoadError` if it fails: https://github.com/mperham/dalli/blob/master/lib/dalli/socket.rb

If we did the same thing here, we could refuse to enable the feature unless `BindingOfCaller` is defined.

eugeneius:
Since this feature is optional and disabled by default, we shouldn't monkey patch `Kernel` unless it's going to be used. Could we put the monkey patch in a separate file, and only require it when `capture_locals` is set to `true`?

sentry-raven.gemspec
nateberkopec: We should re-use [the `integrations` infrastructure](https://github.com/getsentry/raven-ruby/blob/master/lib/raven/base.rb#L203) already in the project, which deals better with Gemfile ordering issues.

nateberkopec:
I think `Raven.inject` is called before the configuration is loaded, but there may be another way to go about it.

if1live:
I modify raven-ruby to support collecting local variables, and I use it 3 months.
I found that collected local variables are broken when implicit exception raised (ex: 1/0)
I don't know how to fix it. So, I close this is now. If I found a way to fix bug, I will rewrite pull request.

Shelvak:
Hi there, I know that was abandoned, But maybe after almost 2 years you have something similar.
Is there anything that could track the locals in the trace?  I use the python client, and all the locals context is great when you need to debug. Thanks in advance.

nateberkopec:
I'm investigating using FFI to provide this functionality for MRI 2.0+ and maybe JRuby 9k, but no timeline. This feature is only really useful in production, but there are no real production-ready APIs for it, so it's pretty complicated

Shelvak:
Ok, thanks for the info and the answer. 

Cheers ✋

Issue #331, opened 2015-05-12
Why is SSL verification disabled by default?


gregkare:
I've just found out you're disabling SSL verification by default in the config

https://github.com/getsentry/raven-ruby/blob/17e6957046d41a50e401d5cd70e56b0fa2cb0f95/lib/raven/configuration.rb#L116

gregkare:
I've just found this through a friend:

https://github.com/getsentry/raven-ruby/commit/2351f0a98cf7a2be13d52305965080523f6bf225

Do you still have the same view on SSL verification by default as in 2013?

nateberkopec:
paging @dcramer

dcramer:
I'm pretty sure my view is the same. I'm not part of the Ruby ecosystem so I don't know if there's a great solution, but in Python we bundle some basic CA certs which ensures at the very least common providers work, as well as getsentry.com:

https://pypi.python.org/pypi/certifi

I'd prefer we not disable it, but not at the cost of every user struggling to figure out how to make SSL work.

nateberkopec:
[It looks like certifi has a Rubygem](https://github.com/certifi/ruby-certifi), although it looks mostly unused in the last year. But @hone (one of the maintainers, it seems) is a badass, so I'm sure we could work with him to use it.

Would definitely be open to PRs that turned SSL verification to "on" and used Certifi.

dcramer:
I'll see if i cant put together a patch for this tomorrow as it seems fairly trivial.

PR #338, opened 2015-05-13
Default SSL verification via Certifi

dcramer:
Fixes GH-331


dcramer:
I'm not sure if there's any caveats at, here but it works with getsentry.com so :+1: in my book

drewish:
For folks like myself who were wondering where this went, it was removed in #352.

PR #339, opened 2015-05-16
Capture exceptions in runner tasks

### LINKED ISSUE #325: Sentry doesn't seem to be recording errors when started via Rails runner
It seems like Sentry doesn't record exceptions for some reason when running code via Rails runner. I've put together a minimal example Rails app to reproduce the issue: https://github.com/nfm/raven_rails_runner_bug

Basically, running any code via Rails runner that causes an exception (eg. `./bin/rails runner '1 / 0'`) isn't then capturing the exception for me. This happens in production, where other kinds of exceptions are being recorded correctly (from the webserver, but also from rake tasks and delayed jobs). It also happens in development if I configure Raven's environments to include development.

The example repo should be pretty straightforward but if not let me know, I'm happy to clarify what's going wrong however I can. Cheers :)

eugeneius:
Fixes #325.

Install Raven's `at_exit` hook whenever `rails runner` is invoked so that if an exception occurs during the task it will be reported.

I'm not really sure how to write a good test for this, but from manual testing it seems to work.

### DIFF
--- a/lib/raven/integrations/rails.rb
+++ b/lib/raven/integrations/rails.rb
@@ -34,5 +34,9 @@ module Raven
     rake_tasks do
       require 'raven/integrations/tasks'
     end
+
+    runner do
+      Raven.capture
+    end
   end
 end


nateberkopec:
*boom*

That's the second time I've forgotten about the at_exit hook.

PR #343, opened 2015-06-03
Capture exceptions swallowed by rails

robertclancy:
ActionDispatch's ShowExceptions swallows exceptions and stores the cause
in a rack env variable "action_dispatch.exception". Use this to report
on any exceptions swallowed by ShowExceptions.

### DIFF
--- a/spec/raven/integrations/rack_spec.rb
+++ b/spec/raven/integrations/rack_spec.rb
@@ -46,6 +46,22 @@ describe Raven::Rack do
     stack.call(env)
   end
 
+  it 'should capture rails errors' do
+    exception = build_exception
+    env = {}
+
+    expect(Raven::Rack).to receive(:capture_exception).with(exception, env)
+
+    app = lambda do |e|
+      e['action_dispatch.exception'] = exception
+      [200, {}, ['okay']]
+    end
+
+    stack = Raven::Rack.new(app)
+
+    stack.call(env)
+  end
+
   it 'should clear context after app is called' do
     Raven::Context.current.tags[:environment] = :test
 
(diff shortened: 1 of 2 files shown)


spec/raven/integrations/rack_spec.rb
```diff
@@ -46,6 +46,22 @@
     stack.call(env)
   end
 
+  it 'should capture rails errors' do
```
nateberkopec: Can you just change this line to say 'should capture rails errors when ActionDispatch::ShowExceptions is enabled'?

seanlinsley:
This should be a setting. For example, we set `rescue_from ActiveRecord::RecordNotFound` globally in our app. We don't want those showing up in Sentry.

robertclancy:
Any exceptions rescued in a controller using rescue_from do not end up in the "action_dispatch.exception" env variable.

As far as I can tell, the only place that env variable is set is in the [ShowExceptions](https://github.com/rails/rails/blob/e75b92c032ddbd0e664c4fd241bec5668e35ee96/actionpack/lib/action_dispatch/middleware/show_exceptions.rb#L46) middleware, which catches exceptions not rescued in a controller.

nateberkopec:
`rescue_from` doesn't use ShowExceptions. This LGTM.

greysteil:
This is causing some issues for us - our app specifies and [exceptions_app](https://github.com/rails/rails/blob/e75b92c032ddbd0e664c4fd241bec5668e35ee96/railties/lib/rails/application/default_middleware_stack.rb#L111) which handles things like parameter parsing errors. We don't want those rescued exceptions to be reported to sentry.

My hunch is that the best way forward on this one is to make this a setting. Alternatively we could delete the `action_dispatch.exception` env variable once the exception has been handled by our exceptions app, but that doesn't feel right to me. @nateberkopec / @seanlinsley - any thoughts?

PR #344, opened 2015-06-04
Always return the `Event` object, regardless of what the `async` block returns.

lucasmazza:
When using an `async` block, the capture methods would return the block returned value when called or the `Event` object when async isn't defined, which could possibly break any client code that would always expect the event object instead of a `ActiveJob::Base` instance or a `Sidekiq` job, for instance.


nateberkopec:
:+1: for always returning the same type!

PR #346, opened 2015-06-12
Update event.rb

lsb:
Properly scope Digest::MD5, explicitly include md5


nateberkopec:
LGTM, but was this causing issues for you somehow?

lsb:
Yes, if you don't have digest/md5 in Ruby and you'd deep in Raven::Event, exciting scoping rules assume that this unseen Digest::MD5 is referring to Raven::Event::Digest::MD5, and it can't find Raven::Event::Digest.

PR #351, opened 2015-07-28
Fix rack/lint errors

### LINKED ISSUE #292: Storing Time.now in Rack env Violates the Rack Spec
Hi Sentry,

While testing the sentry gem with [one that I maintain](https://github.com/appneta/oboe-ruby), I found that if you also use Rack::Lint, it complains that env contains a non-string value.

getsentry raven time now 2015-01-27 18 11 11

I tracked the value down to [this line](https://github.com/getsentry/raven-ruby/blob/master/lib/raven/integrations/rack.rb#L50).

The [Rack Spec](http://www.rubydoc.info/github/rack/rack/master/file/SPEC) says "The CGI keys (named without a period) must have String values."  You may be able to get away with this by simply renaming `requested_at` to `sentry.requested_at`.

Sorry this isn't a PR - time is a bit short this week. :-(

mattolson:
* The rack spec states that keys added by the application should contain
  one dot and be prefixed by something unique. In addition, it states that
  top level keys in the env should be string values only.
* Prior to this change, when running with rack/lint, you get the following error:
`Rack::Lint::LintError: env variable requested_at has non-string value 2015-07-28 17:16:36 +0000`
* Fixes #292

### DIFF
--- a/lib/raven/integrations/rack.rb
+++ b/lib/raven/integrations/rack.rb
@@ -23,8 +23,8 @@ module Raven
   class Rack
 
     def self.capture_type(exception, env, options = {})
-      if env['requested_at']
-        options[:time_spent] = Time.now - env['requested_at']
+      if env['raven.requested_at']
+        options[:time_spent] = Time.now - env['raven.requested_at']
       end
       Raven.capture_type(exception, options) do |evt|
         evt.interface :http do |int|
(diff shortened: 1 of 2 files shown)


mattolson:
ping @pglombardo @nateberkopec

pglombardo:
Looks great!  Thanks for fixing!

nateberkopec:
If you can fix the tests failing on Rails 3.2, will merge.

mattolson:
@nateberkopec Ok, I fixed the problem with Rails 3.2. There was a minor difference in how Rack handles default headers.

nateberkopec:
Woop woop! :tada:

PR #355, opened 2015-08-12
Fix a problem with the at_exit error capturing

kratob:
I've run into a problem where `Raven` did not capture errors in a `rails runner` task. It worked in `development` but not on `production`.

The issue is that `Raven` uses the `$ERROR_INFO` global variable. This is an alias for `$!`, but only if you first do a `require 'English'` (http://ruby-doc.org/stdlib-2.1.2/libdoc/English/rdoc/English.html).

I fixed it and added a test. If you're uncomfortable with the test (which has to do a `fork`), feel free to just pick the second commit.

### DIFF
--- a/lib/raven/base.rb
+++ b/lib/raven/base.rb
@@ -228,9 +228,9 @@ module Raven
 
     def install_at_exit_hook(options)
       at_exit do
-        if $ERROR_INFO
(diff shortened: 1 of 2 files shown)


foobear:
Any update on this? We could really use it. :)

dcramer:
Sorry about missing this. Thanks for the patch!

PR #360, opened 2015-08-27
Allow configurable integrations

cgthornt:
This allows inclusion of raven without automatic inclusion of integrations. 

For example, to use all default integrations without sidekiq:

```ruby
require 'sentry-raven-without-integrations'
Raven.inject_without(:sidekiq)
```

To only include railties:

```ruby
require 'sentry-raven-without-integrations'
Raven.inject_only(:ralties)
```

## Todos:

- [x] Specs
- [x] Actually test this out in a dummy app

### DIFF
--- a/lib/raven/base.rb
+++ b/lib/raven/base.rb
@@ -21,6 +21,8 @@ if (major == 1 && minor < 9) || (major == 1 && minor == 9 && patch < 2)
 end
 
 module Raven
+  AVAILABLE_INTEGRATIONS = %w[delayed_job railties sidekiq rack rake]
+
   class << self
     # The client object is responsible for delivering formatted data to the Sentry server.
     # Must respond to #send. See Raven::Client.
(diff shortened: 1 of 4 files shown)


lib/raven/base.rb
```diff
@@ -201,23 +203,39 @@ def rack_context(env)
...
+    end
+
+    def inject_only(*only_integrations)
+      only_integrations = only_integrations.map(&:to_s)
+      integrations_to_load = Raven::AVAILABLE_INTEGRATIONS & only_integrations
+      not_found_integrations = only_integrations - integrations_to_load
+      if not_found_integrations.any?
+        self.logger.warn "Integrations do not exist: #{not_found_integrations.join ', '}"
+      end
+      integrations_to_load &= Gem.loaded_specs.keys
```
nateberkopec: TIL

nateberkopec:
Awesome work. Stellar Ruby as well.

PR #361, opened 2015-09-02
sanitize should match strictly on whole words.

alyssa:
We have a Rails app and one of our `filter_parameters` fields is name. Because of the loose matching on words, fields such as `server_name` get masked.

### DIFF
--- a/lib/raven/processor/sanitizedata.rb
+++ b/lib/raven/processor/sanitizedata.rb
@@ -5,6 +5,7 @@ module Raven
     INT_MASK = 0
     DEFAULT_FIELDS = %w(authorization password passwd secret ssn social(.*)?sec)
     CREDIT_CARD_RE = /^(?:\d[ -]*?){13,16}$/
+    REGEX_SPECIAL_CHARACTERS = %w(. $ ^ { [ ( | ) * + ?)
 
     attr_accessor :sanitize_fields, :sanitize_credit_cards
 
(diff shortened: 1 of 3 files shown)


dcramer:
This might make sense for filter_parameters, but we definitely don't want to allow anything through that contains "password" or "passwd" etc, so those should retain CONTAINS matches.

alyssa:
@dcramer good point. updated so that it doesn't do strict matching on default list.

lib/raven/processor/sanitizedata.rb
```diff
@@ -5,6 +5,7 @@ class Processor::SanitizeData < Processor
     INT_MASK = 0
     DEFAULT_FIELDS = %w(authorization password passwd secret ssn social(.*)?sec)
     CREDIT_CARD_RE = /^(?:\d[ -]*?){13,16}$/
+    REGEX_SPECIAL_CHARACTERS = %w(. $ ^ { [ ( | ) * + ?)
```
xaviershay: alternate suggestion: allow regexes directly in `DEFAULT_FIELDS | sanitize_fields`. That way we don't need to guess (and you can still keep backwards compat).

alyssa: We want to strictly match on whole words such as "name" and "foo_bar", which do not include `REGEX_SPECIAL_CHARACTERS `, in `sanitize_fields`. So, if we allow regexes in both, we would still have the same check?

xaviershay: ah, I see, you can't keep backwards compat with my suggestion. You'd have to specify them as `/\Aname\Z/` instead of `"name"`. Kinda weird.

So my recommendation is probably:

1. Merge this as is, release in next point release.
2. Next major version, switch to doing exact matching of strings and allowing regexes. It's backwards incompatible, but probably less surprising.

alyssa:
@dcramer thoughts on merging this?

lib/raven/processor/sanitizedata.rb
nateberkopec: I agree with this plan.

PR #365, opened 2015-09-10
Expose more debugging info to active_job integration

tonywok:
Aims to address https://github.com/getsentry/raven-ruby/issues/364

Although, now that I'm running into the issue alluded to here in the raven-ruby codebase: https://github.com/getsentry/raven-ruby/blob/master/lib/raven/integrations/rails/active_job.rb#L9 where it's capturing two exceptions - one for active job and one for resque...


nateberkopec:
Neat!

PR #367, opened 2015-09-17
Fix activejob integration

greysteil:
Looks like this was developed against Rails 5, which isn't released yet. Until then we don't have access to `provider_job_id`, so the existing code will error out during the `Raven.capture_exception` call.


greysteil:
@nateberkopec / @dcramer - would love a speedy 0.15.1 release for this, if you're :+1:. With the current setup I'm pretty sure ActiveJob errors won't be making it into Sentry.

greysteil:
cc @tonywok (patch looks awesome otherwise :heart:)

dcramer:
Looks sane to me. Thanks!

dcramer:
0.15.1 on rubygems

greysteil:
Wow, you work fast @dcramer - thanks!

tonywok:
@dcramer Good catch! I was digging against master without thinking...

PR #374, opened 2015-10-03
Remove deprecated methods

greysteil:
Looks like these were scheduled for deletion in 0.14.

### DIFF
--- a/lib/raven/base.rb
+++ b/lib/raven/base.rb
@@ -85,12 +85,6 @@ module Raven
       client.send_event(event)
     end
 
-    def send(event)
-      Raven.logger.warn "DEPRECATION WARNING: Calling #send on Raven::Base will be \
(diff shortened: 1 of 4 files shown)


spec/raven/integrations/rack_spec.rb
```diff
@@ -102,7 +102,7 @@
     end
 
     stack = Raven::Rack.new(Rack::Lint.new(app))
-    expect { stack.call(env) }.to_not raise_error(Rack::Lint::LintError)
```
nateberkopec: Can you remove this, just because it's unrelated?

nateberkopec:
Can't even keep up with my own deprecation schedule!

greysteil:
Updated!

nateberkopec:
Kaboom.

PR #375, opened 2015-10-05
Don't specify error type when speccing that a method doesn't raise

greysteil:
That.


nateberkopec:
Thanks. This test is sort of silly now, but we'll leave that discussion for another day.

PR #377, opened 2015-10-15
Attempt to pull release version from heroku, then from git repo

iloveitaly:


lib/raven/configuration.rb
```diff
@@ -127,6 +127,12 @@ def initialize
       self.sanitize_fields = []
       self.sanitize_credit_cards = true
       self.environments = []
+
+      self.release = ENV['HEROKU_SLUG_COMMIT']
+
+      if self.release.nil? || self.release.empty?
+        self.release = `git rev-parse --short HEAD`.strip
```
dcramer: The ENV change is good, but otherwise this assumes too much (version could be mercurial or something unrelated to VCS). Mind just removing this bit?

iloveitaly: If the app isn't versioned with git, ``git rev-parse --short HEAD`.strip`` will gracefully fallback to an empty string. I can go ahead and remove this, but keeping it in place might be a great added feature for devs with no harmful failure case. Thoughts?

dcramer: It's also the fact that git may not even be available on the system -- unless you're saying thats safe (not a rubyist myself)

lib/raven/configuration.rb
```diff
@@ -127,6 +127,12 @@ def initialize
       self.sanitize_fields = []
       self.sanitize_credit_cards = true
       self.environments = []
+
+      self.release = ENV['HEROKU_SLUG_COMMIT']
+
+      if self.release.nil? || self.release.empty?
+        self.release = `git rev-parse --short HEAD`.strip rescue nil
```
iloveitaly: Good call on `git` not existing on the system. I wrapped the sys call in a rescue: this will fail gracefully if the user doesn't have git installed, or if the current directory is not a git repo.

lib/raven/configuration.rb
nateberkopec: Agreed on needing a git fallback, 99% of Ruby devs use it but perhaps not guaranteed in production environments.

lib/raven/configuration.rb
nateberkopec: I guess in that case `release` will be nil. Anything else we can fall back to? I guess not.

iloveitaly: Yup, that's right: if the user isn't on heroku, and isn't using git, then `release == nil`. Although not idea, I think this solution will auto-tag errors with commit in the majority of cases.

seanlinsley:
It'd also be good to grab the release info from Capistrano deploys. This is how we're doing it:
```ruby
config.release = File.read('REVISION').strip if File.exists? 'REVISION'

iloveitaly:
> It'd also be good to grab the release info from Capistrano deploys

Good idea. Added this to the PR.

dcramer:
lgtm unless anyone has any objections

nateberkopec:
Need to fix the dang build

seanlinsley:
Gem::InstallError: rack-cache requires Ruby version >= 1.9.3.

nateberkopec:
Yeah, PR's welcome to fix that part of the gem spec

PR #378, opened 2015-10-15
Active Job w/ Delayed Job Backend

kkumler:
When an exception is caught from delayed job, but was sent via the Active Job backend, other information is available to fetch.

There were situations where only *last_error*, *locked_at*, and *queue* were in the additional data. Depending on point of failure, some of the active_job data may have been included.
This adds the *active_job* data, which may include *arguments*, *job_class*, the Active Job *job_id*, *locale*, and *queue_name*. The Active Job job_id is most useful, as it is included in the default tagged logging.

### DIFF
--- a/lib/raven/integrations/delayed_job.rb
+++ b/lib/raven/integrations/delayed_job.rb
@@ -12,27 +12,30 @@ module Delayed
 
           rescue Exception => exception
             # Log error to Sentry
+            extra = {
+              :delayed_job => {
+                :id          => job.id,
+                :priority    => job.priority,
+                :attempts    => job.attempts,
+                :handler     => job.handler,
+                :last_error  => job.last_error,
+                :run_at      => job.run_at,
+                :locked_at   => job.locked_at,
+                :locked_by   => job.locked_by,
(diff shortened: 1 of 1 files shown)


nateberkopec:
Won't this remove data for DelayedJob users *without* ActiveJob?

kkumler:
The previous delayed_job hash is still there, at the same level of the extra data sent to Sentry.
An example of the extra data sent (as pulled from the json sent to Sentry):

```json
"extra": {
    "active_job": {
        "arguments": [
            {
                "_aj_globalid": "gid://appname/Instance/27"
            }
        ],
        "job_class": "FailingJob",
        "job_id": "4e74c7ac-b309-470a-b6c4-6778642364d6",
        "locale": "en",
        "queue_name": "default"
    },
    "delayed_job": {
        "locked_at": "Thu, 15 Oct 2015 22:56:34 UTC +00:00",
        "queue": "default"
    }
}
```

kkumler:
Rebased from current master, which included the Ruby 1.8.7 test fix.

PR #379, opened 2015-10-16
configuration.unsent_event for Unsent Events

abhillman:
Thank you kindly for considering this PR. This PR addresses https://github.com/getsentry/raven-ruby/issues/376. It does this by implementing a configuration parameter `unsent_event` which is a Proc that is called if:

* the client is in a failure state and is not sending messages
* the server, when using the HTTPTransport, responds with a non-200 status code
* when the transport raises a runtime exception (HTTP, UDP, or any other transport)

Its semantics are similar to `async` — i.e. the default value for the parameter is `false` and we do some checking on the parameter just like `async` to make sure the value is `false` or a callable. These changes are near-robust. I spent quite a long time playing with writing tests in `raven_spec.rb`, but ran into some roadblocks and could use a little help if any maintainers have a spare moment.

Thank you again kindly for considering this PR. This enhancement will be greatly helpful for my active use of raven-ruby and Sentry.

### DIFF
--- a/lib/raven/client.rb
+++ b/lib/raven/client.rb
@@ -23,12 +23,15 @@ module Raven
 
     def send_event(event)
       return false unless configuration_allows_sending
+      unsent_event = configuration.unsent_event
 
       # Convert to hash
       event = event.to_hash
 
       if !@state.should_try?
-        Raven.logger.error("Not sending event due to previous failure(s): #{get_log_message(event)}")
+        message = "Not sending event due to previous failure(s): #{get_log_message(event)}"
+        unsent_event.call(event, message)  if unsent_event.respond_to?(:call)
+        Raven.logger.error(message)
         return
       end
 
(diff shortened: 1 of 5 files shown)


lib/raven/client.rb
```diff
@@ -39,8 +42,12 @@ def send_event(event)
       begin
         transport.send_event(generate_auth_header, encoded_data,
                        :content_type => content_type)
+      rescue Raven::Transports::NonFatalConnectionError => e
```
nateberkopec: Why is this addition necessary?

lib/raven/configuration.rb
```diff
@@ -165,6 +169,13 @@ def async=(value)
 
     alias_method :async?, :async
 
+    def unsent_event=(value)
+      raise ArgumentError.new("unsent_event must be callable (or false to disable)") unless (value == false || value.respond_to?(:call))
+      @unsent_event = value
+    end
+
+    alias_method :unsent_event?, :unsent_event
```
nateberkopec: This alias is confusing and should be removed.

nateberkopec:
If you rebase against master I think your tests should pass.

nateberkopec:
@abhillman ping, had some questions but definitely want to merge this.

nateberkopec:
Closing due to inactivity, feel free to reopen when updated.

PR #383, opened 2015-10-28
Ignore Sinatra::NotFound by default, format for easier diff

drcapulet:


nateberkopec:
Cheers! Thanks for the formatting change as well.

Issue #384, opened 2015-11-03
Issues with ActiveJob (+Sidekiq) reporting errors in jobs with serialized records


lleger:
I'm having issues with Raven reporting errors that happen inside of ActiveJob jobs when Raven is configured to send events in the background. When a job runs that triggers an exception, the background Raven job fires and eventually dies with this message:

> ERROR: !!! ERROR HANDLER THREW AN ERROR !!!
> ERROR: Can't serialize a Hash with reserved key "_aj_globalid"
> ERROR: [edit]/activejob-4.2.4/lib/active_job/arguments.rb:126:in `serialize_hash_key'

I successfully reproduced this issues consistently in [this test repository I created](https://github.com/lleger/testsentrybg). It's a vanilla Rails 4.2.4 app, and all of the gems are up-to-date. I then added Raven and configured it to report in development and to report via background job. I also added Sidekiq. To exhibit the issue, perform the following steps:

1. Clone the repo
2. Migrate and seed the database
3. Boot Sidekiq: `bundle exec sidekiq`
4. Open up Rails console and run:
  - `ExceptionJob.perform_later` — successfully reports to Sentry
  - `ExceptionJob.perform_later([])` — successfully reports to Sentry
  - `ExceptionJob.perform_later(Post.first)` — failure

My assumption is that it fails because of the ActiveJob serialization. When the Sentry reporter job gets run, it serializes a hash that includes the parameters sent to ActiveJob. You can see in [ActiveJob's source](https://github.com/rails/rails/blob/d5890bdf660f670271a8f60bd091bd11db4c23b3/activejob/lib/active_job/arguments.rb#L123-L127) that they have a `RESERVED_KEYS` hash. Since the parameters sent to ActiveJob would contain a hash that has these keys, it would make sense that that is causing the issue.

I also tested this against the ActiveJob inline adapter, but it didn't seem to have the same issues. I only tested it against Sidekiq, which is what I was using when I noticed the issue.

mattrobenolt:
Hey @lleger, just wanted to follow up and let you know that I'm going to be diving into this this week.

I'm not familiar with Rails or Sidekick personally, so I highly appreciate the steps to reproduce.

I'll report back with findings. :)

lleger:
Thanks for the update. Happy to help diagnose if I can. I only tried with Sidekiq, but I would think the problem would exhibit itself in any of the adapters that serialize to a data store.

mattrobenolt:
[...]

PR #386, opened 2015-11-10
Strip ActiveJob keys to prevent recursion

mattrobenolt:
Once an ActiveJob is queued, ActiveRecord references get serialized into
some internal reserved keys, such as _aj_globalid.

The problem is, if this job in turn gets queued back into ActiveJob with
these magic reserved keys, ActiveJob will throw up and error. We want to
capture these and mutate the keys so we can sanely report it.

Fixes GH-384

### DIFF
--- a/lib/raven/integrations/sidekiq.rb
+++ b/lib/raven/integrations/sidekiq.rb
@@ -23,6 +23,35 @@ if Sidekiq::VERSION < '3'
   end
 else
   Sidekiq.configure_server do |config|
-    config.error_handlers << Proc.new {|ex,context| Raven.capture_exception(ex, :extra => {:sidekiq => context}) }
+    config.error_handlers << Proc.new { |ex, context|
+      Raven.capture_exception(ex, :extra => {
+        :sidekiq => filter_context(context),
+      })
+    }
   end
+end
+
+def filter_context(context)
+  case context
+    when Array
+      context.map { |arg| filter_context(arg) }
+    when Hash
+      Hash[context.map { |key, value| filter_context_hash(key, value) }]
+    else
+      context
+    end
+end
+
+def filter_context_hash(key, value)
+  # Strip any `_aj` prefixes from keys.
+  # These keys come from an internal serialized object from ActiveJob.
+  # Internally, there are a subset of keys that ActiveJob references, but
+  # these are declared as private, and I don't think it's wise
+  # to keep chasing what this list is. But they all use a common prefix, so
+  # we want to strip this becuase ActiveJob will complain.
+  # e.g.: _aj_globalid -> _globalid
+  if key[0..3] == '_aj_'
+    key = key[3..-1]
+  end
+  [key, filter_context(value)]
 end


seanlinsley:
From #384:

> I'm having issues with Raven reporting errors that happen inside of ActiveJob jobs when Raven is configured to send events in the background.

It seems like the better option is to disable background Raven event sending when inside of any given background worker (ActiveJob, Resque, Sidekiq, etc.)

Especially when the only other option is to special-case the removal of ActiveJob keys.

seanlinsley:
As far as the implementation in this PR, shouldn't this code be abstracted out to ActiveJob, since it's not Sidekiq that's causing the bug?

lib/raven/integrations/sidekiq.rb
```diff
@@ -23,6 +23,35 @@ def call(_worker, msg, _queue)
...
+  # Strip any `_aj` prefixes from keys.
+  # These keys come from an internal serialized object from ActiveJob.
+  # Internally, there are a subset of keys that ActiveJob references, but
+  # these are declared as private, and I don't think it's wise
+  # to keep chasing what this list is. But they all use a common prefix, so
+  # we want to strip this becuase ActiveJob will complain.
+  # e.g.: _aj_globalid -> _globalid
+  if key[0..3] == '_aj_'
+    key = key[3..-1]
+  end
```
seanlinsley: Wouldn't it be better to remove the keys entirely? If the ActiveJob job to send this event to Sentry fails, the metadata will be overridden if it's re-queued, so at the very least it's not reliable data.

mattrobenolt:
> It seems like the better option is to disable background Raven event sending when inside of any given background worker (ActiveJob, Resque, Sidekiq, etc.)

This is probably true. I'll have to dig around and figure out how that works.

> As far as the implementation in this PR, shouldn't this code be abstracted out to ActiveJob, since it's not Sidekiq that's causing the bug?

I dug into the `active_job.rb`, but it didn't seem to be stemming from there. I'll poke at it some more and see if I can come up with something.

I'm pretty new to ruby, and trying to get familiar with these things. :)

lib/raven/integrations/sidekiq.rb
mattrobenolt: I wasn't sure if it'd be safe to strip them entirely or not. That was my initial thought, but decided it'd probably be better to err on the side of preserving as much data as possible.

nateberkopec:
> As far as the implementation in this PR, shouldn't this code be abstracted out to ActiveJob, since it's not Sidekiq that's causing the bug? 

I think I disagree. The bug is that we're creating an object in a forbidden way. That's our problem.

lib/raven/integrations/sidekiq.rb
```diff
@@ -23,6 +23,35 @@ def call(_worker, msg, _queue)
...
+    when Array
+      context.map { |arg| filter_context(arg) }
+    when Hash
+      Hash[context.map { |key, value| filter_context_hash(key, value) }]
+    else
+      context
+    end
+end
+
+def filter_context_hash(key, value)
```
nateberkopec: Is there any way we can access the `RESERVED_KEYS` constant directly here? (I think that's what it's called)

nateberkopec: That'll dedup the knowledge of what a "forbidden" key is.

mattrobenolt: I wanted to, but it appears to be private.

https://github.com/rails/rails/blob/master/activejob/lib/active_job/arguments.rb#L130-L135

mattrobenolt: So I err'd on the side of being liberal and catching `_aj_` prefix rather than duping the list in here and worrying about maintaining that.

nateberkopec:
I think we're on the right track, but it should live in [active_job.rb](https://github.com/getsentry/raven-ruby/blob/gh-384/lib/raven/integrations/rails/active_job.rb)

lib/raven/integrations/sidekiq.rb
nateberkopec: I suppose the only truly maintainable solution in the long term would be to `rescue` SerializationErrors. Hm.

lleger:
This looks great. Looking forward to deploying this in our app so we can send in the background again. Any idea when you're planning to have this merged and released?

lib/raven/integrations/sidekiq.rb
nateberkopec: @seanlinsley's right, this is internal private state that's likely wrong/unreliable anyway. We should just drop them.

nateberkopec:
> As far as the implementation in this PR, shouldn't this code be abstracted out to ActiveJob, since it's not Sidekiq that's causing the bug?

It took me a little while to understand this, but the reason this code is in the Sidekiq integration is because we don't load the ActiveJob integration when using Activejob with Sidekiq. This is because the Sidekiq integration gives us better access to the Sidekiq `context` hash, which sometimes provides some nice extra data on where/how the error occurred.

So, this is bug happens when all of these conditions are met:

* Using Sidekiq
* Using ActiveJob
* Using Raven-Ruby with an `async` config
* Your exception involved an object with a [globalid](https://github.com/rails/globalid).

This bug doesn't happen with the ActiveJob integration because we're much more stingy with the data we pass along in that integration. With Sidekiq, we just pass everything (like we should - more data is best when debugging).

nateberkopec:
Decided I couldn't think of a better solution, so better to merge a fix. Had to update for style: a96d47babcb0df4d2bc7

mattrobenolt:
:) thx homie

lleger:
Confirmed with the test app I created for the issue that this fix worked. What is the release ETA for this version of the gem? Would like to update and get it running on my apps. Thanks again guys! :clap:

nateberkopec:
@lleger As soon as we figure out #422

PR #391, opened 2015-12-07
Rubocop base implementation with TODOS file

ddrmanxbxfr:
This PR tries to provide a initial configuration of rubocop.

It adds the rubocop checks on CI run.

Non compliant files are ignored currently in the .rubocop_todos.yml file. 

We could fix the code styling issues slowly until we get a fully compliant codebase :).

Thanks for looking into it.

Issue ref : https://github.com/getsentry/raven-ruby/issues/362


Gemfile
```diff
@@ -1,11 +1,11 @@
-source "https://rubygems.org/"
+source 'https://rubygems.org/'
```
nateberkopec: Can you remove these changes? It looks like StringLiterals is set to `false` anyway, it shouldn't matter right?

(In any case, I'd rather enforce always-double-quote).

nateberkopec:
Looks good in general. See my comment about string literals in Gemfile. Need to think on my sins as to why our config file is so long ;)

Gemfile
ddrmanxbxfr: Done!

Old habit of converting them, d'oh.

ddrmanxbxfr:
Thanks for replying pretty fast.

I've reverted the nits in Gemfile.

nateberkopec:
@ddrmanxbxfr Thanks for your contribution!

PR #403, opened 2015-12-09
Avoid single-line function definitions.

pmbrent:


nateberkopec:
I'd like to keep `okjson.rb` as close to their version as possible. We should exclude `okjson.rb` from all Rubocop checks.

PR #409, opened 2015-12-09
Remove unused argument method + blocks

ddrmanxbxfr:
Mark all the unused stuff with the lodash prefix, so we know that they are not used.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -26,22 +26,6 @@ Lint/UnreachableCode:
   Exclude:
     - 'spec/raven/raven_spec.rb'
 
-# Offense count: 4
-# Cop supports --auto-correct.
-# Configuration parameters: IgnoreEmptyBlocks.
-Lint/UnusedBlockArgument:
-  Exclude:
-    - 'spec/raven/integrations/rack_spec.rb'
-    - 'spec/raven/raven_spec.rb'
-
-# Offense count: 2
-# Cop supports --auto-correct.
-# Configuration parameters: AllowUnusedKeywordArguments, IgnoreEmptyMethods.
-Lint/UnusedMethodArgument:
-  Exclude:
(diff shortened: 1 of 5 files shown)


.rubocop_todo.yml
```diff
@@ -26,22 +26,6 @@ Lint/UnreachableCode:
...
-  Exclude:
-    - 'spec/raven/integrations/rack_spec.rb'
-    - 'spec/raven/raven_spec.rb'
-
-# Offense count: 2
-# Cop supports --auto-correct.
-# Configuration parameters: AllowUnusedKeywordArguments, IgnoreEmptyMethods.
-Lint/UnusedMethodArgument:
-  Exclude:
-    - 'lib/raven/backports/uri.rb'
```
nateberkopec: I'd prefer not to touch anything in the backports folder.

Please add `'lib/raven/backports/uri.rb'` to `rubocop.yml` as a globally ignored file, and then drop the changes to that file from this PR.

spec/raven/raven_spec.rb
```diff
@@ -30,7 +30,7 @@
 
     around do |example|
       prior_async = Raven.configuration.async
-      Raven.configuration.async = lambda { |event| :ok }
+      Raven.configuration.async = lambda { |_event| :ok }
```
nateberkopec: instead of a lambda with an unused argument, let's use a Proc:

```ruby
Raven.configuration.async = Proc.new { :ok }
```

.rubocop_todo.yml
ddrmanxbxfr: Ok, on it !

ddrmanxbxfr:
@nateberkopec Requested changes are done !

I've also favored proc over lambda in  spec/raven/integrations/rack_spec.rb.

Thanks for looking into it.

PR #411, opened 2015-12-09
Favor fail over raise on ruby exceptions

ddrmanxbxfr:


nateberkopec:
Not a huge fan of this style, and it especially doesn't really make sense for an exception handling gem (where most of these things aren't really "failures").

Instead, can you submit a PR adding a new cop to `rubocop.yml` that enforces the style of using only `raise`?

PR #413, opened 2015-12-09
Remove semicolons

wdhorton:
Fix style: semicolons shouldn't be used to terminate expressions. @nateberkopec you owe me a PR on [whales](https://github.com/wdhorton/whales) now

### DIFF
--- a/lib/raven/processor/removecircularreferences.rb
+++ b/lib/raven/processor/removecircularreferences.rb
@@ -4,7 +4,10 @@ module Raven
       return "(...)" if visited.include?(v.__id__)
       visited += [v.__id__]
       if v.is_a?(Hash)
-        v.reduce({}) { |memo, (k, v_)| memo[k] = process(v_, visited); memo }
+        v.reduce({}) do |memo, (k, v_)|
+          memo[k] = process(v_, visited)
+          memo
+        end
       elsif v.is_a?(Array)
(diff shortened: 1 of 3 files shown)


lib/raven/processor/removecircularreferences.rb
```diff
@@ -4,7 +4,10 @@ def process(v, visited = [])
       return "(...)" if visited.include?(v.__id__)
       visited += [v.__id__]
       if v.is_a?(Hash)
-        v.reduce({}) { |memo, (k, v_)| memo[k] = process(v_, visited); memo }
```
nateberkopec: I think `each_with_object` would be more appropriate here.

```ruby
  v.each_with_object({}) { |(k, v_), memo| memo[k] = process(v_, visited) }
```

lib/raven/processor/sanitizedata.rb
```diff
@@ -16,7 +16,10 @@ def initialize(client)
     end
 
     def process(value)
-      value.inject(value) { |memo,(k,v)| memo[k] = sanitize(k,v); memo }
```
nateberkopec: Same here.

```ruby
value.each_with_object(value) { |(k,v), memo| memo[k] = sanitize(k,v) }
```

nateberkopec:
Also, please squash into a single commit

PR #414, opened 2015-12-17
Allow fixnums as hash key

michaelsauter:
I got the error "Hash key is not a string" for a number in one of my apps during encoding of the data to be sent. I am not entirely sure where it came from since I wasn't able to reproduce that problem locally. My current assumption is that it was some key set during `Raven.user_context`.

While it could be said that the application is to blame for this, I think Sentry should be more forgiving. After all, Sentry is meant to collect errors, not produce any :) Sentry should do whatever it can to report errors in my application, no matter what is being thrown at it. In that light, I wonder whether "okjson" is the right way to do JSON encoding at all. From their repo:

> Note for people who've used other ruby json
libraries: json does not represent symbols,
so okjson does not encode them. It won't
convert them to strings; that would silently
corrupt your data. The same holds for all of
ruby's other data types that don't exist in
json, including Range, Regexp, Proc, Time,
and user-defined classes.

Sentry should not be concerned with "corrupting data". Already Sentry is handling symbols in the vendored version since this is so common in Ruby. I think you're going to fight "upstream" okjson quite a bit going forward ...

### DIFF
--- a/lib/raven/okjson.rb
+++ b/lib/raven/okjson.rb
@@ -448,6 +448,7 @@ private
     case k
     when String then strenc(k)
     when Symbol then strenc(k.to_s)
+    when Fixnum then strenc(k.to_s)
     else
       raise Error, "Hash key is not a string: #{k.inspect}"
     end
--- a/spec/raven/okjson_spec.rb
+++ b/spec/raven/okjson_spec.rb
@@ -1,21 +1,27 @@
 require 'spec_helper'
 
 describe Raven::OkJson do
-  ['foo', :foo].each do |obj|
-    it "works with #{obj.class} keys" do
-      expect(Raven::OkJson.encode(obj => 'bar')).to eq '{"foo":"bar"}'
+  data = [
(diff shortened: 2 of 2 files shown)


nateberkopec:
Yeah, we've fought JSON before - unfortunately using something like multi_json or even just the built-in JSON causes us more problems for the reasons you describe - we need to be more forgiving. At least with OKJSON we have a recourse (modifying the implementation).

nateberkopec:
@michaelsauter Can you add a test for this? I'm 70% sure I'm OK with this but it will need a test.

michaelsauter:
@nateberkopec Thanks for commenting so quickly. I can look into that, but won't get to it until next week. Will update on Monday.

michaelsauter:
@nateberkopec Added some tests. Thought I'd restructure the original test a bit so it is easier to extend it with other types if there's a need to add more later.

nateberkopec:
Oh neat, I like the additions. Thank you!

Issue #388, opened 2015-11-19
Rails/Rack recording exceptions twice


mattrobenolt:
Repository to test: https://github.com/mattrobenolt/sentry_test

When an exception occurs, we get two exceptions, one for the page itself, then one for the corresponding `/500` route.

As noted by a user:

> So it looks like that even new Rails app is affected. We think that it’s caused by catching exception in a raven rails integration first and then the second exception is caught by raven rack integration. It can be avoided by skipping rack integration entirely but we don’t want to do that. So I think that there should be some kind of mechanism to catch the exception only once and by only by one integration.

Refs: https://sentry.zendesk.com/agent/tickets/1054

dcramer:
It's unclear what the right solution to this is. Would appreciate anyone familiar w/ Rack/Rails ecosystem's opinion.

mattrobenolt:
Hey @martinhladil, just following up here.

I clone this repository, and was unable to reproduce. I was running with just `rails server`. Is there something I should be doing differently?

Could this be something with running behind Unicorn?

martinhladil:
We can reproduce this on the new Rails app in it's default state without any modifications etc. so the test repository is actually not needed.

It's not Unicorn related as this happens on Webrick and Passenger.

We think that the reason of the duplication is quite simple. Exception is catched and reported twice. Once for Rails integration and once for Rack integration. We can solve it by manually including Rails integration only and skipping Rack integration but this is not acceptable as we would miss Rack only errors.

mattrobenolt:
Yeah, I understand. I just haven't been able to see this myself in testing to be able to begin to fix.

I've tested against a brand new Rails app, as well as your repository without success. :(

martinhladil:
Hmm, I would really like to help you, but I don't have a clue we are doing different. We know for sure that it's not related to a webserver, ruby or Rails version. We are not using any custom Raven configuration. DSN is set as an ENV variable.

mattrobenolt:
If you're familiar with Docker, I've been doing this inside a container, and I forked and updated your project to include a `Dockerfile`.

You should be able to simply:

```bash
git clone https://github.com/mattrobenolt/sentry_test.git
cd sentry_test
docker build --rm -t sentry-test .
docker run --rm --name sentry-test -e SENTRY_DSN=<dsn> -p 3000:3000 -v $PWD:/usr/src/app -it sentry-test
```

mattrobenolt:
[...]

PR #422, opened 2015-12-21
WIP: Fix Rails multi-exception reports

nateberkopec:


nateberkopec:
ping @mattrobenolt 

This fixes the issue for the app you provided in #388, and, as far as I can tell, doesn't break any additional use cases. Can you take a look?

nateberkopec:
Can also confirm this works correctly with `rescue_from` - errors which are not re-raised by `rescue_from` are not reported.

greysteil:
:+1: :heart:

mattrobenolt:
I tested this locally, and it seems to record the wrong one now.

It only logged one event, but it only logged the event wiht the url of `/500`, not the original url. So seems like we're filtering out the wrong one.

mattrobenolt:
For posterity, I ran against my docker container:

```
$ docker run --rm --name=sentry-test -e RAILS_ENV=production -e SENTRY_DSN=https://xxx:xxx@app.getsentry.com/3825 -p 3000:3000 -v $PWD:/usr/src/app -v /Users/matt/code/getsentry/raven-ruby:/usr/local/bundle/gems/sentry-raven-0.15.2 -it sentry-test
```

And mounted in this PR.

Without mounting this in, it was recording the error twice. Now only records once, but the wrong one.

nateberkopec:
For posterity: I worked this out with @mattrobenolt and made sure the right exception was recorded.

PR #423, opened 2015-12-21
Rails 5 support

nateberkopec:


nateberkopec:
Blocked by a bug in Rails 5, waiting for that fix to be merged and beta2 released.

boblin:
@nateberkopec is there any way how to use sentry with 5.0.0.beta1 now?

nateberkopec:
@boblin I believe it does work, just with a deprecation error.

boblin:
I see. Thank you. I'll try it.

boblin:
I've tried, but it doesn't work for me. It appears that it sends data, but I don't see anything catched on app.getsentry.com.  I've tried also testing rake task with the same result.

Elektron1c97:
I think you should fix 
```
DEPRECATION WARNING: alias_method_chain is deprecated. Please, use Module#prepend instead. From module, you can access the original method using super.
``` too.

**Comment by nateberkopec** [comment]
.beta2 is out, I'll update.

**Comment by nateberkopec** [comment]
Link #456

File: lib/raven/integrations/rails/middleware/debug_exceptions_catcher.rb [hunk 1]
```diff
@@ -6,8 +6,10 @@ def self.included(base)
           base.send(:alias_method_chain, :render_exception, :raven)
         end
 
-        def render_exception_with_raven(env, exception)
+        def render_exception_with_raven(env_or_request, exception)
+          env = env_or_request.respond_to?(:env) ? env_or_request.env : env_or_request
           Raven::Rack.capture_exception(exception, env)
+        ensure
           render_exception_without_raven(env, exception)
```
senny: shouldn't you be passing the original argument `env_or_request` instead of `env`?

nateberkopec: herrrrp thanks @senny

PR #430, opened 2015-12-30
Cleanup Rack interface

nateberkopec:


lib/raven/integrations/rack.rb
```diff
@@ -66,4 +66,62 @@ def call(env)
...
+      env_hash.each_with_object({}) do |(key, value), memo|
+        key = key.to_s # rack env can contain symbols
+        value = value.to_s
+        next unless key.upcase == key # Non-upper case stuff isn't either
+        # Rack adds in an incorrect HTTP_VERSION key, which causes downstream
+        # to think this is a Version header. Instead, this is mapped to
+        # env['SERVER_PROTOCOL']. But we don't want to ignore a valid header
+        # if the request has legitimately sent a Version header themselves.
+        # See: https://github.com/rack/rack/blob/028438f/lib/rack/handler/cgi.rb#L29
+        next if key == 'HTTP_VERSION' && value == ENV['SERVER_PROTOCOL']
```
matthewd: @nateberkopec should this be `env_hash`?

nateberkopec: Whoops! It absolutely should be. Looks like my editor autocorrected 'env' to ENV. Writing a test and fixing.

Since you're here, can you weigh in on https://github.com/rack/rack/issues/970, which is the reason this hack exists?

nateberkopec: Fixed in #628

PR #432, opened 2016-01-06
⚠️ Ruby Warnings

amatsuda:
This PR eliminates several Ruby warnings.

`require 'raven'`  has been removed here and there in order to fix the circular require warning, which is totally safe because `require 'raven'` has to be already performed by bundler => lib/sentry-raven.rb.


nateberkopec:
Thank you so much @amatsuda!

amatsuda:
:beers:

Issue #438, opened 2016-01-15
Handle `Not a git repository`


dex1t:
`Raven::Configuration.new` outputs fatal message on non-git project. This is noisy :cry: 
```
fatal: Not a git repository (or any of the parent directories): .git
```

dex1t:
Thanks for the quick fix :sparkles:

nateberkopec:
Fixed in 0.15.4

dex1t:
:+1:

PR #440, opened 2016-01-16
Avoid failure log message if not git repository

nateberkopec:
#438

### DIFF
--- a/lib/raven/configuration.rb
+++ b/lib/raven/configuration.rb
@@ -132,15 +132,7 @@ module Raven
       self.sanitize_credit_cards = true
       self.environments = []
 
-      self.release = ENV['HEROKU_SLUG_COMMIT']
-
-      if self.release.nil? || self.release.empty?
-        self.release = File.read(File.join(::Rails.root, 'REVISION')).strip rescue nil
-      end
-
-      if self.release.nil? || self.release.empty?
-        self.release = `git rev-parse --short HEAD`.strip rescue nil
-      end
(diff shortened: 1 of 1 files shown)


yuku-t:
:+1: 

I really need this change. Because I use crontab's email forwarding feature, whenever a cron task starts a Rails server, I receive a mail with the git error message. :expressionless:

nateberkopec:
I'm not sure I got the `File.directory` call right here, need to test this.

yuku-t:
Checking the existence of ".git" directory makes sense as long as current working directory is the root of a git repository. `git` commands can be used in subdirectories so this patch is not perfect, but I think it's probably sufficient.

As I wrote in #444, redirection is another choice. I checked it works as expected in Linux (*NIX) machine. The problem is that I'm not sure whether it works in Windows.

nateberkopec:
:+1: looks like this works

nateberkopec:
Pushing a teeny release soon

dex1t:
:+1:

michaelsauter:
@nateberkopec Any news on that release? Holding off upgrading from 0.15.2 at the moment.

dosire:
@nateberkopec We would also love to see that release. GitLab users are confused by the message https://gitlab.com/gitlab-org/gitlab-ce/issues/12657

nateberkopec:
Fixed in 0.15.4, released.

michaelsauter:
Thanks!

PR #441, opened 2016-01-16
Fix autocorrectable Cops

nateberkopec:


nateberkopec:
Needs some changes to the cops enabled:

* We have a convention of commas in the last element of a literal, let's just keep that
* I dont understand the changes from alias_method and alias
* The parallel assignment changes are silly
* Array %w should be delimited by brackets, not parens

lib/raven/base.rb
```diff
@@ -227,22 +225,22 @@ def inject_only(*only_integrations)
...
+    alias_method :annotate, :annotate_exception
 
     private
 
     def install_at_exit_hook(options)
       at_exit do
-        if $!
-          logger.debug "Caught a post-mortem exception: #{$!.inspect}"
-          capture_exception($!, options)
+        if (exc = $!) # rubocop:disable Style/SpecialGlobalVars
```
olleolleolle: You always have the option of including [English.rb](https://github.com/ruby/ruby/blob/trunk/lib/English.rb#L23) to have `$ERROR_INFO` handy.

That file's been with us a _long_ while.

nateberkopec: unfortunately that seems to cause a bug in this particular instance. I haven't had time to figure out why yet.

nateberkopec:
Closing until I move forward on #442

PR #444, opened 2016-01-19
Stop writing a git error message to stderr

yuku-t:
Executing `git rev-parse` out of git repository writes

    fatal: Not a git repository (or any of the parent directories): .git

to stderr.


yuku-t:
Oh, I missed #440

PR #450, opened 2016-02-01
Detect release before application initializers run

eugeneius:
The release can be set manually by the user. The typical way to do that would be to create an initializer:

```ruby
# config/initializers/raven.rb
Raven.configure do |config|
  config.release = ENV['RELEASE']
end
```

The hook in the Rails integration that tries to automatically detect a release is run after the app's initializers, which means it clobbers any value that has already been set - even if it can't detect a release, it sets the option back to nil.

If the user has set a release themselves, we shouldn't try to detect it.

### DIFF
--- a/lib/raven/integrations/rails.rb
+++ b/lib/raven/integrations/rails.rb
@@ -13,7 +13,7 @@ module Raven
       end
     end
 
-    config.after_initialize do
+    config.before_initialize do
       Raven.configure do |config|
         config.logger ||= ::Rails.logger
         config.project_root ||= ::Rails.root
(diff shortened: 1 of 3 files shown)


nateberkopec:
Ack, that's definitely not intended.

The problem is with a Cap-deployed app with a .git folder. In that case, when the client is initialized, we set the `release` to the short Git SHA ("asdaf" or whatever). `project_root` only gets properly set when *Rails* is initialized, which happens *after* the client is initialized, so that's why I was re-running detect_release with a bare `=`.

eugeneius:
Okay, I've taken a different approach here: we now set `project_root` and try to detect the release _before_ the application initializers are run.

The `before_initialize` hook was [added in Rails 3.0](https://github.com/rails/rails/blob/v3.0.0/railties/lib/rails/railtie/configuration.rb#L48-L50), which is the oldest version we support. Both `Rails.logger` and `Rails.root` are available when it runs.

nateberkopec:
Thanks!

PR #451, opened 2016-02-01
Fix build failures on Rails 3.2 and Ruby 1.8

eugeneius:
The latest release of Rails 3.2 accidentally introduced some Ruby 1.9 hash syntax, making it incompatible with Ruby 1.8: https://github.com/rails/rails/issues/23248

It will be fixed in an upcoming release, but until then this will get the build passing again.


nateberkopec:
Cheers!

PR #452, opened 2016-02-01
Makefile: version extractor tweaked

olleolleolle:
:golf: 

On my system, the existing code extracted version as `0.15.4".freez`.

Here, a tweak, which outputs the `0.15.4` for the current value.

Or - Perhaps something like this?

```shell
grep 'VERSION =' lib/raven/version.rb | grep '\d+\.\d+\.\d+' -o -E --color=never
```

Which is the same as:

```shell
grep '\d+\.\d+\.\d+' -o -E --color=never lib/raven/version.rb
```


nateberkopec:
Thank you!

Issue #453, opened 2016-02-03
NoMethodError in delayed_job integration


dcramer:
`/lib/raven/integrations/delayed_job.rb:22`

```
NoMethodError: 
undefined method `[]' for nil:NilClass 
```

Reported that job.last_error is nil

PR #454, opened 2016-02-03
Handle nil last_error in delayed_job (fixes GH-453)

dcramer:


nateberkopec:
We should do the same for `handler` too, even if it can't be nil now maybe it will be in the future.

sebastianvera:
Are you planning to release a new version with this fix soon? :flushed:

PR #457, opened 2016-02-12
Rails 5 deprecation warning fix (Module#prepend)

Elektron1c97:
Remove the deprecation warning, WITH legacy support :pray:
```
DEPRECATION WARNING: alias_method_chain is deprecated. Please, use Module#prepend instead. From module, you can access the original method using super.
```


nateberkopec:
Thank you!

PR #458, opened 2016-02-13
Don't check `catch_debugged_exceptions` setting before user has set it

greysteil:
Looks like a side effect of https://github.com/getsentry/raven-ruby/pull/450 was to remove the usefulness of the `catch_debugged_exceptions` setting. Checking this setting obviously need to be done *after* it's set.


nateberkopec:
Whoops - yup!

eugeneius:
:weary: thanks for fixing this, and apologies for breaking it in the first place.

I'm trying to write a spec to cover this behaviour now, although `Rails::Application` being a singleton is making things difficult.

PR #459, opened 2016-02-14
Check catch_debugged_exceptions option at runtime

eugeneius:
Instead of adding the hook to report debugged exceptions based on the value of the configuration option while Rails is initialising, we can always add the hook, but have it check the option at runtime. This is more in line with how other configuration options are used, and means that the option can be set or changed at any time, which would have prevented #458.

This also lets us write tests that exercise this behaviour, since the logic is no longer tied to the Rails initialisation process, which can only happen once during the test suite.

### DIFF
--- a/lib/raven/integrations/rails.rb
+++ b/lib/raven/integrations/rails.rb
@@ -22,19 +22,17 @@ module Raven
     end
 
     config.after_initialize do
-      if Raven.configuration.catch_debugged_exceptions
-        require 'raven/integrations/rails/middleware/debug_exceptions_catcher'
-        if defined?(::ActionDispatch::DebugExceptions)
-          exceptions_class = ::ActionDispatch::DebugExceptions
-        elsif defined?(::ActionDispatch::ShowExceptions)
-          exceptions_class = ::ActionDispatch::ShowExceptions
-        end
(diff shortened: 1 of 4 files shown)


nateberkopec:
Punting on this for 0.15.6. It feels really weird that we inject our middleware no matter what here.

nateberkopec:
Yeah, I'm not comfortable with this. I think the long term fix for making this testable is probably related to #462, because working with this singleton Raven configuration in our tests is making us bend over backwards. We should fix how we deal with this in the tests rather than the code.

eugeneius:
:+1:

eugeneius:
Thinking about this some more, testing this code is difficult because Rails is a singleton, not Raven - addressing #462 won't make it any easier.

Interestingly, Rails had the same choice to make with `ActionDispatch::ShowExceptions` - [it doesn't do anything](https://github.com/rails/rails/blob/v4.2.5.1/actionpack/lib/action_dispatch/middleware/show_exceptions.rb#L32-L33) unless `env['action_dispatch.show_exceptions']` is set, but it's [always included anyway](https://github.com/rails/rails/blob/v4.2.5.1/railties/test/application/middleware_test.rb#L128-L133).

PR #461, opened 2016-02-16
Return original response in debug middleware patch

### LINKED ISSUE #460: Sentry not working with Rails 4.2.5.1 ? (undefined method `[]=' for nil:NilClass)
When adding Sentry to a new Rails 4.2.5.1 project i get ``` undefined method `[]=' for nil:NilClass ``` after every exception

It makes stack traces harder to read, and worse : 404 errors such as routing errors turn into 500.

Here is a sample stack trace illustrating the problem : https://gist.github.com/vdaubry/b8be537646c4f3536bc8

Here is the list of gems (i only added sentry to a new rails project) :
https://gist.github.com/vdaubry/5511ef807134b3ee3e07

Steps to reproduce :
* Create a new rails 4.2.5.1 project
* Add sentry gem

Edit :
If i remove the ``` sentry-raven ``` gem the ``` undefined method `[]=' for nil:NilClass ``` disappears

eugeneius:
Fixes #460.

The call to the original `render_exception` method was moved into an ensure block to prevent errors in this module from interfering with Rails' own error handling behaviour. However an ensure block has no effect on the return value of the method, and we were returning the result of `Raven::Rack.capture_exception` instead of the response generated by the original Rails middleware.

### DIFF
--- a/lib/raven/integrations/rails/middleware/debug_exceptions_catcher.rb
+++ b/lib/raven/integrations/rails/middleware/debug_exceptions_catcher.rb
@@ -3,9 +3,11 @@ module Raven
     module Middleware
       module DebugExceptionsCatcher
         def render_exception(env_or_request, exception)
-          env = env_or_request.respond_to?(:env) ? env_or_request.env : env_or_request
-          Raven::Rack.capture_exception(exception, env)
-        ensure
+          begin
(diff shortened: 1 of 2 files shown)


nateberkopec:
> an ensure block has no effect on the return value of the method

TIL

Issue #442, opened 2016-01-16
Drop 1.8.7 support


nateberkopec:
It's time (maybe). Surveys peg 1.8.7 usage at 1% of active applications, and has been EOL for years now. It's the most pain to maintain (I can't even compile it right now).

I'm willing keep 1.9 for now, we don't nearly have to do the same amount of work and backporting to support it.

nateberkopec:
Going for it, since no one commented here. Will include a version bump.

seanlinsley:
As of a year ago today (what a coincidence) Ruby 1.9.3 is no longer supported: https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended/

nateberkopec:
Closed by #465.

Worth noting 2.0.0 went EOL today. TBH there's very little work involved in supporting 1.9 and 2.0 so I'm not worried about dropping those at this time.

PR #465, opened 2016-02-23
Drop Ruby 1.8.7 support

nateberkopec:
See #442


lib/raven/client.rb
```diff
@@ -104,11 +104,7 @@ def generate_auth_header
     end
 
     def strict_encode64(string)
-      if Base64.respond_to? :strict_encode64
-        Base64.strict_encode64 string
-      else # Ruby 1.8
-        Base64.encode64(string)[0..-2]
-      end
+      Base64.strict_encode64(string) if Base64.respond_to?(:strict_encode64)
```
seanlinsley: Do you need to keep the `respond_to?` check around? Seems like that's just there for 1.8 compatibility

seanlinsley: It's defined in 1.9.3:
http://ruby-doc.org/stdlib-1.9.3/libdoc/base64/rdoc/Base64.html#method-i-strict_encode64

gemfiles/rails32.gemfile
```diff
@@ -1,7 +1,6 @@
 source "http://rubygems.org"
 
 gem "rails", ["~> 3.2.6", "!= 3.2.22.1"]
-gem "rack-cache", "~> 1.2.0" # Ruby 1.8.7 requires this
 gem "i18n", "~> 0.6.0"
```
eugeneius: This was added in #273, with the note:

> Fix Rails 3.2 gemfile for Ruby 1.8

Looks like it can be dropped here too.

sentry-raven.gemspec
```diff
@@ -14,18 +14,18 @@ Gem::Specification.new do |gem|
...
   gem.add_development_dependency "rspec", "~> 3.0"
   gem.add_development_dependency "rspec-rails"
   gem.add_development_dependency "mime-types", "~> 1.16"
-  gem.add_development_dependency "rest-client", "< 1.7.0" if RUBY_VERSION == '1.8.7'
-  gem.add_development_dependency "rest-client" if RUBY_VERSION > '1.8.7'
-  gem.add_development_dependency "timecop", "0.6.1" if RUBY_VERSION == '1.8.7'
-  gem.add_development_dependency "timecop" if RUBY_VERSION > '1.8.7'
+  gem.add_development_dependency "rest-client", "< 1.7.0"
+  gem.add_development_dependency "rest-client"
+  gem.add_development_dependency "timecop", "0.6.1"
```
eugeneius: These pinned dependencies should be dropped, since they only applied for 1.8.7.

nateberkopec:
@eugeneius @seanlinsley Thanks for the review guys!

Issue #445, opened 2016-01-20
Leaking session cookies?!


jarmo:
I've tried to add `sentry` to my `Rails` app and tested it out in `development` environment. For my surprise I saw my session cookie (with value) on sentry's web interface.

I tried to find cookie-specific code or issue, but did not find any - it seems that all headers/cookies are sent.

[Only issue](https://github.com/getsentry/raven-ruby/issues/253) I found was pointing to a non-existing wiki page.

Sentry **should not ever** send out session cookies, but it seems that this is currently the case.

I'm using `sentry` from `master`.

nateberkopec:
> Sentry should not ever send out session cookies, but it seems that this is currently the case.

Sentry's current positon is that we send cookies by default. Filtering session cookies out from this is sort of difficult with the current processor infrastructure, but not impossible. We should probably add processors to the pipeline if using Rails or Rack that deal with common session cookies and filter them out.

jarmo:
Since sentry has an integration for Rails, then discovering session cookie should not be that difficult. Of course it depends how many Rails versions are officially supported etc.

nateberkopec:
Leaving this open because I want to add an optional SessionCookie processor.

nateberkopec:
Changed my mind. If you want to send cookies but *not* send session cookies, write your own processor based on the Cookies one.

jarmo:
Went "unsecure by default" route? That's bad :-1:

nateberkopec:
No cookies are sent at all by default.

jarmo:
That's even better :) This information was not included in here before. Thanks!

PR #466, opened 2016-02-23
Add PostData and Cookies processors by default

### LINKED ISSUE #252: Client should not send form POST data by default
See [spec](http://sentry.readthedocs.org/en/latest/developer/client/index.html#client-usage-end-user).

Also verify we're not sending cookie data by default.

nateberkopec:
Would close #252 and #445.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -14,3 +14,6 @@ Style/SignalException:
 
 Performance/Casecmp:
   Enabled: false
+
+Style/ClassAndModuleChildren:
+  Enabled: false
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -127,17 +127,6 @@ Style/CaseEquality:
 Style/CaseIndentation:
   Enabled: false
 
-# Offense count: 6
-# Configuration parameters: EnforcedStyle, SupportedStyles.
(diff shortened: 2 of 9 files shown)


nateberkopec:
https://docs.getsentry.com/hosted/clientdev/

> Feature based support is required for the following:

> If cookie data is available, it’s not sent by default
> If POST data is available, it’s not sent by default

nateberkopec:
Blocked on decision at https://github.com/getsentry/sentry-docs/pull/30

dcramer:
I think I'd prefer this change. When I added the bit to the docs it was more of "this is how things should work" vs "this is how SDKs are currently implemented".

seanlinsley:
If this ends up being merged, it seems like y'all should notify all Ruby users on Sentry of this change. I'd certainly want to know, instead of wondering why events stopped having potentially vital information to identify the problem.

nateberkopec:
@seanlinsley :+1: and agree

seanlinsley:
What's the versioning policy? I'd consider this (and potentially dropping OkJSON in #480) as worthy of a major version bump.

nateberkopec:
We're still "unstable", so 0.x.0 releases may still contain breaking changes. I try to keep them out of 0.0.x releases.

PR #471, opened 2016-02-26
Expand context docs

dcramer:


dcramer:
I dont know how to rubby but this seems like the current trendy way of calling functions with dicts :)

docs/context.rst
```diff
@@ -36,21 +36,64 @@ generally constists of something like the following:
...
+You can provide a set of key/value pairs called tags which Sentry will index and aggregate. This will help you understand the distribution of issues, as well as enabling easy lookup via search.
+
+.. sourcecode:: ruby
 
     # tag the request with something interesting
-    Raven.tags_context interesting: 'yes'
+    Raven.tags_context {
+      language: 'enUS',
+      timezone: 'PST'
+    }
```
seanlinsley: This isn't valid Ruby, because the curly braces get interpreted as a block. This is valid though:
```ruby
Raven.tags_context language: 'enUS', timezone: 'PST'

Comment (seanlinsley) [inline docs/context.rst]: Same with all the examples you added above this. Here are a couple valid options:
```ruby
Raven.user_context \
  # a unique ID which represents this user
  id: 1,

  # the actor's email address, if available
  email: 'foo@example.com',

  # the actor's username, if available
  username: 'foo',

  # the actor's IP address, if available
  ip_address: '127.0.0.1'

Raven.user_context(
  # a unique ID which represents this user
  id: 1,

  # the actor's email address, if available
  email: 'foo@example.com',

  # the actor's username, if available
  username: 'foo',

  # the actor's IP address, if available
  ip_address: '127.0.0.1'
)

Comment (seanlinsley) [inline docs/context.rst]: But the arguments themselves are a bit confusing, because they're sending static values instead of calling something inside the person's app. This might be better:
```ruby
# Assuming you have a `current_user` method
Raven.user_context id: current_user.id, email: current_user.email
```
For the final argument example, `ip_address`, what's the theoretical value in having that in the documentation? Raven already records the request IP address. Or maybe #466 made it so the IP address is no longer sent by default?

dcramer: Spec + situations where they have a proxy in place. It wasn't clear to me what we did to resolve x-forwarded-for (or other situations) and clarifying that they can simply send the IP makes it nice and simple.

dcramer:
Will fix my terribad Ruby for these examples

dcramer:
p.s. if someone beats me to this and really wants to get it done feel free to take over/claim credit :)

I'm hoping to get it done sometime this week, but this was easier than opening an issue to describe what we needed to improve on.

PR #472, opened 2016-02-26
Remove UDP transport

dcramer:
UDP was removed from the Sentry server several versions back (prior to 8.x).


nateberkopec:
:white_check_mark:

PR #477, opened 2016-03-10
Use Module#prepend only when it exists

RobinDaugherty:
[JRuby 1.7.x does not have a `Module#prepend` method](https://github.com/jruby/jruby/issues/751). #457 made the assumption that if Ruby version is 2.0 or higher, then the `prepend` method exists. This affected me because I'm [using JRuby 1.7 in "2.0 mode"](http://stackoverflow.com/a/15281071/1589422).

It's always better to use duck typing instead of assuming that a version of Ruby has specific methods.

I believe it's correct to treat a missing `prepend` method by using the pre-Ruby 2.0 method (`insert`ing `OldDebugExceptionsCatcher`), and it seems to work correctly. But it may not be the correct behavior.


nateberkopec:
Can you rebase this against the latest `travis.yml`?

.travis.yml
```diff
@@ -21,6 +21,19 @@ notifications:
     on_failure: change
 
 matrix:
+  include:
+    - rvm: jruby-1.7.24
+      env: JRUBY_OPTS="--2.0"
+      gemfile: gemfiles/rails41.gemfile
```
nateberkopec: There is no rails41.gemfile on latest master, these need to be omitted.

RobinDaugherty:
`LoadError: no such file to load -- raven`
Does that failure mean anything to you? I get the feeling that it's not related to a change that I made.

nateberkopec:
Heh, no, I was hoping you would know :laughing:

RobinDaugherty:
Doh! It makes sense...the gemfile doesn't exist, so raven doesn't exist in the bundle.

Elektron1c97:
Wow, thanks for the fix, didn't know that something like JRuby 1.7 in 2.0 mode exists..
(I was the one who made the fix for the `Module#prepend` stuff)
Good work!

RobinDaugherty:
Thanks @Elektron1c97. I hate to admit that the reason I know about it is that I'm using it in production!

nateberkopec:
Merged in a4624495bf99c8da04f6

Issue #97, opened 2013-05-13
Circular references exhaust available stack space


johnskopis:
When an object containing circular references is passed as an :extra oj crashes after exhausting all of the available stack space.

PR #480, opened 2016-03-11
Optimizations

zanker:
Hey, we were noticing some performance issues with the Raven gem. I dug in and optimization a bunch of the major hotspots I found. All numbers below are base vs optimization, rather than cumulative of all optimizations done.

I added an option to use `JSON` instead of `OkJson`. `JSON` has a native C client which is much much faster. This was a 35% improvement.

Each `Backtrace::Line` instance would cache `in_app_pattern` regex. Since nothing in that regex relies on the line data, I moved it to be cached by the `Backtrace` class instead. That was 10%.

The `modules` and `stacktrace` portions of the event are all internal data that doesn't need to be redacted normally. I added an option to let people disable redaction, as those are the biggest parts of the event hash. This was a 52% improvement.

Those were the big ones, I did two smaller optimizations since they were easy. Calling `resolve_hostname` per exception is a little excessive, I moved it to be called when `Raven::Configuration` is initialized. I also did a small optimization to `stacktrace_interface_from` by reducing the amount of lines it needed to iterate over.

```
                    user     system      total        real
base   3.820000   0.040000   3.860000 (  3.974137)

                    user     system      total        real
optimizations   0.360000   0.000000   0.360000 (  0.365436)
```

RubyProf traces on Ruby 2.3.0, against the latest Raven master: [pre-optimization](https://gist.github.com/zanker/683c9346af06453a8b14), [post-optimization](https://gist.github.com/zanker/5278a25a2904f7542150).

In general, I tried to preserve the existing functionality and make the improvements be optional to avoid breaking anything.

cc @nerdrew

### DIFF
--- a/lib/raven/backtrace.rb
+++ b/lib/raven/backtrace.rb
@@ -12,8 +12,6 @@ module Raven
       # org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:170)
       JAVA_INPUT_FORMAT = %r{^(.+)\.([^\.]+)\(([^\:]+)\:(\d+)\)$}
 
-      APP_DIRS_PATTERN = /(bin|exe|app|config|lib|test)/
-
       # The file portion of the line (such as app/models/user.rb)
       attr_reader :file
 
@@ -49,18 +47,11 @@ module Raven
       end
 
       def in_app
-        app_dirs = Raven.configuration.app_dirs_pattern || APP_DIRS_PATTERN
-        @in_app_pattern ||= Regexp.new("^(#{project_root}/)?#{app_dirs}")
-
-        if self.file =~ @in_app_pattern
+        if self.file =~ Backtrace.in_app_pattern
           true
         else
           false
         end
-      end
-
-      def project_root
-        @project_root ||= Raven.configuration.project_root && Raven.configuration.project_root.to_s
       end
 
       # Reconstructs the line in a readable fashion
@@ -81,6 +72,8 @@ module Raven
       attr_writer :file, :number, :method, :module_name
     end
 
+    APP_DIRS_PATTERN = /(bin|exe|app|config|lib|test)/
+
     # holder for an Array of Backtrace::Line instances
     attr_reader :lines
 
@@ -99,6 +92,13 @@ module Raven
       end
 
       new(lines)
+    end
+
+    def self.in_app_pattern
+      @in_app_pattern ||= begin
+        project_root = Raven.configuration.project_root && Raven.configuration.project_root.to_s
+        Regexp.new("^(#{project_root}/)?#{Raven.configuration.app_dirs_pattern || APP_DIRS_PATTERN}")
+      end
     end
 
     def initialize(lines)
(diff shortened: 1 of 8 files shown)


nateberkopec:
Love it. 

> I added an option to use JSON instead of OkJson. JSON has a native C client which is much much faster. This was a 35% improvement.

Unfortunately, due to a lot of edge cases, we can't offer this option. Check the history of the repo (git blame on the OKJSON file might help), but there's a lot of reasons why we had to vendor OKJson.

I'll check out the rest soon :+1:

zanker:
No problem, it's not the end of the world to us to lose the `OkJson` one since I can monkeypatch that back in and fix whatever issues we had.

Do you have more info on the edge cases? Looking at https://github.com/getsentry/raven-ruby/commits/master/lib/raven/okjson.rb and running `JSON` with the `OkJson` it looks like the main one is Nan/Infinity handling which we can do with the `JSON` gem natively via `allow_nan`. Would there be a set of additional tests that could be added to increase confidence in `JSON`, or is it mostly infeasible?

seanlinsley:
Circular references in the JSON was the big reason: #97, #118

But perhaps that's no longer needed because #223 added the `Raven::Processor::RemoveCircularReferences` processor.

zanker:
That probably leaves the NaN/Infinity serializer as the big one then? JSON doesn't actually allow NaN, we can make it serialize it as just `Infinity` (eg `JSON.generate([ 1.0 / 0.0 ])` -> `[Infinity]`) through the `allow_nan` flag. I'm not sure how the Sentry server would handle deserializing it.

One alternative, would be to add a processor that converts `Infinity` or `NaN` into a string to match what you modified OkJson to do. Then we can use the JSON gem for everything.

nateberkopec:
I've cherry-picked some of these onto master. Still thinking about JSON and disabling stacktrace processing.

zanker:
Thanks! If it helps, we've been running the full set of patches for the last ~5 days on our largest Rails app without a problem.

nateberkopec:
Closing this out for now - let's continue the discussion on JSON performance at #481, and on optimizing how we process stacktraces at #485.

zanker:
Sounds good!

dcramer:
I'd say we could also cherry pick in some of these changes (like app_dirs_patterns)

dcramer:
I mispoke, and it turns out we already did cherry pick the obvious ones :)

zanker:
Yea I believe the only thing that wasn't picked in are the stack trace and the JSON optimizations if I recall!

nateberkopec:
@zanker Could you provide your original benchmark here?

zanker:
I can provide the code, but providing the test data I used is hard because it has a lot of internal data in it that's slightly hard to sanitize. I hacked these together, but here's the set of scripts I used: https://gist.github.com/zanker/c28cc2b2bcfe189a83f6ec09e589ea06

The one I can't include is `sanitize_test.json` which had 400 gems listed then about ~450 key/values for `exception`. Mostly with it traversing through a socket error.

PR #483, opened 2016-03-17
catch_debugged_exceptions -> rails_report_rescued_exceptions

### LINKED ISSUE #474: `config.catch_debugged_exceptions` in production
I was wondering why `config.catch_debugged_exceptions` prevents errors from being reported in production. I assumed that `debugged` is referring to the exception view in the development mode. In reality this also skips exceptions handled by `ActionDispatch::ShowExceptions` which is active in production:

```
RAILS_ENV=production bin/rake middleware | grep ShowExceptions
```

Is it the expected behavior that setting `config.catch_debugged_exceptions = false` will prevent error reports in production?

nateberkopec:
Closes #474 

cc @greysteil

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -122,6 +122,15 @@ Optional settings
           config.processors -= Raven::Processor::Cookies # Do this to send cookies by default
         end
 
+.. describe:: rails_report_rescued_exceptions
+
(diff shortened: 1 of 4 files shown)


greysteil:
Makes sense to me!

Issue #479, opened 2016-03-11
Context for Custom Exceptions


philipp-spiess:
I think adding Context to a Custom Exception through the `@__raven_context` member variable (see [here](https://github.com/getsentry/raven-ruby/wiki/Add-Context-for-Custom-Exceptions) and [here](https://github.com/getsentry/raven-ruby/blob/56761da7df7941f386c85605745df1bbb1a3e149/lib/raven/event.rb#L125)) is not ideal. To improve this, I suggest two changes:

1. Switch from a member variable to an actual public method. This allows us for a more straight forward API design and doesn't require the context to be set during initialisation.

  ```ruby
  # custom_exception.rb
  class CustomException < StandardError
    def raven_context
      {
        extra: {
          foo: "bar"
        }
      } 
    end
  end
  ```

2. We should use `deep_merge` instead of a regular `merge`. This way we won't override global context options. See [here](https://github.com/getsentry/raven-ruby/blob/56761da7df7941f386c85605745df1bbb1a3e149/lib/raven/event.rb#L126)
  
  Imagine the previous `CustomException` with the following `Raven` call:
  
  ```ruby
  Raven.capute_exception(CustomException.new, extra: { bar: "foo"} )
  ```
 
  A regular `#merge` would look like this:

  ```ruby
  ({extra: {foo: "bar"}}).merge({extra: {bar: "foo"}})
   # => {:extra=>{:bar=>"foo"}} 
  ```

  Compared to a `#deep_merge` implementation, which would preserve previous the context. 

  ```ruby
  ({extra: {foo: "bar"}}).deep_merge({extra: {bar: "foo"}})
   # => {:extra=>{:foo=>"bar", :bar=>"foo"}} 
  ```

  The problem with `#deep_merge` is that it would require another dependency as it's not part of the standard library. But [simple implementations are available](https://github.com/danielsdeleo/deep_merge/blob/master/lib/deep_merge/core.rb#L68).

dcramer:
While I realize patching isn't overly frowned upon in Ruby, it'd likely be better if we continue to provide an abstraction that is like ``Raven.annotate_exception(exc, context)`` which promotes a better programming style. I forget if that function already exists, but it can easily apply the private local store on the exception instance itself.

philipp-spiess:
I not sure what you mean by "patching". I'm proposing that we add an API to provide  context for custom exception classes. You could also do this in Java, for example, by implementing an Interface.

[...]

PR #491, opened 2016-04-07
Exception-based context

nateberkopec:
Re: #479

### DIFF
--- a/lib/raven/event.rb
+++ b/lib/raven/event.rb
@@ -69,6 +69,120 @@ module Raven
       @level      = LOG_LEVELS[@level.to_s.downcase] if @level.is_a?(String) || @level.is_a?(Symbol)
     end
 
+    class << self
+      def from_exception(exc, options = {}, &block)
+        exception_context = get_exception_context(exc) || {}
+        options = Raven::Utils::DeepMergeHash.deep_merge(exception_context, options)
+
+        configuration = options[:configuration] || Raven.configuration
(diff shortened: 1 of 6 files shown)


lib/raven/event.rb
```diff
@@ -69,6 +69,119 @@ def initialize(init = {})
...
+            evt.interface(:stacktrace) do |int|
+              stacktrace_interface_from(int, evt, options[:backtrace])
+            end
+          end
+        end
+      end
+
+      private
+
+      def get_exception_context(exc)
```
nateberkopec: hard to see in this diff, but here's the new important method

nateberkopec:
Gonna add docs, fix the build, then I think this is good to go.

juanibiapina:
We want to use annotate_exception and it seems this patch fixes a bug we've found (the need for the DeepMerge there). Without the deep merge, extra data we pass to annotate exception will be overwritten by the options passed to Event.from_exception.

Any plans on getting this merged?

nateberkopec:
Yes - it'll be in this week, I promise!

juanibiapina:
@nateberkopec hey, sorry to push you, but I took liberty to look at the tests and it seems the needed changes are merely style. Would it be possible to release it any time soon?

nateberkopec:
@juanibiapina yup, doing more work today and then its probably new version time

juanibiapina:
@nateberkopec Would you release a version with these changes if your current work is not yet finished?

PR #493, opened 2016-04-20
Expose last Event ID

dcramer:


dcramer:
Pardon my likely awful Ruby :)

nateberkopec:
Delaying merging this b/c I'm 99% certain this causes threading problems.

dcramer:
Should be threadsafe now

PR #496, opened 2016-05-02
Allow for bypassing the proxy

databus23:
Faraday by default picks up the proxy from `http_proxy` environment variable. This allows to explicitly bypass the proxy. This is useful in `on-premise` sentry scenarios where a proxy is needed to access other resources on the internet.


databus23:
This is not working. Faraday is not cooperating.

PR #497, opened 2016-05-02
Initial support on Breadcrumbs

dcramer:
undefined

### DIFF
--- a/lib/raven/breadcrumbs.rb
+++ b/lib/raven/breadcrumbs.rb
@@ -1,0 +1,74 @@
+module Raven
+  class Breadcrumb
+    attr_accessor :category, :data, :message, :level, :timestamp, :type
+
+    def initialize
+      @category = nil
+      @data = {}
+      @level = nil
+      @message = nil
+      @timestamp = Time.now.to_i
+      @type = nil
+    end
+
+    def to_hash
+      {
+        :category => @category,
+        :data => @data,
+        :level => @level,
+        :message => @message,
+        :timestamp => @timestamp,
+        :type => @type
+      }
+    end
+  end
+end
+
+module Raven
+  class BreadcrumbBuffer
+    include Enumerable
+
+    def self.current
+      Thread.current[:sentry_breadcrumbs] ||= new
+    end
+
+    def self.clear!
+      Thread.current[:sentry_breadcrumbs] = nil
+    end
+
+    def initialize(size = 100)
+      @buffer = Array.new(size)
+    end
+
+    def record(crumb = nil)
+      if block_given?
+        crumb = Breadcrumb.new if crumb.nil?
+        yield(crumb)
+      end
+      @buffer.slice!(0)
+      @buffer << crumb
+    end
+
+    def members
+      @buffer.compact
+    end
+
+    def peek
+      members.last
+    end
+
+    def each(&block)
+      members.each(&block)
+    end
+
+    def empty?
+      !members.any?
+    end
+
+    def to_hash
+      {
+        :values => members.map(&:to_hash)
+      }
+    end
+  end
+end
--- a/lib/raven/breadcrumbs/logger.rb
+++ b/lib/raven/breadcrumbs/logger.rb
@@ -1,0 +1,70 @@
+require 'logger'
+
+module Raven
+  module BreadcrumbLogger
+    LEVELS = {
+      ::Logger::DEBUG => 'debug',
+      ::Logger::INFO  => 'info',
+      ::Logger::WARN  => 'warn',
+      ::Logger::ERROR => 'error',
+      ::Logger::FATAL => 'fatal'
+    }.freeze
+
+    EXC_FORMAT = /^([a-zA-Z0-9]+)\:\s(.*)$/
+
+    def self.parse_exception(message)
+      lines = message.split(/\n\s*/)
+      # TODO: wat
+      return nil unless lines.length > 2
+
+      match = lines[0].match(EXC_FORMAT)
+      return nil unless match
+
+      _, type, value = match.to_a
+      [type, value]
+    end
+
+    def add(severity, message = nil, progname = nil, &block)
+      return if progname == "sentry" || Raven.configuration.exclude_loggers.include?(progname)
+      return if message.nil? || message == ""
+
+      # some loggers will add leading/trailing space as they (incorrectly, mind you)
+      # think of logging as a shortcut to std{out,err}
+      message = message.strip
+
+      last_crumb = Raven.breadcrumbs.peek
+      # try to avoid dupes from logger broadcasts
+      if last_crumb.nil? || last_crumb.message != message
+        error = Raven::BreadcrumbLogger.parse_exception(message)
+        # TODO(dcramer): we need to filter out the "currently captured error"
+        if error
+          Raven.breadcrumbs.record do |crumb|
+            crumb.level = Raven::BreadcrumbLogger::LEVELS.fetch(severity, nil)
+            crumb.category = progname || 'error'
+            crumb.type = 'error'
+            crumb.data = {
+              :type => error[0],
+              :value => error[1]
+            }
+          end
+        else
+          Raven.breadcrumbs.record do |crumb|
+            crumb.level = Raven::BreadcrumbLogger::LEVELS.fetch(severity, nil)
+            crumb.category = progname || 'logger'
+            crumb.message = message
+          end
+        end
+      end
+
+      super
+    end
+  end
+end
+
+class Logger
+  Raven.safely_prepend(
+    "BreadcrumbLogger",
+    :from => Raven,
+    :to => self
+  )
+end
--- a/lib/raven/configuration.rb
+++ b/lib/raven/configuration.rb
@@ -100,6 +100,9 @@ module Raven
     # Deprecated accessor
     attr_accessor :catch_debugged_exceptions
 
+    # Turns on ActiveSupport breadcrumbs integration
+    attr_accessor :rails_activesupport_breadcrumbs
+
     # Provide a configurable callback to determine event capture
     attr_accessor :should_capture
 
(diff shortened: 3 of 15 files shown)


lib/raven/event.rb
```diff
@@ -222,6 +223,7 @@ def to_hash
       data[:extra] = @extra if @extra
       data[:tags] = @tags if @tags
       data[:user] = @user if @user
+      data[:breadcrumbs] = @breadcrumbs.to_hash if @b
```
dcramer: lol tests passed, thast bad :(

dcramer:
@nateberkopec let me know your thoughts. I didn't update tests yet, but this changes a bunch of things. I think the approach to monkey patching Logger is good, but there's still a big TODO in that we need to filter out the Rails error logger message (which is a dupe), as well as ensure actual errors (if possible) get parsed into the correct types:

screenshot 2016-05-02 18 06 02

lib/raven/configuration.rb
```diff
@@ -137,7 +140,7 @@ def initialize
       self.processors = DEFAULT_PROCESSORS.dup
       self.ssl_verification = true
       self.encoding = 'gzip'
-      self.timeout = 1
+      self.timeout = 2
```
dcramer: im a bit concerned I had to do this, but we might update the default to 3 or something

dcramer:
I think the primary pieces missing now are:

1. Move the breadcrumb instrumentation to a new mechanism (vs integrations/)
2. Deal with filtering out the "current error" which often gets sent to logging
3. Fixing tests (and expanding them)

lib/raven/breadcrumbs.rb
```diff
@@ -0,0 +1,81 @@
...
+      @buffer[@pos] = crumb
+      @pos = (@pos + 1) % @size
+      @count += 1
+    end
+
+    def peek
+      return @buffer[@pos - 1]
+    end
+
+    def each
```
dcramer: Ideally this would function like blocks, but there might be some concerns there. Alternatively a better name is good enough (not sure whats standard).

dcramer:
Today I plan to:

- Add a logger stream so that you can manually push this as the output device (if you want to whitelist logger instead of use the default)
- Add 'breadcrumbs' hooks which will work in a similar way to integrations, and can be disabled one by one
- Find a way to ensure we don't send a duplicate breadcrumb when out of order execution happens (which it seems to in Rails, at the very least). // @mitsuhiko [[1]](https://app.getsentry.com/sentry-test/sentry-ruby-test/issues/126565924/)

dcramer:
so tl;dr we should focus on activesupport's broadcasts as it seems to be a better common vector

dcramer:
It looks like maybe the better solution will be to provide this logging adapter for people who want it, possibly disabled by default, but focus on Rails + http://edgeguides.rubyonrails.org/active_support_instrumentation.html

dcramer:
ActiveSupport works:

screenshot 2016-05-05 13 27 03

dcramer:
Alright all. Thoughts?

Logger is disabled by default. ActiveSupport is enabled for Rails.

I'm likely going to spend some time (it could be in a followup) to improve what ActiveSupport actually does (rather than blinding sending all events as data blobs).

spec/raven/breadcrumbs_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'spec_helper'
+
+describe Raven::BreadcrumbBuffer do
+  before(:each) do
+    @breadcrumbs = Raven::BreadcrumbBuffer.new(10)
+  end
+
+  it 'works' do
+    (0..10).each do |i|
+      @breadcrumbs.record(Raven::Breadcrumb.new {|b| b.message = i})
```
nateberkopec: Breadcrumb.new.tap { |b| b.message = i }

lib/raven/breadcrumbs.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    def initialize
+      @category = nil
+      @data = {}
+      @level = nil
+      @message = nil
+      @timestamp = Time.now.to_i
+      @type = nil
+
+      yield self if block_given?
```
nateberkopec: not necessary in ruby, just use #tap (see other comment)

lib/raven/breadcrumbs.rb
```diff
@@ -0,0 +1,79 @@
...
+end
+
+module Raven
+  class BreadcrumbBuffer
+    def self.current
+      Thread.current[:sentry_breadcrumbs] ||= new
+    end
+
+    def self.clear!
+      current
```
nateberkopec: doesnt work

lib/raven/breadcrumbs.rb
```diff
@@ -0,0 +1,79 @@
...
+      @buffer[@pos] = crumb
+      @pos = (@pos + 1) % @size
+      @count += 1
+    end
+
+    def peek
+      @buffer[@pos - 1]
+    end
+
+    def each
```
nateberkopec:
```ruby
(0..(@size - 1)).each_with_object([]) do |i, results|
  node = @buffer[(@pos + i) % @size]
  results << node unless node.nil?
end
```

nateberkopec: should we memoize this result?

lib/raven/breadcrumbs/logger.rb
```diff
@@ -0,0 +1,79 @@
...
+            crumb.category = progname || 'logger'
+          end
+        end
+      end
+    end
+  end
+end
+
+class Logger
+  prepend Raven::BreadcrumbLogger
```
nateberkopec: We need to deal with the Ruby 1.9 case here, I can fix it after merge.

lib/raven/breadcrumbs/logger.rb
```diff
@@ -0,0 +1,79 @@
...
+      return nil unless lines.length > 2
+
+      match = lines[0].match(EXC_FORMAT)
+      return nil unless match
+
+      _, type, value = match.to_a
+      [type, value]
+    end
+
+    def add(severity, message = nil, progname = nil, &block)
```
nateberkopec: rather than reimplement Logger#add, we should use `super` or an aliased method. can fix after merge.

lib/raven/breadcrumbs.rb
dcramer: seems overly complex since it will change frequently

dcramer:
Id like to come up with a solution for people who might want to disable whatever default breadcrumbs implementation we provide. Any thoughts there? Mostly focused on the railtie hook

dcramer:
Going to get some UI improvements in Sentry here, but I've tweaked some of the data. I think we can continue towards whitelisting breadcrumb events, and possibly event have the ActiveSupportBreadcrumbs class let you plug in a render per event type (so people with custom events could add their own).

seanlinsley:
I plan on trying this out in my application fairly soon, and providing feedback on the code itself. Excited to see this in action!

dcramer:
I put ActiveSupport behind configuration (opt-out), so I think at this point its ready to go.

dcramer:
build failures seem unrelated (looks like master is red)

seanlinsley:
Two weeks later, I'm now giving this a try 😅

seanlinsley:
Here are couple observations from the Sentry UI:

### Params show up as `[object Object]`

If they're being intentionally excluded, we should just remove the params key from this breadcrumb.

screen shot 2016-06-27 at 3 49 06 pm

### There are a ton of unreadable request objects

The fact that there are so many may be due to a bug in a gem we're using in development (https://github.com/MiniProfiler/rack-mini-profiler/issues/71), but it's odd that the breadcrumb is including them when they show no useful data.

screen shot 2016-06-27 at 3 52 02 pm


### The events appear to be out of order

Also it'd probably be good to show milliseconds.

screen shot 2016-06-27 at 3 53 02 pm


### Deprecations include a badly formatted stack trace

screen shot 2016-06-27 at 4 40 57 pm

seanlinsley:
Those issues can probably be resolved after this is merged but before it's released on Rubygems.

Today I'll also be testing this on our staging server, and looking over the code itself.

seanlinsley:
Is there a good way to record how long generating and sending the event to Sentry takes? I'm worried that this will cause timeouts on requests for people that aren't currently submitting them in the background (like me).

Perhaps the default behavior should be changed to send events in a thread.

dcramer:
This is good feedback

- the object issue I noticed and is a problem with how we capture crumbs today in ruby
- repetitive crumbs there's little we can do about other than letting you filter them out
- we can override the deprecated event to make the data function better
- we can reorder breadcrumbs on the sdk side
- we decided against ms as timing shouldn't be that important here (it's not perf)
- plus 1 on default threaded transport

seanlinsley:
> repetitive crumbs there's little we can do about other than letting you filter them out

One of our pages renders over 50 individual templates, each of which create a breadcrumb. Instead of collapsing the list to only show the newest breadcrumbs, it would be better to collapse sections of similar breadcrumbs.

So instead of having a UI saying "Show 91 collapsed crumbs", there'd be potentially multiple separate collapsed sets of breadcrumbs, titled something like "Show 50 collapsed render_template.action_view crumbs". In addition, it would be nice to be able to re-collapse that set without having to reload the page.

A lot of my comments here are Sentry server responsibilities. Should I ticket them individually on the main repo?

seanlinsley:
Given a `Raven.capture_message` call followed by an unhandled exception, neither event in Sentry has a breadcrumb for the other. Unless I'm mistaken, the JS Raven library supports this.

dcramer:
@seanlinsley yeah that makes sense -- probably also an issue in some of our others.

I'm thinking that maybe we start by disabling ActiveSupport by default so we at least get breadcrumb functionality available, and then we work towards improving the data capture of it.

dcramer:
@seanlinsley also no worries on ticketing things, we have an internal list of TODOs for breadcrumbs, and its just coming down to prioritization

nateberkopec:
So the unreadable request objects ( and I think [object Object]) are both because we're calling `to_s` on objects that don't have that method defined, and so we're just getting the output of `Object#to_s`, which is useless. An example of where this occurs is `ActionDispatch::Http::Headers`, which, although Hash-like, doesn't have a `to_s` method defined.

nateberkopec:
ActiveSupport disabled for now. Perf testing/more improvements to come.

nateberkopec:
I think I bungled the prepend bit here for Ruby 1.9. Fixing.

PR #499, opened 2016-05-03
Skip identical backtraces from "re-raised" exceptions

databus23:
Using sentry in a rails application I noticed that recorded exception contain a lot of duplicated backtraces blowing up the submitted JSON unnecessary.

This happens because in the great big rails onion there are a lot of places where exception are rewrapped preserving the backtrace.
E.g. all errors in views get wrapped in an [ActionView::Template::Error](https://github.com/rails/rails/blob/master/actionview/lib/action_view/template/error.rb#L72)

Depending on used gems and middleware the JSON payload gets quite big.
I'm seeing reported exceptions with 300KB of json containing 9 exceptions but only three distinct stacktraces.

I'm not sure if this PR is the best solution for tackling this but I wanted to have this elicit some feedback on how this could be improved.

### DIFF
--- a/lib/raven/event.rb
+++ b/lib/raven/event.rb
@@ -127,6 +127,7 @@ module Raven
         evt.interface(:exception) do |exc_int|
           exceptions = [exc]
           context = Set.new [exc.object_id]
+          backtraces = Set.new
 
           while exc.respond_to?(:cause) && exc.cause
             exc = exc.cause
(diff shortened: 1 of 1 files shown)


dcramer:
@nateberkopec thoughts on implementation? For correctness I realize this isn't great, but from a usability point of view I def agree with it. The approach to me seems pretty ok. We can def maintain the list of exceptions, and just not dupe the stacktrace.

PR #501, opened 2016-05-11
Convert CONTENT_{LENGTH,TYPE} correctly

mattrobenolt:
Previously, they were being converted into `Content_{Length,Type}` when
it should be `Content-{Length,Type}`

### DIFF
--- a/lib/raven/integrations/rack.rb
+++ b/lib/raven/integrations/rack.rb
@@ -106,8 +106,10 @@ module Raven
           # Header
           http_key = key[5..key.length - 1].split('_').map(&:capitalize).join('-')
           memo[http_key] = value
-        elsif %w(CONTENT_TYPE CONTENT_LENGTH).include? key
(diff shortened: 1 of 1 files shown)


lib/raven/integrations/rack.rb
```diff
@@ -106,8 +106,10 @@ def format_headers_for_sentry(env_hash)
           # Header
           http_key = key[5..key.length - 1].split('_').map(&:capitalize).join('-')
           memo[http_key] = value
-        elsif %w(CONTENT_TYPE CONTENT_LENGTH).include? key
-          memo[key.capitalize] = value
+        elsif key == 'CONTENT_TYPE'
```
mattrobenolt: I just chose to spell these out explicitly since there are only 2 of them and this set isn't going to be expanding. These are a part of the spec.

chrisdebruin:
👍

PR #503, opened 2016-05-18
Pass event id to Rack env

radiohead:
Record last event id to Rack's env so that it can be used in later stages of request.


nateberkopec:
Can you just use [`Raven::Base.last_event_id`](31f1f239ea10faa3b9f7bf21d53a1bca099e3ce7)?

radiohead:
@nateberkopec Sorry, I didn't realize that this has been already implemented since it's not in `1.0.0` that's available on rubygems. I'm not sure that there is benefit of having an alternative way so feel free to close this PR 👍

PR #504, opened 2016-05-24
Extract Raven::Instance and delegate from Raven to it

### LINKED ISSUE #462: Multiple Raven Instances?
Is there a way to configure multiple Raven instances? For example: sending a special subset of errors to one DSN and most errors to another DSN.

phillbaker:
This refactor was done in 2 steps:

1. Extract methods from Raven's top level module (in base.rb) into Raven::Instance, copy the tests, get them green
1. Delegate from `Raven` to an instance of `Raven::Instance`

These instances may be used like this:
```
class Foo
  def initialize
    @other_raven = Raven::Instance.new
    @other_raven.configure do |config|
      config.server = 'http://...'
    end
  end

  def foo
    ...
  rescue => e
    @other_raven.capture_exception(e)
  end
end
```

Fixes #462.

### DIFF
--- a/lib/raven/base.rb
+++ b/lib/raven/base.rb
@@ -17,187 +17,33 @@ require 'raven/interfaces/single_exception'
 require 'raven/interfaces/stack_trace'
 require 'raven/interfaces/http'
 require 'raven/utils/deep_merge'
+require 'raven/instance'
+
+require 'forwardable'
 
 module Raven
   AVAILABLE_INTEGRATIONS = %w[delayed_job railties sidekiq rack rake].freeze
 
   class << self
-    # The client object is responsible for delivering formatted data to the Sentry server.
-    # Must respond to #send. See Raven::Client.
-    attr_writer :client
+    extend Forwardable
 
-    # A Raven configuration object. Must act like a hash and return sensible
-    # values for all Raven configuration options. See Raven::Configuration.
-    attr_writer :configuration
-
-    def context
-      Context.current
+    def instance
+      @instance ||= Raven::Instance.new
     end
 
-    def logger
-      @logger ||= Logger.new
-    end
+    def_delegators :instance, :client=, :configuration=, :context, :logger, :configuration,
+                   :client, :report_status, :configure, :send_event, :capture, :capture_type,
+                   :last_event_id, :should_capture?, :annotate_exception, :user_context,
+                   :tags_context, :extra_context, :rack_context
 
-    # The configuration object.
-    # @see Raven.configure
-    def configuration
-      @configuration ||= Configuration.new
-    end
-
-    # The client object is responsible for delivering formatted data to the Sentry server.
-    def client
-      @client ||= Client.new(configuration)
-    end
-
-    # Tell the log that the client is good to go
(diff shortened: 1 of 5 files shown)


phillbaker:
@nateberkopec @dcramer what do you think?

nateberkopec:
@phillbaker Thanks for getting this running against Rubocop. I'll take a look as soon as I can.

lib/raven/instance.rb
```diff
@@ -0,0 +1,219 @@
+require 'raven/base'
+require 'English'
```
nateberkopec: ?

lib/raven/instance.rb
```diff
@@ -58,7 +67,6 @@ def report_status
         logger.info "Raven #{VERSION} configured not to send errors."
       end
     end
-    alias_method :report_ready, :report_status
```
nateberkopec: Why remove these? It seems like it only complicates your `Forwardable` related code.

lib/raven/instance.rb
```diff
@@ -90,13 +98,17 @@ def send_event(event)
     #     MyApp.run
     #   end
     def capture(options = {})
```
nateberkopec: Why was this changed?

nateberkopec:
Could we alias `Raven.client` to `Raven.instance`?

lib/raven/instance.rb
phillbaker: This was from rubocop - which request that `$!` on L211 be replaced with $ERROR_INFO from this module.

lib/raven/instance.rb
phillbaker: `Forwardable` doesn't support forwarding to aliases, (but you can set up an alias at the top level that forwards to the same underlying method) so all the aliases were moved up to `Base`. That also kept all the aliases consolidated, which seemed a bit cleaner if more needed to be added in the future.

lib/raven/base.rb
```diff
@@ -241,12 +87,6 @@ def rails_safely_prepend(module_name, opts = {})
...
 
-    # For cross-language compat
-    alias :captureException :capture_exception
-    alias :captureMessage :capture_message
-    alias :annotateException :annotate_exception
-    alias :annotate :annotate_exception
-
     private
 
     def install_at_exit_hook(options)
```
phillbaker: Whoops, this was moved to Instance, should have cleaned it up here.

lib/raven/instance.rb
phillbaker: Ah, sorry, this was an artifact of the 2 step process. In the 1st step, I had to remove the `install_at_exit_hook` during the copy to get Instance to pass the specs, then made it match the initial implementation when Base delegated to it.

phillbaker:
@nateberkopec just double checking I understand - I think `Raven.client` will delegate to the instance, it's listed here: https://github.com/phillbaker/raven-ruby/blob/dbb290e9810d5ebf27fabfdb68c047fb7b2fca39/lib/raven/base.rb#L35. I could move that next to the `client=` in the list for clarity

PR #507, opened 2016-06-02
refer to Rack doc in docs/index

cuzelac:
#### Issue Description

The [ruby client landing page's Configuration section](https://docs.getsentry.com/hosted/clients/ruby/#configuration) incorrectly says "If you use Rails, Rake, Rack etc, you’re already done".

If using Rack, you have to include a `use Raven::Rack`, as it says on [the integrations/rack page](https://docs.getsentry.com/hosted/clients/ruby/integrations/rack/)

Having a reference to the extra Rack setup on the ruby landing page would have saved me a few minutes so I bet it would save other people time too.

#### Disclaimer

I gave it a shot in this PR, but couldn't get `make html` to work, so I'm not totally sure it does what I expect.  Also, I see a sentry-docs repo, so maybe this isn't even the right place anymore!

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -43,9 +43,11 @@ environment variable or manually configure it with ``Raven.configure``:
 Reporting Failures
 ------------------
 
-If you use Rails, Rake, Rack etc, you're already done - no more
+If you use Rails, Rake, Sidekiq, etc, you're already done - no more
 configuration required! Check :doc:`integrations/index` for more details on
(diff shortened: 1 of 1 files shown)


dcramer:
Thanks! Our docs setup is a bit complex, but this is the right place. We pull this repo into sentry-docs and build them together as a larger site.

nateberkopec:
Thanks! I chopped up the README a while back to shorten it up and introduced this issue.

Issue #481, opened 2016-03-15
OkJSON doesn't handle large JSON bodies


brandonjoyce:
We had a JSON message w/ base64 encoded data that it turned out Sentry was locking up on while parsing with the included OkJSON code.  It doesn't error or anything, it just takes a very long time.  (hours) to parse just a ~9mb attachment.  

https://github.com/getsentry/raven-ruby/blob/master/lib/raven/processor/sanitizedata.rb#L75

I'm willing to help w/ this fix, but I want some advice on what direction to go.  Here are a few options I can think of:

- Switch to the built-in ruby json library.  Guessing this is not possible for other reasons, but it does handle the large JSON without a problem.
- Add a default processor to Raven that truncates large json bodies at a limit that's acceptable for OkJSON.  
- Fix OkJSON?  - Not sure how feasible it is to go this route.

seanlinsley:
There's recent conversation on the subject of dropping OkJSON on #480

brandonjoyce:
Very cool.  I think that move to the built-in json library would solve our issue for sure.

PR #510, opened 2016-06-10
Switch to built-in JSON module

dcramer:
Refs GH-480, GH-481


dcramer:
If circular references are indeed an issue we need to be concerned about we can whip up some tests and a quick solution. I'm not sure if the cost is too high to always attempt to remove them before encoding, or if theres a better way to do it with JSON.

seanlinsley:
Huh... I could've sworn there was a test for circular references back when I worked on #166.

Looks like the original PR adding OkJSON (#118) did indeed add a test.

#223 moved it to [(implementation)](https://github.com/getsentry/raven-ruby/blob/944e72b67b3e/lib/raven/processor/removecircularreferences.rb) & [(spec)](https://github.com/getsentry/raven-ruby/blob/944e72b67b3e/spec/raven/processors/removecirculareferences_spec.rb)

seanlinsley:
Since the test is still around, I'd say :shipit:

mdayaram:
Cool! Are we good to go with this? Would love to see this in the main gem, we've been working off a fork in the interim that uses the native JSON module instead.

dcramer:
I'll defer to @seanlinsley and @nateberkopec if they think this is mergable

dcramer:
But since @seanlinsley suggests so, lets do it!

mdayaram:
\o/ woo! Do we have any idea when the next release that includes this change be?

seanlinsley:
🎉 🍻 It's great that we're finally relying on the built-in JSON library

PR #512, opened 2016-06-13
Fix silence_ready not working for Rails apps

ream88:
by invoking the first configure block after your initializers are called. Before, using an initializer would generate a second clients, with the first one still logging to STDOUT.

Until this is merged, someone can use this in his application.rb to get the same result:

    Rails.application.config.before_configuration do
      Raven.configuration.silence_ready = true
    end

See also: https://github.com/getsentry/raven-ruby/issues/492

### DIFF
--- a/lib/raven/integrations/rails.rb
+++ b/lib/raven/integrations/rails.rb
@@ -24,15 +24,13 @@ module Raven
       end
     end
 
-    config.before_initialize do
+    config.after_initialize do
       Raven.configure do |config|
         config.logger ||= ::Rails.logger
         config.project_root ||= ::Rails.root
-        config.release = config.detect_release # if project_root has changed, need to re-check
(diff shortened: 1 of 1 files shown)


ream88:
Does anyone know what happened with the Travis build?

dcramer:
@ream88 i re-ran the build

ream88:
Ah, ok my local rspec wasn't running at all, and I misinterpreted the rubocop output as test output!

ream88:
Yeah, fixed it now :)

dcramer:
Thanks!

eugeneius:
I originally did this in https://github.com/getsentry/raven-ruby/pull/450, but @nateberkopec pointed out a use case that it broke: https://github.com/getsentry/raven-ruby/pull/450#issuecomment-178000161

Issue #513, opened 2016-06-13
Raven crashes during rspec, when it is not supposed to be running at all


pratik60:
We have only configured raven for production and staging, but it seems to be executing the blog even while running tests. 

This was a controller test which was failing, but debugging is hard because this error crops up. In the controller we were attempting to capture exception using Raven and I thought by default it would only happen in the configured environments.

How can we fix this?

/Users/pratikbothra/.rvm/gems/ruby-2.2.2/gems/sentry-raven-1.1.0/lib/raven/utils/deep_merge.rb:10:in `deep_merge!': undefined method `each_pair' for #<RSpec::Mocks::MockExpectationError:0x007fb70acc04c0> (NoMethodError)
	from /Users/pratikbothra/.rvm/gems/ruby-2.2.2/gems/sentry-raven-1.1.0/lib/raven/utils/deep_merge.rb:6:in `deep_merge'
	from /Users/pratikbothra/.rvm/gems/ruby-2.2.2/gems/sentry-raven-1.1.0/lib/raven/event.rb:75:in `from_exception'
	from /Users/pratikbothra/.rvm/gems/ruby-2.2.2/gems/sentry-raven-1.1.0/lib/raven/base.rb:111:in `capture_type'
	from /Users/pratikbothra/.rvm/gems/ruby-2.2.2/gems/sentry-raven-1.1.0/lib/raven/base.rb:256:in `block in install_at_exit_hook'

seanlinsley:
Here's what I do:
```ruby
case Rails.env
when 'production', 'staging'
  require 'raven'

  Raven.configure do |config|
    config.dsn = 'foo'
  end
when 'development', 'test'
  class Raven
    def self.capture_message(*)
    end

    def self.capture_exception(*)
    end

    def self.capture(*)
      yield if block_given?
    end

    def self.user_context(*)
    end
  end
end
```

nateberkopec:
That's intense. I guess we still try to capture exceptions even when the DSN isn't configured. That's kind of dumb, we should just turn off completely.

pratik60:
@seanlinsley That works as a good temp solution. We end up doing something like Raven.capture_exception(e) if Rails.env.staging? || Rails.env.production? etc.

@nateberkopec I agree, it should ideally be turned off completely.

PR #518, opened 2016-07-03
Do not capture if env doesn't match or DSN not set

nateberkopec:
Previously, even when event *sending* to the server was disabled by lack of a DSN or the current environment not matching the env list, we would still capture a Raven event. This commit causes Raven to not even capture events when it would not send them.

cc @dcramer - any legitimate reason people want to capture but not send?

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-ruby/518)


nateberkopec:
this is all vis a vis #513

PR #521, opened 2016-07-10
Add whitelist to sanitizer

nateberkopec:
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-ruby/521)


nateberkopec:
Not satisfied with this. We could do something like http://api.rubyonrails.org/v4.2/classes/ActiveSupport/SafeBuffer.html, and then construct objects internally as "Blessed" objects, which can be bypassed by processors.

PR #530, opened 2016-08-05
Send environment from current_environment

dcramer:
Future change might involve renaming current_environment to environment and environments to something that looks like like the core attribute.

@getsentry/ruby

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-ruby/530)

### DIFF
--- a/lib/raven/event.rb
+++ b/lib/raven/event.rb
@@ -26,7 +26,7 @@ module Raven
     attr_reader :id
     attr_accessor :project, :message, :timestamp, :time_spent, :level, :logger,
       :culprit, :server_name, :release, :modules, :extra, :tags, :context, :configuration,
-      :checksum, :fingerprint
+      :checksum, :fingerprint, :environment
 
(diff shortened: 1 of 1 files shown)


lib/raven/event.rb
```diff
@@ -220,6 +221,7 @@ def to_hash
       data[:culprit] = @culprit if @culprit
       data[:server_name] = @server_name if @server_name
       data[:release] = @release if @release
+      data[:environment] = @envirionment if @environment
```
aortez: Looks like ```@envirionment``` is misspelled.

dcramer: this was fixed in a followup commit

aortez: Ok, thanks!  (and sorry for not looking more closely)

nateberkopec: Don't sweat it, I merged it and missed the same thing! 😆

PR #533, opened 2016-08-17
Ensure @port is always initialized

mhenrixon:
We are getting warnings about the port not being initialized. This change ensures that the `@port` variable is always defined before it is used as to avoid these warnings.

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-ruby/533)

### DIFF
--- a/lib/raven/configuration.rb
+++ b/lib/raven/configuration.rb
@@ -181,7 +181,7 @@ module Raven
 
       @scheme = uri.scheme
       @host = uri.host
-      @port = uri.port if uri.port
+      @port = uri.port ? uri.port : nil
       @path = uri_path.join('/')
 
       # For anyone who wants to read the base server string


esparta:
> We are getting warnings about the port not being initialized

I don't get it. Who gives the warnings?

On ruby all the assignments get initialized, and the ivars will always get nil

```ruby
  t = 'wow' if 1 == 0
  p t # => nil
  if 1 == 0
     r = 'no wow'
  end
  p r  # => nil
  p @non_existent # => nil
  p non_existent # NameError
```

So, setting explicitly to nil is actually the same as the current code.

mhenrixon:
Usually I would agree with you but that is not the result we where getting. We are doing some multithreading with concurrent-ruby. Not sure why but the problem goes away with that change in my local copy with this change.

nateberkopec:
> We are getting warnings 

I'm not sure I understand this either. What warning?

nateberkopec:
I think OP may be referring to running Ruby with warnings on. I just ran the suite with `RUBYOPT="-w"` and...whoo boy. I'm taking a look at it.

mhenrixon:
Yeah, sorry about the late reply. I am referring to running RSpec or ruby with warnings on. I suggest you turn it on for the test suite. You might find something interesting :)

nateberkopec:
Thanks for the reminder @mhenrixon, we're now clean in verbose mode thanks to dfe70220830c1a5e8e191809b2fa382c255286af

Issue #534, opened 2016-08-26
Log4r Incompatibility with Sentry (1.2.1) and greater


brentonmorton:
Started to upgrade some applications to Sentry 1.2.1 and found an incompatibility with log4r. 

There was a change made between 1.1.0 and 1.2.1. In ( lib/raven/logger.rb ) that went from:

- 1.1.0 (line: 19) `logger.send(level, "#{LOG_PREFIX}#{msg}") if logger`
- 1.2.1 (line: 29) `logger.add(LEVELS[level], "#{LOG_PREFIX}#{msg}", "sentry") if logger`

The problem with this is there is a bug that log4r has that conflicts with (add).

https://github.com/colbygk/log4r/issues/30

This is valid for Ruby's Logger class (see Logger.add), however for log4r it tries to add an outputter in Log4r::Logger.add

Wondering if you guys know about this already?

nateberkopec:
Did not know, will check it out 👍

nateberkopec:
This is kinda log4r's problem, but the switch to `add` for us was just sort of incidental, so I can switch it back to `send`.

brentonmorton:
Thanks for the fix!

PR #535, opened 2016-08-30
Log4r compat problems

nateberkopec:
When this was changed from `send` to `add` I think we just did it because it was easier to test. I can't think of a great way to test the provided block here, so we're not really testing the log output.

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-ruby/535)


nateberkopec:
Re: #534

Issue #511, opened 2016-06-10
Event.to_hash sometimes contains anonymous modules or classes, which cannot be dumped to YAML


nateberkopec:
It's common to have code like this:

```ruby
SentryJob = Struct.new(:event) do
  def perform
    Raven.send_event(event)
  end
end

Raven.configure do |config|
  config.async = lambda { |event|
    Delayed::Job.enqueue SentryJob.new(event.to_hash)
  }
end
```

`event.to_hash` can contain an anonymous module, which cannot be dumped to YAML formats for use in a job processor.

dcramer:
It might make sense for us to change how async works so we encode the object before passing it in.

nateberkopec:
JSON backends seem to work fine (any storage adapter that calls `to_json` instead of `to_yaml`).

nateberkopec:
We could just hand the encoded version to the job processor, though that will probably break everyones existing jobs.

nateberkopec:
Confirmed that #506 is a duplicate of this. The error message there is slightly different: 

> ActiveJob::SerializationError: Unsupported argument type: Class

nateberkopec:
I think it's Rails that's doing this (something in the Rack env, perhaps), so I'm going to have to try to get a repro case.

nateberkopec:
Also it would be good to try to fallback to sync'd event sending if something in `config.async` raises an exception, as suggested.

nuclearpidgeon:
I am experiencing a similar issue trying to async report with ActiveJob, using DelayedJob as the backend.

I get an `Unsupported argument type: Time` error that comes out of [activejob-4.2.6/lib/active_job/arguments.rb:69](https://github.com/rails/rails/blob/v4.2.6/activejob/lib/active_job/arguments.rb#L69)

Our job code at the moment is pretty much the same as the Sentry example:

```
class ReportSentryEventJob < ActiveJob::Base
  queue_as :default

  def perform(event)
    Raven.send_event(event)
  end
end
```
```
# config/production.rb
  Raven.configure do |config|
    # Process sending exceptions using ActiveJob
    config.async = lambda do |event|
      ReportSentryEventJob.perform_later(event.to_hash)
    end
  end
```

I logged the `event.to_hash` hash to take a look at it and spotted a few Time/DateTime values nested under the 'extra' key. Note the lack of string quotes around the time values:

```
:extra => {
  :delayed_job => {
    :id => 766,
    :priority => 0,
    :attempts => 2,
[...]

PR #545, opened 2016-09-20
Explicitly convert selected Delayed Job values to strings

nuclearpidgeon:
Fixes issues that occur when using DelayedJob as an asynchronous handler for Sentry/Raven itself.

These issues are related to #511

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-ruby/545)

### DIFF
--- a/lib/raven/integrations/delayed_job.rb
+++ b/lib/raven/integrations/delayed_job.rb
@@ -11,16 +11,19 @@ module Delayed
 
           rescue Exception => exception
             # Log error to Sentry
+            #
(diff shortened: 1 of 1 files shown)


nateberkopec:
Thanks for the example and PR. I won't be merging this, though, because it only fixes *your* problem (bad types in DelayedJob/ActiveJob), and we have a general problem of the wrong types being fed to job backends. A general solution here will fix your problem too.

nuclearpidgeon:
No worries - yeah it definitely only felt like a quick fix.

PR #547, opened 2016-09-20
Pass JSON-compatible hashes to `async`

### LINKED ISSUE #511: Event.to_hash sometimes contains anonymous modules or classes, which cannot be dumped to YAML

nateberkopec:
Fixes #511.

This changes the type of the object given to the `async` callback.

This change will *break* any async callback like the following:

```ruby
SentryJob = Struct.new(:event) do
  def perform
    Raven.send_event(event)
  end
end

Raven.configure do |config|
  config.async = lambda { |event|
    event.extra = { 'some_stuff' => 'modified_here' }
    Delayed::Job.enqueue SentryJob.new(event.to_hash)
  }
end
```

It *will not break* the most common async setup, which looks like:

```ruby
SentryJob = Struct.new(:event) do
  def perform
    Raven.send_event(event)
  end
end

Raven.configure do |config|
  config.async = lambda { |event|
    Delayed::Job.enqueue SentryJob.new(event.to_hash)
  }
end
```

Benchmarks indicate that this slows down capture by 2-3x, though using `async` is still an order of magnitude faster than not using it (not including HTTP costs).

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-ruby/547)

### DIFF
--- a/lib/raven/event.rb
+++ b/lib/raven/event.rb
@@ -247,6 +247,10 @@ module Raven
       "#{lastframe.filename} in #{lastframe.function} at line #{lastframe.lineno}" if lastframe
     end
 
+    def to_json_compatible
+      JSON.parse(JSON.generate(to_hash))
+    end
+
     # For cross-language compat
     class << self
       alias captureException from_exception
--- a/lib/raven/instance.rb
+++ b/lib/raven/instance.rb
@@ -120,7 +120,9 @@ module Raven
         yield evt if block_given?
         if configuration.async?
           begin
-            configuration.async.call(evt)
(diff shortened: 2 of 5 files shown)


nateberkopec:
Next release will be raven-ruby 2.0.0.

Issue #506, opened 2016-05-25
Async w/ ActiveJob silently fails


budnik:
I just tried to setup `raven-ruby` following docs:
```ruby
Raven.configure do |config|
  config.async = ->(event) { SentryJob.perform_later(event.to_hash) }
end
```
It was silently failing to enqueue the job util I changed to:
```ruby
Raven.configure do |config|
  config.async = ->(event) { SentryJob.perform_later(event.to_hash.except(:request) }
end
```
I tried both `:inline` and `:sidekiq` AJ backends, I think there's some serialization issue that is swallowed silently. I'm going to provide repro gists soon meanwhile I opening this issue now to share this workaround.

nateberkopec:
:( We should make sure we fail loudly if `async` is misconfigured.

jeremy:
Hit this too. May be nice to fall back to non-async capture of both the event and the async exception.

nateberkopec:
[FWIW we're getting `Thread.report_on_exception` in Ruby 2.4](https://bugs.ruby-lang.org/issues/6647).

nateberkopec:
I think this is related to #511.

nateberkopec:
PR for sync fallback is available at #548

PR #548, opened 2016-09-20
Fallback to sync sending if async fails

nateberkopec:
Note this won't catch exceptions in threads, obviously.

See also #511 and #506

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-ruby/548)

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -23,8 +23,10 @@ Optional settings
           Thread.new { Raven.send_event(event) }
         }
 
-    Using a thread to send events will be adequate for truly parallel Ruby platforms such as JRuby, though the benefit on MRI/CRuby will be limited.
+    Using a thread to send events will be adequate for truly parallel Ruby platforms such as JRuby, though the benefit on MRI/CRuby will be limited. Threads also won't report any exceptions raised inside of them, so be careful!
 
+    If the async callback raises an exception, Raven will attempt to send synchronously.
+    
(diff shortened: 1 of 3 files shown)


nateberkopec:
Anyone know of any jiu-jitsu I can do if someone is using a Thread that raises an exception?

nuclearpidgeon:
You could perhaps wrap the exception handling code in a lambda/block, and pass that as another argument to the `config.async` lambda/block - but it's pretty superfluous at that point as that handler would just call 'send_event' anyway.. which presumably would have just failed for some reason. Unless you want to just pass in the Logger call.

Also just a heads up, that [config.rst page doesn't seem to render properly in GitHub](https://github.com/nateberkopec/raven-ruby/blob/2f67d78722db5fe19782d0a745878e3249e695d3/docs/config.rst). Not sure if this is just a GitHub problem or some kind of syntax error

nateberkopec:
Hmm. [We could `call` the `transport_failure_callback`.](https://github.com/getsentry/raven-ruby/blob/master/lib/raven/configuration.rb#L139). Normally that's called when there's a 500 error from Sentry, but it might be appropriate here too.

dcramer:
Of note 429 and 403 codes, for example, should not be retried. So as long as it's only 500 it might be ok. That's still not great behavior though as it means sdks could hammer sentry in certain situations.

nateberkopec:
Actually, `send_event` will just call `transport_failure_callback` anyway if Sentry returns an error. I think a loud (error-level) log message is probably enough here.

PR #555, opened 2016-09-26
Sidekiq: identify culprit + drop Sidekiq 2

nateberkopec:
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-ruby/555)

### DIFF
--- a/lib/raven/integrations/sidekiq.rb
+++ b/lib/raven/integrations/sidekiq.rb
@@ -3,56 +3,62 @@ require 'sidekiq'
 
 module Raven
   class Sidekiq
-    def call(_worker, msg, _queue)
-      started_at = Time.now
-      yield
-    rescue Exception => ex
-      Raven.capture_exception(ex, :extra => { :sidekiq => msg },
-                                  :time_spent => Time.now - started_at)
-      raise
+    ACTIVEJOB_RESERVED_PREFIX = "_aj_".freeze
+
+    def call(ex, context)
+      context = filter_context(context)
+      Raven.capture_exception(
+        ex,
+        :message => ex.message,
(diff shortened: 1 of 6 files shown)


nateberkopec:
screen shot 2016-09-26 at 2 47 22 pm

nateberkopec:
note to self, not sure if I should be clearing context/breadcrumb buffers

lib/raven/integrations/sidekiq.rb
```diff
@@ -3,56 +3,57 @@
...
-def filter_context(context)
-  case context
-  when Array
-    context.map { |arg| filter_context(arg) }
-  when Hash
-    Hash[context.map { |key, value| filter_context_hash(key, value) }]
-  else
-    context
+    def culprit_from_context(context)
+      classname = (context["class"] || (context["job"] && context["job"]["class"]))
```
nateberkopec: need to add a comment about this data structure is consistent in future versions of sidekiq, a la https://github.com/mperham/sidekiq/pull/3161

lib/raven/integrations/sidekiq.rb
```diff
@@ -3,56 +3,57 @@
...
-    rescue Exception => ex
-      Raven.capture_exception(ex, :extra => { :sidekiq => msg },
-                                  :time_spent => Time.now - started_at)
-      raise
-    ensure
-      Context.clear!
-      BreadcrumbBuffer.clear!
+    ACTIVEJOB_RESERVED_PREFIX = "_aj_".freeze
+
+    def call(ex, context)
```
nateberkopec: this is a really simple middleware that can be tested in isolation from sidekiq, should add some tests

PR #560, opened 2016-10-03
Add OS, Runtime context

nateberkopec:


nateberkopec:
Event needs a refactor real bad. Working on that next.

Issue #563, opened 2016-10-06
Heroku release version sometimes inaccurate 


nfm:
Related to #377.

It looks like `ENV['HEROKU_SLUG_COMMIT']` is only available for git-push based deploys. We used to use git-push to deploy, and switched to Heroku's Pipelines feature several months ago. The `HEROKU_SLUG_COMMIT` environment variable remains present, but is set to a very old value - the last SHA that we deployed via git-push.

I couldn't find another environment variable that Heroku supplies that contains the SHA from a Pipeline deploy. However, `heroku labs:enable runtime-dyno-metadata` writes this metadata to `/etc/heroku/dyno` (see https://github.com/heroku/fix/blob/6c8ab7a/lib/heroku_dyno_metadata.rb).

There might be a better way to do this, I'm not sure - this was the only solution I could find!

Would you be open to a PR that read the metadata from this file first, then fell back to `HEROKU_SLUG_COMMIT`, and finally to `git rev-parse --short HEAD`?

nateberkopec:
Weird, I'll take a look.

nateberkopec:
Opened up a ticket with Heroku support, we'll see what's up.

nfm:
Hopefully they can do something on their end, reading it from an environment variable is much nicer!

nateberkopec:
Well I basically got a non-answer from Heroku. We could just eliminate the 'HEROKU_SLUG_COMMIT' check and just shell out to git immediately.

nateberkopec:
@nfm To clarify - HEROKU_SLUG_COMMIT the environment variable is *incorrect*, but in the dyno metadata it *is* correct?

nfm:
@nateberkopec That's right, for us, `HEROKU_SLUG_COMMIT` is present and incorrect.

The dyno metadata is correct, but it's an opt-in feature. The file `/etc/heroku/dyno` doesn't exist unless you enable the `runtime-dyno-metadata` labs feature.

I just made a new Heroku app, and did a Pipeline deploy (rather than a git-push deploy), and can confirm that the `HEROKU_SLUG_COMMIT` environment variable isn't set at all. So I think it would only be present and incorrect if you were initially deploying via git-push, and then switched to Pipelines for deploying. If you always deploy with git-push, it should be present and correct. And if you never deploy with git-push, it seems to be absent.

Unfortunately, we can't shell out to git on Heroku, as it removes the `.git` directory from the compiled slug.

atmos:
[...]

PR #566, opened 2016-10-11
Heroku detection works on /etc/dyno

nateberkopec:


lib/raven/configuration.rb
```diff
@@ -267,15 +267,23 @@ def project_root=(root_dir)
...
     end
 
     private
 
     def detect_release_from_heroku
-      ENV['HEROKU_SLUG_COMMIT']
+      sys_dyno_info = `cat /etc/heroku/dyno` rescue nil
+      return unless sys_dyno_info && sys_dyno_info != ""
+
+      begin
```
nateberkopec: being overly cautious here because failure means Raven fails to start

nateberkopec:
Bummer that Heroku users have to enable a labs extension to get release detection to work. #563

gudmundur:
The plan is that this information becomes a default. The question, afaik, is how it'll be delivered.

nateberkopec:
@gudmundur 👍  Where's the best place for me to keep to up-to-date on platform-level changes like that?

PR #567, opened 2016-10-19
Fix Rails 5 `params.to_hash` deprecation warning in docs

Nitrodist:
In Rails 5.0, calling `params.to_hash` logs a deprecation warning [0],
so we should be advising our users to call to_unsafe_h as rails 5.1 will
just raise an error. This code is compatible with rails 4.2.

[0] - https://github.com/rails/rails/issues/23084


nateberkopec:
LGTM. Unsafe hash is appropriate here because it isn't being fed into any security-sensitive context.

PR #570, opened 2016-10-20
MessageInterface support for invalid format string msg when params is…

jmalves:
We were having a problem were certain errors were not being reported.

The cause was that `MessageInterface#unformatted_message` was producing an _ArgumentError: malformed format string_.  This was due to the message being an invalid format string (e.g.: _Process failed at char '%'_). This occurred even tough there were no `params` defined.

I changed the behaviour of `MessageInterface#unformatted_message` to not apply the `params` to the `message` using `%` if there were no `params` defined. 

Not sure if there a more appropriate way to solve this issue. In any case the test case might be helpful.

### DIFF
--- a/lib/raven/interfaces/message.rb
+++ b/lib/raven/interfaces/message.rb
@@ -10,7 +10,7 @@ module Raven
     end
 
     def unformatted_message
-      message % params
+      params.nil? ? message : message % params
     end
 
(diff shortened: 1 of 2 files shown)


nateberkopec:
Whoops that's pretty bad. Will get this fixed and a 2.2.1 out ASAP.

Issue #571, opened 2016-10-20
JRuby - ensure we're not starting a whole new Ruby process with version detection


nateberkopec:
`-J-Djruby.launch.inproc=true`

nateberkopec:
^^^^

PR #572, opened 2016-10-20
Travis: add ruby.launch.inproc

olleolleolle:
The JRuby option `ruby.launch.inproc` is mentioned in #571.

This PR tries to run a build with that option. [The JRuby wiki page on performance explains](https://github.com/jruby/jruby/wiki/Improving-startup-time#avoid-spawning-sub-rubies).

This is the bit where it would be beneficial to not have to start another JVM:
https://github.com/getsentry/raven-ruby/blob/master/lib/raven/context.rb#L37

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ sudo: false
 matrix:
   include:
     - rvm: jruby-1.7.26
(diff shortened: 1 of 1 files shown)


nateberkopec:
Thank you!

PR #573, opened 2016-10-20
Travis: Ruby 2.4.0-preview2 supported, just

olleolleolle:
This PR elbows in support for 2.4.0-preview2 to be able to run the test suite.

It conditionally adds these gems to the Gemfile:

- json 2.x
- yayl-ruby at a very recent commit (no release of the gem as yet)

And it removes 2.4.0-preview2 from allowed failures.


olleolleolle:
Darn – too fiddly. But it's close for many of these gems.

Issue #568, opened 2016-10-20
Give info what configuration is missing to capture errors


clupprich:
When the Raven client starts up it emits a log line staging `Raven xxx configured not to capture errors`.

I was too excited to follow the instructions when signing up on sentry.io, so I missed the `server` configuration. Would it be helpful to state the reason for not capturing errors (like: "no server configured" or "environment not listed")? If yes, I'm happy to contribute a PR!

nateberkopec:
@clupprich Absolutely!

nateberkopec:
Closed by #602 - this will be release in 2.2.0 very soon.

PR #574, opened 2016-10-23
Add reason to capture info message

clupprich:
This PR addresses issue #568 and adds an explanatory messages to the `#report_status` method (which gets outputted on startup) telling the user why events won't be captured.
- I'm not totally happy with the naming of `verify_messages` - maybe "reasons" somewhere in there would fit better?
- I've introduced `Configuration#server_configured?`, which is replacing `!!server` in `#capture_in_current_environment?`. Turns out that the double negation always returned `true` because of the way how `@server` is [assembled](https://github.com/getsentry/raven-ruby/blob/d2c51187ae944f80a9212173d7872f82fe9368d4/lib/raven/configuration.rb#L207-L209). How about we use `uri.to_s` here? That would returns an empty string for an empty configuration value.
- The specs for the `Raven` class were not green when only running that spec. That's fixed by d7b14bd, which resets the instance variable for each test run.
- I've made the specs for `#report_status` in the `Raven` class more integration-like, as they were basically the same specs as for the `Instance` class. They are now changing "real" configuration values.

Happy for any kind of feedback!

### DIFF
--- a/lib/raven/configuration.rb
+++ b/lib/raven/configuration.rb
@@ -248,7 +248,8 @@ module Raven
     end
 
     def capture_allowed?(message_or_exc = nil)
-      capture_in_current_environment? &&
+      server_configured? &&
+        capture_in_current_environment? &&
         capture_allowed_by_callback?(message_or_exc)
     end
 
@@ -259,6 +260,13 @@ module Raven
       %w(server public_key secret_key project_id).each do |key|
         raise(Error, "No #{key} specified") unless public_send key
       end
+    end
+
+    def verify_messages
+      messages = []
+      messages << 'no server configured' unless server_configured?
+      messages << 'current environment not listed' unless capture_in_current_environment?
+      messages
     end
 
     def project_root=(root_dir)
@@ -299,7 +307,11 @@ module Raven
     end
 
     def capture_in_current_environment?
-      !!server && (environments.empty? || environments.include?(current_environment))
+      environments.empty? || environments.include?(current_environment)
+    end
+
+    def server_configured?
+      host && path
     end
 
     def capture_allowed_by_callback?(message_or_exc)
--- a/lib/raven/configuration.rb
+++ b/lib/raven/configuration.rb
@@ -257,7 +257,7 @@ module Raven
     alias sending_allowed? capture_allowed?
 
     def verify!
-      %w(server public_key secret_key project_id).each do |key|
+      %w(server host path public_key secret_key project_id).each do |key|
         raise(Error, "No #{key} specified") unless public_send key
       end
     end
@@ -311,7 +311,10 @@ module Raven
     end
 
     def server_configured?
-      host && path
+      verify!
+      true
+    rescue Error
+      false
     end
 
     def capture_allowed_by_callback?(message_or_exc)
--- a/lib/raven/configuration.rb
+++ b/lib/raven/configuration.rb
@@ -160,6 +160,8 @@ module Raven
       Raven::Processor::HTTPHeaders
     ].freeze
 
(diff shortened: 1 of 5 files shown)


clupprich:
Hm, the order of the commits seem to be mixed up - [here's the branch on my fork](https://github.com/clupprich/raven-ruby/commits/add-reason-to-capture-info-message). Let's see what Travis comes up with...

lib/raven/configuration.rb
```diff
@@ -299,7 +307,11 @@ def detect_release_from_git
     end
 
     def capture_in_current_environment?
-      !!server && (environments.empty? || environments.include?(current_environment))
+      environments.empty? || environments.include?(current_environment)
+    end
+
+    def server_configured?
```
nateberkopec: Let's try and re-use `verify!` from `Configuration` for this.

lib/raven/configuration.rb
```diff
@@ -261,6 +262,13 @@ def verify!
       end
     end
 
+    def verify_messages
```
nateberkopec: shouldn't this method also cover `capture_allowed_by_callback`?

clupprich: I agree - but it's a bit hard to check ;) `capture_allowed_by_callback?` is always returning `true` if the message is `nil`, so we'd need to send a bogus value.

I guess really interesting is only a filter that will always block, like `Raven.configuration.should_capture = -> { false }`.

lib/raven/configuration.rb
```diff
@@ -311,7 +311,10 @@ def capture_in_current_environment?
     end
 
     def server_configured?
-      host && path
+      verify!
```
jmalves: Wouldn't it be better to make this method without using `rescue`?  According to the fast-ruby lecture I got to know due to @nateberkopec this pattern not only looks IMO bad also seems to be quite slow. `capture_allowed?` uses this method and it seems to be called quite often. 

I would propose the following:

1. Move the needed settings to a final variable (i.e. the `%w(...)` )
2. Create a method that returns the index of the first missing setting. (e.g.: something like `firstMissingSetting`)
3. Adapt `verify!` and `server_configured?` to use this new `firstMissingSetting`.

PS - Thx @clupprich for this feature!

lib/raven/configuration.rb
clupprich: Exactly - and that's the issue, imo: we don't get any useful information, cause it will always be  `true`, i.e. we will never report a blocked capture because of the callback when testing with `nil`.

lib/raven/configuration.rb
clupprich: Will move the settings to a constant.

With `verify!` returning true/false, we would loose the descriptive errors (e.g. "No server specified"). We could return an array of messages from `verify!` and then raise the first one if not empty (although I'm not sure if I really like that - I guess no ;) ).

clupprich:
@nateberkopec Sorry, been away for a couple of days. I've updated with the proposed changes regarding the `verify!` method - let me know if that's :+1:

Regarding the missing check for `capture_allowed_by_callback?` - I still don't know how to solve this. Say the `verify_messages` method looks like this:

```
def verify_messages
  messages = []
  messages << 'no server configured' unless server_configured?
  messages << 'current environment not listed' unless capture_in_current_environment?
  messages << 'capture filtered out in callback' unless capture_allowed_by_callback?(message)
  messages
end
```

`capture_allowed_by_callback?` will return `true` if `message` is `nil`. We would need to pass a more educated guess to the callback check to make sense!? Or am I missing something?

nateberkopec:
Closing in favor of #602 - I decided to go with a separate errors atribute, similar to how ActiveRecord validation works.

PR #576, opened 2016-10-27
Always match excluded exceptions by case equality

jage:
... no matter if they are defined by string or constant in
Raven.configuration[:excluded_exceptions].

The upside of doing it like this is that Raven will have the same match
behavior no matter how the exception was excluded. Previously a
string-defined excluded exception wouldn't match inherited/extended
exceptions.

Use case example:

```Ruby
  # Raven setup
  Raven.configure do |config|
    config.excluded_exceptions = ["SentryExcludedException"]
  end

  # Application
  module SentryExcludedException; end

  # The application can decide what to exclude
  begin
    raise StandardError, "oh snap"
  rescue => error
    error.extend SentryExcludedException
    raise
  end
```

This only worked if the excluded_exceptions was configured with
constants, but not with strings, which this fixes.

By using strings the coupling between the Raven setup/environment
and the application is reduced, improving unit testing etc.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,7 +4,7 @@ AllCops:
     - 'lib/raven/utils/deep_merge.rb'
 
 Metrics/ClassLength:
-  Max: 233
+  Max: 246
   CountComments: false
 
 Metrics/AbcSize:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,7 +4,7 @@ AllCops:
     - 'lib/raven/utils/deep_merge.rb'
 
 Metrics/ClassLength:
-  Max: 246
+  Max: 248
   CountComments: false
 
 Metrics/AbcSize:
--- a/lib/raven/event.rb
+++ b/lib/raven/event.rb
@@ -100,7 +100,7 @@ module Raven
           Raven.logger.info "Refusing to capture Raven error: #{exc.inspect}"
           return nil
         end
-        if configuration[:excluded_exceptions].any? { |x| (x === exc rescue false) || x == exc.class.name }
+        if configuration[:excluded_exceptions].any? { |x| get_exception_class(x) === exc }
           Raven.logger.info "User excluded error: #{exc.inspect}"
           return nil
         end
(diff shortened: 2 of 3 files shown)


.rubocop.yml
```diff
@@ -4,7 +4,7 @@ AllCops:
     - 'lib/raven/utils/deep_merge.rb'
 
 Metrics/ClassLength:
-  Max: 233
+  Max: 246
```
nateberkopec: lol, I really need to refactor Event.

lib/raven/event.rb
```diff
@@ -134,6 +134,22 @@ def from_message(message, options = {})
 
       private
 
+      def get_exception_class(x)
+        x.is_a?(Module) ? x : qualified_const_get(x)
+      rescue NameError # There's no way to safely ask if a constant exist for an unknown string
```
nateberkopec: This `rescue` should be in `qualified_const_get`

lib/raven/event.rb
```diff
@@ -134,6 +134,22 @@ def from_message(message, options = {})
...
+        nil
+      end
+
+      # In Ruby <2.0 const_get can't lookup "SomeModule::SomeClass" in one go
+      def qualified_const_get(x)
+        parts = x.to_s.split("::").delete_if(&:empty?)
```
nateberkopec: Let's convert `x` to a string immediately here so we only have to do it once.

nateberkopec: i.e `x = x.to_s`

lib/raven/event.rb
```diff
@@ -134,6 +134,22 @@ def from_message(message, options = {})
...
+        if parts.size < 2
+          Object.const_get(x.to_s)
```
nateberkopec: Can we use `const_defined` to avoid raising an exception here?

lib/raven/event.rb
nateberkopec: also, I'm surprised there's no Rubocop rule for this, but I would prefer `reject!` over `delete_if` here.

lib/raven/event.rb
```diff
@@ -134,6 +134,22 @@ def from_message(message, options = {})
...
+        else
+          parts.inject(Object) { |a, e| a.const_get(e) }
```
nateberkopec: Clever!

lib/raven/event.rb
jage: Hmm, not sure it's worth it, we're not completely exception-safe even with `const_defined?` (this was not the what I expecting).

```Ruby
[1] pry(main)> Object.const_defined?("String")
=> true
[2] pry(main)> Object.const_defined?("string")
NameError: wrong constant name string
from (pry):2:in `const_defined?'
```

I did a quick benchmark comparing a rescue and checking with `const_defined?` (this was before I realised `const_defined?` could raise exceptions for non-camel-case strings). When the constant is undefined the rescue-version was actually faster. This shouldn't be that performance sensitive since it's only run when exceptions are being captured, but still. :)
https://gist.github.com/jage/b0d3c397b5b46de421b537cbe41219f7#benchmark-how-to-look-up-a-class-from-string-in-ruby

Still think we should do a check before getting the constant?

lib/raven/event.rb
jage: Yes, that's better!

lib/raven/event.rb
jage: Hmm, this comment was falsely marked as "outdated" since I fixed the other issues. Will wait on a response from @nateberkopec before doing anything more regarding this.

nateberkopec:
Thank you for your contribution!

PR #585, opened 2016-11-11
Adding header info to the error message if any

rafadc:
In case we are debugging any strange error this information can be
super useful.

I was finding a lot of errors and the header sometimes adds useful hints. 

I don't know if the format of the error is adequate or if you prefer logging in a different fashion.

Thanks a lot!


nateberkopec:
Huge thumbs up on adding this info, will take a look at the implementation soon.

PR #594, opened 2016-11-22
Do not sanitize frozen objects

nateberkopec:


nateberkopec:
I think I'm OK with this. If you expect frozen objects to be sanitized you can write your own processor. Anyone freezing security-sensitive data would be...weird.

PR #598, opened 2016-11-22
Raven::Instance should use its own context/config when creating Event

### LINKED ISSUE #597: Raven::Instance still uses global context and configuration
When creating a `Raven::Event`, it just calls `from_message` or `from_exception` and those use `Raven.context` and `Raven.configuration` (there appear to be some ways to override this behaviour, but `Raven::Instance` is not using them)

singpolyma-shopify:
Closes #597

### DIFF
--- a/lib/raven/event.rb
+++ b/lib/raven/event.rb
@@ -28,8 +28,8 @@ module Raven
                   :context, :configuration, :checksum, :fingerprint, :environment,
                   :server_os, :runtime, :breadcrumbs, :user, :backtrace
 
-    def initialize(init = {})
-      @configuration = Raven.configuration
+    def initialize(init = {}, configuration = Raven.configuration)
+      @configuration = configuration
       @interfaces    = {}
       @breadcrumbs   = Raven.breadcrumbs
       @context       = Raven.context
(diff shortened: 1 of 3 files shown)


lib/raven/event.rb
```diff
@@ -39,20 +39,21 @@ def initialize(init = {})
...
       @server_os     = {} # TODO: contexts
       @runtime       = {} # TODO: contexts
       @tags          = {} # TODO: contexts
       @checksum      = nil
       @fingerprint   = nil
-      @environment   = @configuration.current_environment
 
       yield self if block_given?
 
+      @environment   = @configuration.current_environment
```
nateberkopec: Why reorder this section?

nateberkopec: Placing it after `yield self` is a problem.

singpolyma-shopify: @nateberkopec I've pushed an alternate implementation that solves my issue without the re-ordering

nateberkopec:
@singpolyma-shopify Can you see if #601 also meets your needs? I don't want to change method signatures where possible.

PR #600, opened 2016-11-22
Travis: Cache bundler and RVM

nateberkopec:


nateberkopec:
NVM, this is pointless. RVM is just downloading a huge binary in the first place, this won't make it any faster.

PR #601, opened 2016-11-23
Instance config/context passed correctly to new Events

nateberkopec:
Alt to #598


singpolyma-shopify:
This seems to work for my purposes.

PR #603, opened 2016-11-24
Allow providing source to files not on local disk

singpolyma-shopify:
This allows source context to be provided for backtraces that reference interactive code, etc.

### DIFF
--- a/lib/raven/configuration.rb
+++ b/lib/raven/configuration.rb
@@ -119,6 +119,10 @@ module Raven
     # Silences ready message when true.
     attr_accessor :silence_ready
 
+    # Map of filename to Array of source lines
+    # for sources not easily read off disk
+    attr_accessor :sources
+
     # SSL settings passed directly to Faraday's ssl option
     attr_accessor :ssl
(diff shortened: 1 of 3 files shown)


nateberkopec:
Interesting, can you describe your use case a bit more here? What kind of sources can't be read off the disk for your app?

singpolyma-shopify:
@nateberkopec we have scripts stored in a DB (for versioning, etc) that we would like to report errors in to a sentry instance

singpolyma-shopify:
@nateberkopec I'll fix the rubocop if there's a chance of this approach going in.  I'm open to other approaches, though, if you are willing to have a feature that solves this use case, but don't like this one :)

nateberkopec:
@singpolyma-shopify I want to ask internally about it to see how we handle it in the other language clients - I have no problem really with this approach though.

singpolyma-shopify:
@nateberkopec I've pushed a refactor to make rubocop happy

nateberkopec:
Merged #606

PR #604, opened 2016-11-28
Allows faraday 0.10.x

gfx:
Strict dependency versions are considered harmful. Semantic versioning style seems better.

See https://github.com/lostisland/faraday/releases for faraday change logs.


nateberkopec:
We had a reason for this. I'll revisit.

gfx:
Thank you!

PR #606, opened 2016-11-28
Allow providing your own LineCache

nateberkopec:
Also, removing the constant cache in LineCache for thread-safety. Still not perfect, but improved.


nateberkopec:
@singpolyma-shopify This work for you?

You would just provide a custom LineCache instance wherever you needed it - globally or to a Raven::Instance.

singpolyma-shopify:
Looks great!  Much more extensible than my solution :)

PR #609, opened 2016-12-09
Allow some fields to be excluded from sanitization

rclavel:
That's a feature we need for our project.
In our project, I have monkey-patch `Configuration` and created a new processor, but I think this option could be in the original processor.

### DIFF
--- a/lib/raven/configuration.rb
+++ b/lib/raven/configuration.rb
@@ -96,6 +96,9 @@ module Raven
     # a hash key, will be censored and not sent to Sentry.
     attr_accessor :sanitize_fields
 
+    # Array of fields that should never be sanitized.
+    attr_accessor :excluded_fields
+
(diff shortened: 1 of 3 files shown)


nateberkopec:
We already skip frozen values, so I'm not sure what the use case is for this patch. Can you explain your use?

rclavel:
I have a field with `authorization` in its name, and I don't want it to be sanitized.
For this use case, do you freeze explicitly the value?
It seems better for me to put the list of excluded values in the initializer, where all the configuration is set.

nateberkopec:
@mattrobenolt even if I don't sanitize such a field, won't the upstream Sentry server sanitize it anyway?

rclavel:
@nateberkopec It will, but there is a setting on admin view to exclude some fields from sanitization.

PR #610, opened 2016-12-13
Various logger improvements

### LINKED ISSUE #607: Noisy in development/test
Despite I don't set a dsn in test/development environments, and don't even call `Raven.configure` in these environments, I do get a lot of Raven messages which do nothing but distract me.

Examples:

* `Raven 2.1.4 configured not to capture errors.`
* `<error message> excluded from capture due to environment or should_capture callback`

How to entirely silence Raven logging for test/development?

Cheers - Victor

nateberkopec:
Fixes #607.

End result is that "Not sending due to X" errors are not displayed unless you edit Raven's log level. We still log by default on sending an error and upon startup ("ready to catch" message).

### DIFF
--- a/lib/raven/client.rb
+++ b/lib/raven/client.rb
@@ -32,7 +32,7 @@ module Raven
         return
       end
 
-      configuration.logger.debug "Sending event #{event[:event_id]} to Sentry"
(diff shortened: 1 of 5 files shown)


nateberkopec:
Maybe add some docs about setting logger levels and how2troubleshoot.

PR #612, opened 2016-12-13
WIP: Improve capture performance

nateberkopec:
Currently 2-3x faster.

TODO:

* [x] Any frozen objects that may be bad? Should we provide some sort of
fallback path here?
* [x] Test JSON backends. Which ones can deal with circular references, bad utf8?
* [x] Hashes coming out of storage (delayedjob, sidekiq) are NOT FROZEN.
* [x] Generic modify-in-place tools. How is santizedata so much faster than other processors we use?


nateberkopec:
I'm going to finish local variable capture before working any more on this, because what happens there will definitely affect this.

nateberkopec:
For Ruby 2.4:

Metric | master | this branch
------------ | ------------- | -------------
Allocations | 2656 | 1113
Iterations/second | 114 | 251

lib/raven/interfaces/exception.rb
```diff
@@ -11,6 +11,7 @@ def self.sentry_alias
     def to_hash(*args)
       data = super(*args)
       data[:values] = data[:values].map(&:to_hash) if data[:values]
+      data[:values].freeze
```
nateberkopec: Removes exceptions from sanitizer path. RemoveStacktraceProcessor still works, of course.

lib/raven/processor/removecircularreferences.rb
```diff
@@ -3,10 +3,12 @@ class Processor::RemoveCircularReferences < Processor
     def process(v, visited = [])
       return "(...)" if visited.include?(v.__id__)
       visited += [v.__id__]
-      if v.is_a?(Hash)
-        v.each_with_object({}) { |(k, v_), memo| memo[k] = process(v_, visited) }
+      if v.frozen?
```
nateberkopec: We can't modify a frozen value anyway, but this will also ignore members of frozen collections.

nateberkopec: Modifying this so that it continues to process frozen collections (with possible not-frozen submembers) doesn't significantly impact perf, so I'll do that.

nateberkopec:
YAJL only improves overall perf by 1-3%, so I'm not going to look into supporting alternative JSON backends any more.

spec/raven/processors/removecirculareferences_spec.rb
```diff
@@ -19,7 +19,7 @@
     result = @processor.process(data)
     expect(result['data']).to eq('(...)')
     expect(result['ary'].first['x']).to eq('(...)')
-    expect(result['ary2']).not_to eq('(...)')
+    expect(result['ary2']).to eq("(...)")
```
nateberkopec: TODO: fix

nateberkopec:
This is ready to ship except for the weird Ruby 1.9 hang.

In the long term, the next major perf step is eliminating stacktraces from the processor chain, which would double speed again.

Ideas for that - monkeypatch and add a "blessed" attribute to particular hashes and arrays, create 2 hashes, where only one gets fed to processors.

nateberkopec:
To clarify, when I said:

> YAJL only improves overall perf by 1-3%, so I'm not going to look into supporting alternative JSON backends any more.

I meant that certain backends could, for example, accept events with circular references or bad UTF8, so I could create a "fast path" where we don't use those processors if you have a particular JSON backend. I found the speedups to be not very meaningful, so I abandoned it. You can still use any JSON backend with Sentry (if it overrides the JSON class).

PR #613, opened 2016-12-16
WIP: 3.0.

nateberkopec:
* [x] Event.capture* should redirect to Raven.capture*
* [x] Add performance improvements
* [x] Fix rubocop discrepancies
* [ ] Allow some fields to be excluded from sanitization
* [ ] Fix Sidekiq integration via #635 
* [ ] #665 
* [ ] #661 
* [ ] Stop pretending a `Raven.configuration` global exists.


nateberkopec:
Ultimately I decided not to include local variable capture - it's too unstable and buggy and could never be used in production, so the use case is too limited.

nateberkopec:
Just realized we'll need a 2.x series release where `to_hash` uses string keys and still supports symbol keys in order to provide a smooth transition to 3.0. Otherwise, peoples events in background job processors will get eaten up.

PR #619, opened 2017-01-06
Handle JRuby uri:classloader filenames in backtraces

the-michael-toy:
When we updated our app from JRuby 1.7 to JRuby9000 9.1.5.0 sentry stopped working.

This fixes it for us, but any time I touch a regex I assume I have made a mistake,
so I am not 100% confident I didn't break parsing of error messages that used
to parse correctly.

### DIFF
--- a/lib/raven/backtrace.rb
+++ b/lib/raven/backtrace.rb
@@ -6,7 +6,7 @@ module Raven
     # Handles backtrace parsing line by line
     class Line
       # regexp (optional leading X: on windows, or JRuby9000 class-prefix)
-      RUBY_INPUT_FORMAT = %r/
+      RUBY_INPUT_FORMAT = /
         ^ \s* (?: [a-zA-Z]: | uri:classloader: )? ([^:]+ | <.*>):
         (\d+)
         (?: :in \s `([^']+)')?$
--- a/lib/raven/backtrace.rb
+++ b/lib/raven/backtrace.rb
@@ -5,19 +5,12 @@ module Raven
   class Backtrace
     # Handles backtrace parsing line by line
     class Line
-      # regexp (optional leading X: or windows or JRuby9000 class-prefix)
-      RUBY_INPUT_FORMAT = %r{
+      # regexp (optional leading X: on windows, or JRuby9000 class-prefix)
+      RUBY_INPUT_FORMAT = %r/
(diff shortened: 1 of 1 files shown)


the-michael-toy:
UGH, I have violated rules about regular expressions and Travis hates me.

Happy to fix this to your spec, but not doing anything in case you want to
fix this by changing the style guide.

the-michael-toy:
sorry for the multiple commits, i never figured out how to run tests. all green now.

nateberkopec:
No problem. Looks good, but I want to test this out a bit more / really understand the regex changes.

the-michael-toy:
yup ... i spent some time reading regex web pages to understand the old one

`/^((?:[a-zA-Z]:)?[^:]+|<.*>):(\d+)(?::in `([^']+)')?$/`

i'll point out the differences so you can make an informed call and save yourself some time.

```
      RUBY_INPUT_FORMAT = /
        ^ \s* (?: [a-zA-Z]: | uri:classloader: )? ([^:]+ | <.*>):
        (\d+)
        (?: :in \s `([^']+)')?$
      /x
```

0. I used white space ignoring regex so I could break it apart into pieces and reason about the new regex I was writing.
1. Because I kept cutting and pasting errors during testing and I often grabbed an extra space, I throw away spaces at the beginning of the line, probably harmless but also probably harmless to take out the `\s*`
2. The first capture, which is the file name, in the old regex starts right after the `^` but I felt it was clearer to start the capture right at the point where we were actually capturing.
3. The actual bug fix, now we ignore either `X:` or `uri:classloader:` at the start of the string.
4. In third (method name) capture, I use `\s` instead of the space character after the `in` to match the pattern ``in `method_name'``. To be 100% compatible I would have used `\ ` (escaped space).

So a version which was ONLY different for the bugfix, which I probably should have done, would be

```
      RUBY_INPUT_FORMAT = /
        ^ (?: [a-zA-Z]: | uri:classloader: )? ([^:]+ | <.*>):
        (\d+)
        (?: :in \  `([^']+)')?$
      /x
```

and I guess you could scrunch it up too ..

`/^((?:[a-zA-Z]:|uri:classloader:)?[^:]+|<.*>):(\d+)(?::in `([^']+)')?$/`

PR #621, opened 2017-01-07
Show top level rake tasks instead process name

Bugagazavr:
Before:
screen shot 2017-01-07 at 21 37 35

After:
screen shot 2017-01-07 at 21 37 23


nateberkopec:
Dig it. Thanks!

PR #627, opened 2017-01-25
log backtrace at once

chulkilee:


chulkilee:
Although output is same, it would be useful when custom formatter is used (e.g. producing json for elasticsearch). Since there are only 10 lines, I guess there are little performance impacts.

nateberkopec:
Works for me!

PR #633, opened 2017-02-08
Increase truncation limit on handler to 1000

darrennix:
I am trying to debug a Job error. The problem is that Sentry is truncating the "handler" under "Additional Data". What I see in Sentry is
```
--- !ruby/object:Delayed::PerformableMailer
object: !ruby/class 'GeneralMailer'
method_name: :generi
```

This requires me to go digging outside of Sentry to find the full handler to get the ID of the troublesome entry (in args)
```
--- !ruby/object:Delayed::PerformableMailer
object: !ruby/class 'GeneralMailer'
method_name: :generic_message
args:
- 305294"
```

### DIFF
--- a/lib/raven/integrations/delayed_job.rb
+++ b/lib/raven/integrations/delayed_job.rb
@@ -24,10 +24,10 @@ module Delayed
               }
             }
             # last_error can be nil
-            extra[:last_error] = job.last_error[0...100] if job.last_error
+            extra[:last_error] = job.last_error[0...1000] if job.last_error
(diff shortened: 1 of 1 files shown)


nateberkopec:
Hmm, we could probably even go a little bigger. The main thing is we don't want to trip the limit of a TEXT field (~60k characters, depending on DB). 1k for both the handler and last_error is probably fine. 

@darrennix Wanna change this PR to modify the truncation for last_error and handler to 1,000?

darrennix:
Done!

PR #634, opened 2017-02-20
Temporarily accept string keys in processors.

### LINKED ISSUE #622: Several processors broken when using async
After this change: https://github.com/getsentry/raven-ruby/commit/a786def7b423109c964d26fae90e0b83507aac41 the async handler is passed a hash with string keys whereas `send_event` and the processors included in raven expect symbol keys in the event hash.

It is not clear that the client is expected to transform the event hash back into a hash with symbol keys before passing to `send_event`.

Just enumerating some ideas:
- `send_event` should turn all the keys into symbols
- When working with an event hash throughout the gem string keys should be used
- Revert this change and have clients JSONify the event if needed in their async proc
- Some sort of configuration option to pass the actual event object to async

nateberkopec:
Close #622

This will be removed in 3.0, when all event hashes will be expected
to contain string keys.

### DIFF
--- a/lib/raven/processor/cookies.rb
+++ b/lib/raven/processor/cookies.rb
@@ -1,16 +1,26 @@
 module Raven
   class Processor::Cookies < Processor
     def process(data)
-      if data[:request]
-        # Remove possibly sensitive cookies
-        data[:request][:cookies] = STRING_MASK if data[:request][:cookies]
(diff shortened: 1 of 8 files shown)


garjust:
I think `send_event` also operates on the event hash using symbols

Issue #630, opened 2017-01-31
Exceptions from wrapped ActiveJob Sidekiq jobs lacking context


vindvaki:
After upgrading `raven-ruby` from an ancient version to the latest version, the only context we get for exceptions raised by workers is `Sidekiq`, greatly reducing the usefulness the list view (lots of errors look the same). Basically, all our worker errors are generic stuff like:

<img width="434" alt="screen shot 2017-01-31 at 16 30 26" src="https://cloud.githubusercontent.com/assets/80280/22471464/add13a6c-e7d2-11e6-9170-e6a3618510d2.png">

whereas before we used to have the file name of the template where the exception was raised.

I've been digging around and it seems like https://github.com/getsentry/raven-ruby/blob/master/lib/raven/integrations/sidekiq.rb#L48 might be missing the case of wrapped jobs, in which case

```ruby
context["job"]["wrapped"]
```

contains at least the job class name (in this case `MailJob`), which would already be a huge improvement over simply `Sidekiq`.

nateberkopec:
TIL what wrapped jobs are! Thanks, I'll add it.

nateberkopec:
^^^ closed on master

PR #635, opened 2017-02-20
Use :job instead of "job" and prefer "wrapped" over "class"

vindvaki:
I took a stab at https://github.com/getsentry/raven-ruby/issues/630 myself and noticed that `"job"` key should have been `:job`, which explains why we were getting the culprit

```ruby
Sidekiq
```

(implying `classname == nil`) instead of the value of `context[:job]["class"]`, which is

```ruby
ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper
```

for wrapped jobs.

### DIFF
--- a/lib/raven/integrations/sidekiq.rb
+++ b/lib/raven/integrations/sidekiq.rb
@@ -45,7 +45,7 @@ module Raven
     # this will change in the future:
     # https://github.com/mperham/sidekiq/pull/3161
     def culprit_from_context(context)
-      classname = (context["class"] || (context["job"] && context["job"]["class"]))
+      classname = (context["class"] || (context[:job] && (context[:job]["wrapped"] || context[:job]["class"])))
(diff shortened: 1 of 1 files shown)


vindvaki:
FYI https://github.com/rails/rails/blob/master/activejob/lib/active_job/queue_adapters/sidekiq_adapter.rb

nateberkopec:
Sigh, this is actually #622. The context hash may be symbols or strings.

I'm wondering if I should just leave this for 3.0, when the event hash will always be string-keyed.

vindvaki:
Fair enough. Just remember that there are two things going here:

1. `wrapped` vs `class`
2. symbol keys vs string keys

The former issue remains regardless of the latter, but there's already an issue open for that in #630. And TBH I don't even know if this PR would break other people's code 😝, so feel free to close it.

PR #637, opened 2017-02-21
Clear stale breadcrumbs and context after Sidekiq jobs

### LINKED ISSUE #636: Context and breadcrumbs not cleared after sidekiq job completes
I've been using sidekiq-pro with Raven and it seems like the context is not being cleared between jobs. Looking at the stack trace in an issue I don't see raven or sentry mention at all so I'm not sure if it's filtering itself out of the stacktrace or just isn't getting hooked up correctly. 

Any tips on an easy way to see if the sidekiq middleware is getting included?

drewish:
The error handler is only called after an exception so the context (specifically breadcrumbs) will carry over from previous jobs that completed successfully.

Fixes #636

I'm not sure this is exactly the way the maintainers would want to approach it but it seems to be working better for me. I'm not using ActiveJob so I don't know if that would play into this.

You can test this out by creating two new sidekiq jobs:
```
bundle exec rails g sidekiq:worker Happy
bundle exec rails g sidekiq:worker Sad
```
Then fill in some basic processing logic to have one record breadcrumbs and context:
```rb
class HappyWorker
  include Sidekiq::Worker

  def perform(*args)
    Raven.breadcrumbs.record do |crumb|
      crumb.message = "I'm happy!"
    end
    Raven.tags_context mood: 'happy'
  end
end
```
And the other just raise an exception and doesn't retry
```rb
class SadWorker
  include Sidekiq::Worker

  sidekiq_options :retry => false

  def perform(*args)
    raise "I'm sad!"
  end
end
```

Start up sidekiq with a single thread (so we don't have to worry about which worker picks up which job):
```
bundle exec sidekiq -c 1
```

Open up a rails console in another terminal window and queue up a couple of jobs:
```
HappyWorker.perform_async
HappyWorker.perform_async
SadWorker.perform_async
```

On the master branch the Sentry issue from `SadWorker` will include the breadcrumbs and tags from the `HappyWorker`:
runtimeerror__i_m_sad_

### DIFF
--- a/lib/raven/integrations/sidekiq.rb
+++ b/lib/raven/integrations/sidekiq.rb
@@ -2,7 +2,16 @@ require 'time'
 require 'sidekiq'
 
 module Raven
-  class Sidekiq
+  class SidekiqCleanupMiddleware
+    def call(_worker, _job, _queue)
+      yield
+    ensure
+      Context.clear!
+      BreadcrumbBuffer.clear!
+    end
+  end
+
+  class SidekiqErrorHandler
     ACTIVEJOB_RESERVED_PREFIX = "_aj_".freeze
 
     def call(ex, context)
@@ -13,9 +22,6 @@ module Raven
         :extra => { :sidekiq => context },
         :culprit => culprit_from_context(context)
       )
-    ensure
-      Context.clear!
-      BreadcrumbBuffer.clear!
     end
 
     private
@@ -59,6 +65,9 @@ end
 
 if Sidekiq::VERSION > '3'
   Sidekiq.configure_server do |config|
-    config.error_handlers << Raven::Sidekiq.new
+    config.error_handlers << Raven::SidekiqErrorHandler.new
+    config.server_middleware do |chain|
+      chain.add Raven::SidekiqCleanupMiddleware
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


tarmo:
Hi @drewish did you explicitly also test the case that tags added in SadWorker actually end up recorded in Sentry with the exception raised by it (ie. not leaked tags from previous jobs but those that we intentionally want to report for the current job)?

From some limited testing it seems that the `ensure` block for `SidekiqCleanupMiddleware` actually gets called before Sidekiq calls any error handlers so all the context will be gone by the time the exception is reported to Sentry. This is confirmed by both adding debugger calls to the ensure block and exception reporting code and observing the former being called first and also having lost all tags attached to job exception reports after updating to a raven-ruby version containing this change.

tarmo:
I think it might work if the middleware only cleared context when no exception was raised and the error handler also cleared context. I'll try if that works.

drewish:
@tarmo Can you say more about how the tags would be provided? I was definitely testing that tags and breadcrumbs from the failed job were being reported. Now specifically it was picking up breadcrumbs from the failed jobs but the extra and tags were being passed in when capturing, e.g. `Raven.capture_exception(exception, extra: extra, tags: tags, level: level)`.

drewish:
@tarmo Ah okay yeah I wasn't relying on the default capture so I understand the case you're highlighting now. Definitely a shortcoming in my testing.

PR #638, opened 2017-02-25
Use Rails logger by default again

eugeneius:
Before https://github.com/getsentry/raven-ruby/pull/599, `Raven.configuration.logger` was nil if there was no logger configured; now it's always an instance of `Raven::Logger`. This means that the `||=` assignment in the railtie no longer has any effect.

If we unconditionally set the logger in a `before_initialize` hook, apps will still be able to configure their own logger in an initializer, but the Rails logger will be used by default again if they don't.

There's an existing test which covers this behaviour, but the logger was being set to false before it ran, which doesn't accurately reflect the default configuration any more. Instead of trying to reset the logger before the test, we can temporarily swap in a new configuration object.

### DIFF
--- a/lib/raven/integrations/rails.rb
+++ b/lib/raven/integrations/rails.rb
@@ -32,10 +32,13 @@ module Raven
       end
     end
 
+    config.before_initialize do
+      Raven.configuration.logger = ::Rails.logger
+    end
+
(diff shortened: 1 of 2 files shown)


nateberkopec:
Thank you!

PR #639, opened 2017-02-28
Allow for customisation of the Faraday Adapter

StupidCodeFactory:
**CHANGES:**

We wanted to customise the faraday adapter which currently didn't seem possible.
This allows to provide a custom proc that is then called with the Faraday builder when creating the HTTP connection

### DIFF
--- a/spec/raven/transports/http_spec.rb
+++ b/spec/raven/transports/http_spec.rb
@@ -46,4 +46,17 @@ describe Raven::Transports::HTTP do
 
     stubs.verify_stubbed_calls
   end
+
+  it 'allows to customise faraday' do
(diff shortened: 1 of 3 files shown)


spec/raven/transports/http_spec.rb
```diff
@@ -46,4 +46,17 @@
 
     stubs.verify_stubbed_calls
   end
+
+  it 'allows to customise faraday' do
+    builder = spy('farady_builder')
```
nateberkopec: Typo - `faraday_builder`

StupidCodeFactory: @nateberkopec oops, it's been fixed now.

PR #651, opened 2017-03-13
fix #650 load rack-timeout only if it wasn't loaded yet

### LINKED ISSUE #650: Rack-timeout integration doesn't get loaded
rack-timeout 0.4.2, sentry-raven 2.4.0, results with:

`WARN -- sentry: ** [Raven] Unable to load raven/integrations/rack-timeout: An observer with the id :logger is already set.`

That's caused by `require "rack/timeout/base"` in `integrations/rack-timeout.rb` as rack-timeout was already loaded.

pawelniewie:


nateberkopec:
Something else is going on here. `require` shouldn't load the file if it has already been required. Let's go back to the issue and try to get a reproducing case.

PR #656, opened 2017-03-30
Add PUT and PATCH data processors.

schneiderderek:
Added processors for sanitizing the body of PUT and PATCH requests.


nateberkopec:
Definitely OK with this, but rather than separate processor classes, I'd prefer we have a single RequestMethodProcessor.

schneiderderek:
@nateberkopec OK, just to be clear would you want all 3 (POST, PUT, and PATCH) combined into one processor or just PUT and PATCH?

nateberkopec:
@schneiderderek All three. Ideally you would have a constant or something that was just an array of ["POST", "PUT", "PATCH"] that you would use to show which HTTP methods you were scrubbing. Bonus points if you make it configurable.

schneiderderek:
@nateberkopec Ok, my intent with the current design was to provide flexibility and configurability for the end user while using the existing methods for configuring the way data is processed (using processors). IMO by combining these into one processor we lose that. I'll combine them if you disagree.

When you mention making it configurable, were you thinking something like this?:
```ruby
Raven.configure do |config|
  # Probably needs a better config name
  config.request_methods_sanitized_data = ["POST", "PUT"] 
end
```

nateberkopec:
Yup, exactly like that.

schneiderderek:
@nateberkopec I've updated the code to combine everything into 1 processor. Let me know what you think.

docs/config.rst
```diff
@@ -131,12 +131,12 @@ Optional settings
...
         end
 
-    By default, Sentry does not send POST data or cookies if present. To re-enable, remove the respective processor from the chain:
+    By default, Sentry does not send POST, PUT, PATCH data or cookies if present. To re-enable, remove the respective processor from the chain:
 
     .. code-block:: ruby
 
         Raven.configure do |config|
-          config.processors -= [Raven::Processor::PostData] # Do this to send POST data
+          config.sanitize_data_for_request_methods -= ["POST", "PUT", "PATCH"] # Do this to send POST, PUT PATCH data
```
nateberkopec: This is a little confusing. Can you make this example show how to only send "PATCH" data, and then add a (commented out) line below to show removing the processor entirely (and mention that this will send ALL request methods).

nateberkopec:
One last comment left, then I'm good on this.

schneiderderek:
@nateberkopec I've updated the docs to reflect your feedback.

docs/config.rst
```diff
@@ -136,7 +136,8 @@ Optional settings
     .. code-block:: ruby
 
         Raven.configure do |config|
-          config.sanitize_data_for_request_methods -= ["POST", "PUT", "PATCH"] # Do this to send POST, PUT PATCH data
+          config.sanitize_data_for_request_methods = %w(PATCH) # Only send data for PATCH request method
+          # config.sanitize_data_for_request_methods = [] # Do this to always send request data
```
nateberkopec: I think if someone wants to always send request data, they should just remove the processor altogether: `config.processors -= [Raven::Processors::RequestMethodData]`.

dschneider-scriptdrop: @nateberkopec Yeah, I believe you are looking at an outdated change. I fixed this in the most recent change.

schneiderderek: @nateberkopec Yeah, I believe you are looking at an outdated change. I fixed this in the most recent change.

nateberkopec:
Reverting because I want to put this off for 3.0

schneiderderek:
@nateberkopec Is there a 3.0 RC branch available so I could use this functionality now? Or should I just stick to using my fork?

nateberkopec:
stick to your fork, I won't have something "release-candidate-worthy" for a while

nateberkopec:
also on your fork, you can be reasonably certain that the API you created here won't change and will not be appearing on master until 3.0. so don't worry about conflicts in the meantime with 2.x.

dhirendrx:
@nateberkopec We still haven't included it in the `4.6` release, do we have any plan to include it in a future release.

st0012:
@dhirendrx hey I'm the current maintainer of the SDK 👋 
since version 4.0 (`sentry-ruby`), the concept of SDK-side sanitization (processors) has been removed (like other Sentry SDKs). and the old SDK `sentry-raven` has entered maintenance mode now. so we won't introduce this feature to either one of them in the future.

PR #657, opened 2017-03-30
Only clear context in SidekiqCleanupMiddleware if no exception was raised

tarmo:
Fixes issue referenced at https://github.com/getsentry/raven-ruby/pull/637#issuecomment-290391560

Without this change `SidekiqCleanupMiddleware` clears out context before Sidekiq error handler has a chance to report it.

### DIFF
--- a/lib/raven/integrations/sidekiq.rb
+++ b/lib/raven/integrations/sidekiq.rb
@@ -5,9 +5,17 @@ module Raven
   class SidekiqCleanupMiddleware
     def call(_worker, _job, _queue)
       yield
+    rescue Exception
+      exception_raised = true
+      raise
     ensure
-      Context.clear!
-      BreadcrumbBuffer.clear!
+      # If an exception was raised cleanup must
+      # happen after the exception is reported
+      # in SidekiqErrorHandler#call
+      unless exception_raised
+        Context.clear!
+        BreadcrumbBuffer.clear!
+      end
     end
   end
 
@@ -22,6 +30,9 @@ module Raven
         :extra => { :sidekiq => context },
         :culprit => culprit_from_context(context)
       )
+    ensure
+      Context.clear!
+      BreadcrumbBuffer.clear!
     end
 
     private


drewish:
Could you spell out a test case for this similar to what I'd done in #637 to demonstrate the problem you're encountering? Based on that comment I'm not sure I'd run into it and I'd like to make sure the issue isn't from something else (version of sidekick, active job, etc).

tarmo:
@drewish the test-case would be identical to what you had in #637 but SadWorker should also register a tag like this:

```ruby
class SadWorker
  include Sidekiq::Worker

  sidekiq_options :retry => false

  def perform(*args)
    Raven.tags_context mood: 'sad'
    raise "I'm sad!"
  end
end
```

And this tag is what never reaches sentry. Looking at the code with debugger the order of execution looks like this:

```
    6:   def perform(*args)
    7:     Raven.tags_context mood: 'sad'
    8:     debugger
=>  9:     raise "I'm sad!"
   10:   end
   11: end
(byebug) cont

[5, 14] in /Users/tarmo/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/bundler/gems/raven-ruby-3dde5b735b4c/lib/raven/integrations/sidekiq.rb
    5:   class SidekiqCleanupMiddleware
    6:     def call(_worker, _job, _queue)
    7:       yield
    8:     ensure
    9:       debugger
=> 10:       Context.clear!
   11:       BreadcrumbBuffer.clear!
   12:     end
   13:   end
   14:
(byebug) Raven.tags_context
{:mood=>"sad"}
(byebug) cont
   15:   class SidekiqErrorHandler
   16:     ACTIVEJOB_RESERVED_PREFIX = "_aj_".freeze
   17:
   18:     def call(ex, context)
   19:       debugger
=> 20:       context = filter_context(context)
   21:       Raven.capture_exception(
   22:         ex,
   23:         :message => ex.message,
   24:         :extra => { :sidekiq => context },
(byebug) Raven.tags_context
{}
```

The last line demonstrates that by the time the error handler is invoked there is nothing returned by Raven.tags_context.

lib/raven/integrations/sidekiq.rb
```diff
@@ -5,9 +5,14 @@ module Raven
   class SidekiqCleanupMiddleware
     def call(_worker, _job, _queue)
       yield
+    rescue Exception
+      exception_raised = true
+      raise
     ensure
-      Context.clear!
-      BreadcrumbBuffer.clear!
+      unless exception_raised
```
drewish: We should have a comment explaining the division of labor between the two middlewares. It's not obvious why this condition is necessary. 

Another other option would be to name the variable something that would make the relation clearer, e.g. naming it `context_cleared_by_error_handler`

lib/raven/integrations/sidekiq.rb
```diff
@@ -22,6 +27,9 @@ def call(ex, context)
         :extra => { :sidekiq => context },
         :culprit => culprit_from_context(context)
       )
+    ensure
```
tarmo: True, however it was there in the code before your change: https://github.com/getsentry/raven-ruby/pull/637/files#diff-f18bf682cd7bdb47e4e6bdd40c800adcL16  and it does seem like it might be good to account for the possibility of the exception reporting code itself to also raise an exception, however unlikely and annoying that might be. There's also no obvious downside to this approach that I can think of.

lib/raven/integrations/sidekiq.rb
tarmo: I added a comment to explain this: https://github.com/getsentry/raven-ruby/pull/657/commits/72544b7ef73ed319e39ce9bd0127894ebe8b2984#diff-f18bf682cd7bdb47e4e6bdd40c800adcR12

PR #659, opened 2017-04-02
Fix wrong Raven.user_context usage breaking event sending

syndbg:
When using Raven.user_context with `options` argument that's not a Hash,
the Raven::Event#initialize method will raise an Exception
and prevent an event being sent to Sentry at all.

Of course, this is not acceptable for an exception-sending service,
therefore log a warning that the Raven.user_context `options` argument
will be ignored to ensure that an event will be sent to Sentry.

@nateberkopec 

P.S The reason for this PR is cause I got bit hard by this bug and had 100+ major exception events ignored. :(

### DIFF
--- a/spec/raven/event_spec.rb
+++ b/spec/raven/event_spec.rb
@@ -347,6 +347,7 @@ describe Raven::Event do
   end
 
   describe '.initialize' do
+    # NOTE: This test is completely pointless.
     it 'should not touch the env object for an ignored environment' do
       Raven.configure do |config|
         config.current_environment = 'test'
--- a/spec/raven/instance_spec.rb
+++ b/spec/raven/instance_spec.rb
@@ -219,6 +219,43 @@ describe Raven::Instance do
     end
   end
 
+  describe '.user_context' do
+    context 'with `options` that is a hash' do
+      let(:options) { { 'email' => 'foobar@example.com' } }
+
+      it(:aggregate_failures) do
+        expect(subject.logger).not_to receive(:warn)
+
+        subject.user_context(options)
+
+        expect(
+          subject.context.user
+        ).to eq(options)
+      end
+    end
+
+    context 'with `options` that is not a hash' do
+      let(:options) { ['email', 'foobar@example.com'] }
+
+      it(:aggregate_failures) do
+        expect(subject.logger).to receive(:warn)
+          .once
+          .with(
+            "You're passing `options` - #{options.inspect}" \
+            ' that are not a Hash. Your `options` will be replaced with' \
+            ' an empty Hash to ensure that an event will be sent to Sentry.io'
+          )
+          .and_call_original
+
+        subject.user_context(options)
+
+        expect(
+          subject.context.user
+        ).to eq({})
+      end
+    end
+  end
+
   describe '.last_event_id' do
     let(:message) { "Test message" }
 
--- a/lib/raven/instance.rb
+++ b/lib/raven/instance.rb
@@ -171,7 +171,16 @@ module Raven
     # @example
     #   Raven.user_context('id' => 1, 'email' => 'foo@example.com')
     def user_context(options = nil)
-      context.user = options || {}
+      context.user = if options.respond_to?(:has_key?)
+                       context.user = options
+                     else
+                       logger.warn(
+                         "You're passing `options` - #{options.inspect}" \
+                         ' that are not a Hash. Your `options` will be replaced with' \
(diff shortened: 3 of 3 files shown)


spec/raven/event_spec.rb
```diff
@@ -347,6 +347,7 @@ def raven_context
   end
 
   describe '.initialize' do
+    # NOTE: This test is completely pointless.
```
syndbg: Just an observation.

spec/raven/instance_spec.rb
```diff
@@ -219,6 +219,43 @@ def ivars(object)
     end
   end
 
+  describe '.user_context' do
+    context 'with `options` that is a hash' do
```
syndbg: Arguments can be made whether the context is right given that the check is `respond_to?(:has_key?)`, but lets be real here - this is a Hash structure.

spec/raven/event_spec.rb
nateberkopec: Why?

syndbg: There's no `rspec` expectation here.

nateberkopec: The expectation is that the code will not raise an exception. You can look back through the `git blame` for why.

syndbg: There's https://www.relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/raise-error-matcher for that.

syndbg:
@nateberkopec Regardless of that specific test are there any blockers for this to get merged and released?

spec/raven/event_spec.rb
nateberkopec: There's no point. Here, let me implement `assert_doesnt_raise`:

```ruby
def assert_doesnt_raise
  yield
end
```

nateberkopec:
I need to take a look at this more. I don't like the solution, though I agree we should do something more here.

There are a million possible ways to misuse the client in ways that could cause an exception, so we should look at the whole picture rather than just this particular instance.

spec/raven/event_spec.rb
syndbg: Still, there's RSpec matcher for that.

And the reason it's used is not because you can't `yield` from a block, of course you can and I get that you'll get an exception either way.

The point is that the matcher will:

1st - autogenerate a meaningful test description if you don't manually set one
2nd - make it obvious what you're trying to achieve here.


The point is that this test does not expect anything against the actual `subject`.

`should not touch the env object for an ignored environment` can replace the word `env` with `UFO` and still it would be true, cause we don't expect anything here.

syndbg: Btw a negative test would make it more obvious why there's a need for a `raise` matcher. (and this is a test coverage miss atm)

syndbg:
@nateberkopec Well, I can say the different approaches I had in mind.

Approach A:

Try to convert the given data structure e.g. Array that is received as an argument `options` to Hash, when using `Raven.user_context`. 

Problems: Uphill battle with all the data structures (and even objects that can be passed) and we're never going to be able to make everyone happy. Blame the usage and still not send events? I don't think so.

Approach B:

Don't care about the data structure given when using `Raven.user_context` and instead make `Raven::Event` "safer" to data structures different than Hash.

Problems: Instead of losing user contexts, we're going to lose Sentry data in the `user_context`  such as `ip_address` (which in fact is the only data added to the `user_context` from what I see). However the complexity of the code will increase for no business value imo.

Approach C:

The one I chose in the PR. Instead of taking the risks in A and B, discard the `options` argument as soon as it's given when using `Raven.user_context`. Log a message and proceed with the normal `Raven::Event` initialization and sending.

Problems: Also loses user defined context data, but at least it logs quite often given the usage of the developer is mostly around setting `Raven.user_context` in a `before_action` (if using Rails), or right before sending an event in any other Ruby service/application code.

spec/raven/event_spec.rb
nateberkopec: > Btw a negative test would make it more obvious why there's a need for a raise matcher.

I absolutely agree with this and would accept a PR to that effect.

syndbg: @nateberkopec Don't quote me on this, but I would take a look at it in the weekend, most likely.

nateberkopec:
Thanks, but no thanks. Sentry's policy is log to stderr when theres an internal error, and that's what we've done and will continue to do.

syndbg:
@nateberkopec And this still doesn't solve the problem. The logging to stdeer is the last problem here.

PR #675, opened 2017-05-12
Allow disabling HTTP body truncation

jlogsdon:
We're using an On-Premise install and would like to prevent truncation of the raw body. I considered requiring truncation if the DSN points to `sentry.io` but decided to ask if that would be proper here first. Another consideration would be changing this from a boolean to an integer configuration of the max length.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -221,6 +221,14 @@ Optional settings
           AdminMailer.email_admins("Oh god, it's on fire!").deliver_later
         }
 
+.. describe:: truncate_http_body
+
+    By default the HTTP body of a request is truncated to 2048 characters. It can be disabled.
+
+    .. code-block:: ruby
+
+        config.truncate_http_body = false
+
 Environment Variables
 ---------------------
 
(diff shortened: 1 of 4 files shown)


dcramer:
I have mixed feelings here-

1. Sentry lets you configure this technically because the settings are exposed
2. Generally configuring this to anything outside of the server defaults is dangerous and product may not work correctly (especially with larger values)

I think disabling it isn't useful, and if we were going to do anything I think it should be making it configurable. That said, in other SDKs we've been moving towards the stance of "implement truncation per the server's defaults", and one day will likely rely on the server telling our agents what those values are.

jlogsdon:
Ah, I didn't look closely enough for the actual server config. I did see some discussion on the raven-js repo with regards to the the server indicating the value.

While I agree with you on the points given we would really like to at least configure the SDK to match the server's setting (which defaults to 16kb if I'm looking at this correctly). We could use a fork enabling this for us specifically, but we try to keep on upstream releases as much as possible (for obvious reasons). If there's a timeline for respecting the server's configuration that's reasonable I would be okay using a fork for a short time, but the overhead of keeping the fork in sync with any updates can turn into a lot (especially with V3 of this SDK in progress).

Would you accept the variation where the configuration is for truncation length with some clear documentation that it is only supported for On-Premise deploys?

PR #682, opened 2017-05-24
Fix stack_trace error in ruby 2.4.1

davidlawson:
```/Users/david/.rvm/gems/ruby-2.4.1/bundler/gems/raven-ruby-cbf2d6033c3e/lib/raven/interfaces/stack_trace.rb:76:in `start_with?': no implicit conversion of Pathname into String (TypeError)```

Seems to be related to #113

### DIFF
--- a/lib/raven/interfaces/stack_trace.rb
+++ b/lib/raven/interfaces/stack_trace.rb
@@ -73,7 +73,7 @@ module Raven
       end
 
       def longest_load_path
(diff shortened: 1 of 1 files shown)


nateberkopec:
whoops, but we really need a test on this so we don't break it again.

nateberkopec:
Closed by cf7a35d2dd79

PR #691, opened 2017-06-05
Fix raven test executable not being included in gemspec

### LINKED ISSUE #688: Test tasks and exe don't work as expected

nateberkopec:
+ Fix docs not being included in gem
+ DRY up development dependencies

Closes #688


nateberkopec:
Wow, these builds are a million times faster. I must have been breaking the bundler cache before.

PR #696, opened 2017-06-14
Convert binary

### LINKED ISSUE #689: Encoding::UndefinedConversionError while tracking something that contains Unicode characters
While tracking, either with capture_exception() or capture_message(), works fine generally, there were times where it would fail to report an error.

Here are the errors and the relevant backtrace:

**Encoding::UndefinedConversionError in** (method name...)

**"\xE2" from ASCII-8BIT to UTF-8**

```
Application Trace
app/controllers/application_controller.rb:64:in `respond_with_server_error'
app/controllers/application_controller.rb:31:in `block in <class:ApplicationController>'


Framework Trace
json (1.7.7) lib/json/common.rb:254:in `encode'
json (1.7.7) lib/json/common.rb:254:in `generate'
json (1.7.7) lib/json/common.rb:254:in `fast_generate'
sentry-raven (2.5.1) lib/raven/client.rb:67:in `encode'
sentry-raven (2.5.1) lib/raven/client.rb:37:in `send_event'
sentry-raven (2.5.1) lib/raven/instance.rb:81:in `send_event'
sentry-raven (2.5.1) lib/raven/instance.rb:126:in `capture_type'
activesupport (4.2.4) lib/active_support/rescuable.rb:114:in `instance_exec'
activesupport (4.2.4) lib/active_support/rescuable.rb:114:in `block in handler_for_rescue'
activesupport (4.2.4) lib/active_support/rescuable.rb:80:in `call'
activesupport (4.2.4) lib/active_support/rescuable.rb:80:in `rescue_with_handler'
....
```

Now, this seemed to happen only when the payload contained some funky characters, perhaps 4-byte unicode characters (emojis).

As you can see above it was triggered with json (1.7.7) gem, but I upgraded to latest json gem and it still happened.

You can probably reproduce this by including some emojis such as  ▷▷  ⬧  ✈ in the messgae.

Also, FYI I reverted to the earlier version raven-ruby 2.4.0 and this fixed the problem. It did happen with gem versions 2.5.0 and 2.5.1.

If you cannot reproduce this, it might be that my ruby version is to blame ― granted it's a bit old: ruby 2.1.1p76 (2014-02-24 revision 45161) [x86_64-linux]  In that case let me know so that I can do more tests with newer ruby versions.

nateberkopec:
Close #689

### DIFF
--- a/lib/raven/processor/utf8conversion.rb
+++ b/lib/raven/processor/utf8conversion.rb
@@ -18,6 +18,13 @@ module Raven
         clean_exc.set_backtrace(value.backtrace)
         clean_exc
       when String
+        # Encoding::BINARY / Encoding::ASCII_8BIT is a special binary encoding.
+        # valid_encoding? will always return true because it contains all codepoints,
+        # so instead we check if it only contains actual ASCII codepoints, and if
+        # not we assume it's actually just UTF8 and scrub accordingly.
+        if value.encoding == Encoding::BINARY && !value.ascii_only?
+          value.force_encoding(Encoding::UTF_8)
+        end
         return value if value.valid_encoding?
         remove_invalid_bytes(value)
       else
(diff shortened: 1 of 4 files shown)


sgrif:
Yo I'm just some rando that you pinged, but your commit messages are pretty short. ;)

PR #708, opened 2017-07-11
Add `user` to the available additional context attributes list

kyanny:
Hi, I found that `Raven.capture_exception` receives `user` context option.
But it is not listed at [document](https://docs.sentry.io/clients/ruby/context/), so I made a patch.

```ruby
Raven.capture_exception(e, user: { id: 123 }) # it works just like Raven. user_context(id: 123)
```

Ref: https://github.com/getsentry/raven-ruby/blob/master/lib/raven/event.rb#L46


nateberkopec:
Thank you!

PR #710, opened 2017-07-11
Activejob refix

nateberkopec:
Yikes! We haven't properly integrated with ActiveJob for years. Time to fix that!

### DIFF
--- a/lib/raven/integrations/rails/active_job.rb
+++ b/lib/raven/integrations/rails/active_job.rb
@@ -1,31 +1,53 @@
 module Raven
   class Rails
-    module ActiveJob
+    module ActiveJobExtensions
+      ALREADY_SUPPORTED_SENTRY_ADAPTERS = %w(
+        ActiveJob::QueueAdapters::SidekiqAdapter
+        ActiveJob::QueueAdapters::DelayedJobAdapter
+      ).freeze
+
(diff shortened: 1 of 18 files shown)


pjg:
Thanks for this integration!

lib/raven/integrations/rails/active_job.rb
```diff
@@ -1,31 +1,53 @@
...
-            # The same exception twice.
-            unless self.class.queue_adapter.to_s == 'ActiveJob::QueueAdapters::SidekiqAdapter'
-              active_job_details = {
-                :active_job => self.class.name,
-                :arguments => arguments,
-                :scheduled_at => scheduled_at,
-                :job_id => job_id,
-                :locale => locale
-              }
+          around_perform do |job, block|
```
sharshenov: @nateberkopec I was browsing source code and found that this change can _potentially_ lead to Sentry ignoring errors because ActiveJob callbacks are not chained.
[Calling the same callback multiple times will overwrite previous callback definitions](https://github.com/rails/rails/blob/v5.2.3/activejob/lib/active_job/callbacks.rb#L18)

sharshenov: Never mind. Everything works. Sorry for tagging you

PR #714, opened 2017-07-18
Make message/http interface sizes match current server defaults

nateberkopec:
Close #675


jlogsdon:
Thank you for following up on this!

nateberkopec:
Sorry it took so long!

PR #719, opened 2017-07-22
Defer to rescuable/rescue_from inside of ActiveJob

bensheldon:
This PR ensures that ActiveJob exceptions that are rescued using `rescue_from` are not sent to Sentry.

### DIFF
--- a/lib/raven/integrations/rails/active_job.rb
+++ b/lib/raven/integrations/rails/active_job.rb
@@ -17,6 +17,7 @@ module Raven
       def capture_and_reraise_with_sentry(job, block)
         block.call
       rescue Exception => exception # rubocop:disable Lint/RescueException
+        return if rescue_with_handler(exception)
         return if already_supported_by_specific_integration?(job)
         Raven.capture_exception(exception, :extra => raven_context(job))
         raise exception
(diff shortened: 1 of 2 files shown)


nateberkopec:
Would this have the effect of calling the rescue_from handler twice?

bensheldon:
Good question. I _think_ `rescue_with_handler` will prevent the exception from bubbling up further and re-triggering. I'll insert a spy into the `rescue_from` and assert it's only called once.

bensheldon:
I just pushed a commit that asserts that the `rescue_from` callback is called only once.

nateberkopec:
There are some problems here.

* This does call the rescue handler twice, despite what rSpec seems to say here.
* We can't call rescue handlers twice, because they may not be idempotent.
* We can't tell if rescue handlers have run from outside.
* The presence of a rescue_from handler does not necessarily mean we shouldn't capture, as the handler may re-raise.

nateberkopec:
I think the reason why your Rspec catch doesn't catch this is because of how`handler_for_rescue` works.

If you change the test class to:

```ruby
class RescuedActiveJob < MyActiveJob
    attr_accessor :called_rescue_from

    rescue_from TestError, :with => :rescue_callback

    def rescue_callback(error)
      raise "Rescue callback called twice!" if called_rescue_from
      self.called_rescue_from = true
    end
  end
``` 

... the test fails.

nateberkopec:
oioioi, ignore everything I said and then deleted. It's been a long day I guess! 😆

Issue #720, opened 2017-07-24
Exceptions coming from env['action_dispatch.exception'] are not caught by Raven::Rack


Ferdy89:
Hi, we're currently using Sentry Raven 2.5.3 on a Rails 4.2.8 application with pretty much no configuration for the gem other than the automatic one it provides. We're noticing that some exceptions are not caught by the `Raven::Rack` middleware.

Here's how our middleware stack looks like in production:
```
use Raven::Rack
# ...
use ActionDispatch::ShowExceptions
# ...
run Rails::Application.routes
```

In production, `Rails.configuration.action_dispatch.show_exceptions` is set to `true` by default, which means [the `ActionDispatch::ShowExceptions` middleware catches and stops any exceptions](https://github.com/rails/rails/blob/v4.2.8/actionpack/lib/action_dispatch/middleware/show_exceptions.rb#L35). It is graceful enough to [leave the exception in the `env`](https://github.com/rails/rails/blob/v4.2.8/actionpack/lib/action_dispatch/middleware/show_exceptions.rb#L44) under the name of `'action_dispatch.exception'`, though.

However, Raven then [fails to look at that part of the `env`, even though it does look at some parts of the `env` for "hidden" exceptions](https://github.com/getsentry/raven-ruby/blob/master/lib/raven/integrations/rack.rb#L58).

We wonder if it's that we're using the gem in a way that it's not designed for or if it's that the gem should be also looking at that part of the `env`. Could you tell us which direction to go through? We'd be happy to submit better reproduction steps or a patch for the gem.

Thank you!

nateberkopec:
> some exceptions

Can you expand on what you mean by that?

Ferdy89:
Sorry about the ambiguity. We have our application to rescue errors at the `ApplicationController` level so we explicitly handle most errors, like:
```ruby
class ApplicationController < ActionController::Base
  rescue_from StandardError do |exception|
    Raven::Rack.capture_exception(exception)
  end
end
```
In this particular case, the error was being generated in the routes, so the controller wasn't able to catch it. That's why I said "some" exceptions, I should have said all exceptions.

Now, I went ahead and created a brand new Rails app and followed these steps for reproduction:
* Create a controller and an action which just raises an exception.
* Add the `sentry-raven` gem to the project.
* Add a key in `config/secrets.yml` to be able to boot in production mode.
[...]

PR #722, opened 2017-07-28
rescue errors occuring in #format_headers_for_sentry

cmoylan:
This is the cause of #720 and the reason we were not seeing this error in Sentry.

`ActionDispatch::RemoteIP` raises an exception when certain IP origination headers do not match. (https://github.com/rails/rails/blob/b70fc698e157f2a768ba42efac08c08f4786b01c/actionpack/lib/action_dispatch/middleware/remote_ip.rb#L134) Sentry was correctly capturing this error, but during `format_headers_for_sentry` was calling `to_s` on the `RemoteIP` object, which re-raised the error.

The exception can be reproduced by setting up the basic Rails app from #720 and running
```
curl -H "X_FORWARDED_FOR: 1.2.1.1" \
-H "CLIENT_IP: 2.2.2.2" \
"localhost:3000"
```

### DIFF
--- a/lib/raven/integrations/rack.rb
+++ b/lib/raven/integrations/rack.rb
@@ -94,22 +94,30 @@ module Raven
 
     def format_headers_for_sentry(env_hash)
       env_hash.each_with_object({}) do |(key, value), memo|
-        key = key.to_s # rack env can contain symbols
-        value = value.to_s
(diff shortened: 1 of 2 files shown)


nateberkopec:
Nice sleuthing! 👏

PR #733, opened 2017-08-07
sanitize_fields accepts a Proc

### LINKED ISSUE #307: Support lambdas for Filter Params
We use a lambda for `Rails.application.config.filter_parameters` so that we can whitelist attributes instead of blacklisting them, but this means we can't pass it into `config.sanitize_fields`. I'm willing to do the work to get a PR put together for this if it's functionality that you're willing to support, though I'm not even completely sure the right direction to start in...

nateberkopec:
Close #307.

On second thought, I'm not even sure this would be useful. What's the benefit over just:

```ruby
config.sanitize_fields = MyProc.call
```

That way you'd at least get to provide arguments if you wanted. 

@GeekOnCoffee I know it's been literally years, but do you have any memory of this or any thoughts? 

If no one chimes in here I think I'll just drop this.

### DIFF
--- a/lib/raven/configuration.rb
+++ b/lib/raven/configuration.rb
@@ -97,7 +97,8 @@ module Raven
     # By default, Sentry censors Hash values when their keys match things like
     # "secret", "password", etc. Provide an array of Strings that, when matched in
     # a hash key, will be censored and not sent to Sentry.
-    attr_accessor :sanitize_fields
+    # Also takes a Proc, but .call on that Proc must return an array of Strings.
+    attr_reader :sanitize_fields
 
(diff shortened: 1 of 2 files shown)


GeekOnCoffee:
@nateberkopec taking a look. It's been years, but I got an email about a bug I filed on a perl package in 2009 the other day, so let's see if I know what this was.

GeekOnCoffee:
I no longer have this need, my guess is that what you listed above would work fine. We were trying to whitelist params, rather than blacklist, so that we didn't accidentally leak data because a param was missed.

PR #736, opened 2017-08-09
Allow `current_environment` to be set via an env variable

tijmenb:
This commit allows developers to set a `SENTRY_CURRENT_ENV` variable to configure the `current_environment` setting.

It serves as a fix for this scenario:

We run our production/staging/integration environments with Rails env "production". To differentiate errors from the environments we set `config.current_environment` with an environment variable:

https://github.com/alphagov/content-tagger/blob/bac2e1903aaf6ac12b3a103490517355673648dd/config/initializers/sentry.rb#L1-L3

However, this doesn't work in all situations. For example, when you run a rake task that doesn't exist, an error will be reported to Sentry. Because the initialiser that sets the `current_environment` hasn't run (yet) Sentry will use the `RAILS_ENV` to determine the current env. This causes all errors to be reported as "production" instead of production/staging/integration.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -65,7 +65,7 @@ Optional settings
 
         config.environments = %w[ production ]
 
-    Sentry automatically sets the current environment to RAILS_ENV, or if it is not present, RACK_ENV. If you are using Sentry outside of Rack or Rails, or wish to override environment detection, you'll need to set the current environment yourself:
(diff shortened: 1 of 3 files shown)


nateberkopec:
Couldn't you also fix this by requiring the initializer at the top of your Rakefile?

tijmenb:
Yep, that would also work in this case. However, we're migrating over 50 applications to Sentry, so for us this would be a not insignificant amount of work. Additionally this option would probably allow us to configure everything via the environment and skip the initialiser entirely, which makes adoption a lot easier for us.

Issue #673, opened 2017-05-11
Don't let 403's go to transport_failure_callback


nateberkopec:
For our client, this is going to be a filtered event 99.99% of the time, which is just noise.

nateberkopec:
probably shouldn't even retry them.

st0012:
the `transport_failure_callback` has been removed from the new SDK, so closing now.

PR #742, opened 2017-08-15
Add Configuration to Silence Send Event Failure Logs

serdary:
Currently 80% of our logs are filled with these logs since we're hitting rate limits a lot.
our log size: `-rw-r--r-- 1 deploy nginx 102331944 Aug 15 00:10 production.log`

Obviously underlying problem is sending too many events and having a smaller plan than required but thought this should still be a config to reduce the log sizes in case bumping up the plan is not convenient for some people.

Since this is my first PR here, please let me know if I'm missing any conventions in this repo. Already ran tests, specs and rubocop. They are all green.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,7 +7,7 @@ AllCops:
      - 'vendor/**/*'
 
 Metrics/ClassLength:
-  Max: 262
+  Max: 263
   CountComments: false
 
 Metrics/AbcSize:
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -212,6 +212,21 @@ Optional settings
 
         config.silence_ready = true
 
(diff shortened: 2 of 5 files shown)


nateberkopec:
I'm trying to avoid adding more configuration options to the already labyrinthine configuration.rb. Instead, here's what I'll do - In 3.0, #673 will be fixed and we'll have a separate code path for transport failures due to 4XX responses from Sentry. That code path will log at a lower level, probably DEBUG or INFO. You can then set your log levels as appropriate.

nateberkopec:
Also, check out the new `sample_rate` option.

serdary:
@nateberkopec sure, we'll wait for 3.0. thanks!

PR #743, opened 2017-08-15
Culprits -> Transactions

### LINKED ISSUE #667: Change Rails web culprit setting
Instead of marking the filename/lineno as transaction, mark it as ClassName#method

nateberkopec:
Close #667.

This also allows anyone to set a culprit by setting `culprit` in the context object.

### DIFF
--- a/docs/context.rst
+++ b/docs/context.rst
@@ -25,6 +25,7 @@ The following attributes are available:
 * ``tags``: a mapping of tags describing this event
 * ``extra``: a mapping of arbitrary context
 * ``user``: a mapping of user context
+* ``transaction``: An array of strings. The final element in the array represents the current transaction, e.g. "HelloController#hello_world" for a Rails controller.
 
 Providing Request Context
 -------------------------
@@ -116,3 +117,25 @@ for you, otherwise you'll need to ensure you perform it manually:
 
 Note: the rack and user context will perform a set operation, whereas tags
 and extra context will merge with any existing request context.
+
+Transactions
+~~~~~~~~~~~~
+
+The "transaction" is intended to represent the action the event occurred during.
+In Rack, this will be the request URL. In Rails, it's the controller name and
+action ("HelloController#hello_world").
+
+Transactions are modeled as a stack. The top item in the stack (i.e. the last
+element of the array) will be used as the ``transaction`` for any events:
+
+.. sourcecode:: ruby
+
+    Raven.context.transactions.push "User Import"
+    # import some users
+    Raven.context.transactions.pop
+
+Transactions may also be overridden/set explicitly during event creation:
+
+.. sourcecode:: ruby
+
+    Raven.capture_exception(exception, transaction: "User Import")
--- a/docs/integrations/puma.rst
+++ b/docs/integrations/puma.rst
@@ -23,7 +23,7 @@ Puma provides a config option for handling low level errors.
         ex,
         :message => ex.message,
         :extra => { :puma => env },
-        :culprit => "Puma"
+        :transaction => "Puma"
       )
       # note the below is just a Rack response
       [500, {}, ["An error has occurred, and engineers have been informed. Please reload the page. If you continue to have problems, contact support@example.com\n"]]
--- a/lib/raven/context.rb
+++ b/lib/raven/context.rb
@@ -10,7 +10,7 @@ module Raven
       Thread.current[:sentry_context] = nil
     end
 
-    attr_accessor :extra, :server_os, :rack_env, :runtime, :tags, :user
+    attr_accessor :transaction, :extra, :server_os, :rack_env, :runtime, :tags, :user
 
     def initialize
       self.server_os = self.class.os_context
@@ -19,6 +19,7 @@ module Raven
       self.rack_env = nil
       self.tags = {}
       self.user = {}
+      self.transaction = []
     end
 
     class << self
--- a/lib/raven/event.rb
+++ b/lib/raven/event.rb
@@ -25,7 +25,7 @@ module Raven
     SDK = { "name" => "raven-ruby", "version" => Raven::VERSION }.freeze
 
     attr_accessor :id, :timestamp, :time_spent, :level, :logger,
-                  :culprit, :server_name, :release, :modules, :extra, :tags,
+                  :transaction, :server_name, :release, :modules, :extra, :tags,
                   :context, :configuration, :checksum, :fingerprint, :environment,
                   :server_os, :runtime, :breadcrumbs, :user, :backtrace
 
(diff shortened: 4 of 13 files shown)


nateberkopec:
Intended to be the "ruby" version of https://github.com/getsentry/raven-python/pull/768

nateberkopec:
We're gonna need a stack as well for things like someone running Sidekiq inline (transaction top-level is the Rails controller action, then lowest-level will be the Sidekiq job.)

nateberkopec:
This PR also fixes some bugs with Sidekiq:

* Fixes transaction name.
* Adds some context about worker class name and queue.

SirRawlins:
@nateberkopec this PR looks to have broken something for me. I have my own around_action callbacks defined within my application.

```ruby
# Wrap all actions in delayed touch.
prepend_around_action :delay_touching

# Delay touching to once per
# transaction to prevent the noise
def delay_touching
    # Use the gem to wrap every action within the app.
    ActiveRecord::Base.delay_touching { yield }
end
```

This worked up until this PR was merged in, all builds after that date result in an error being raised by my own around_action. 

```text
LocalJumpError:
    no block given (yield)
```

Is the new around_action behavior you've defined augmenting the request in a way it shouldn't? Meaning that the block has been stripped by the time it arrives in my own handler?

nateberkopec:
Yeah I'm seeing the same on Codetriage. Working on it.

bjeanes:
I think it's also worth noting that this change broke compatibility with Rails 3. The README states that it works with Rails 4.2+, so I guess this is fair game. Nonetheless, for posterity and future readers, I thought it noteworthy.

Rails 3.x will get the following exception: 

```
NoMethodError: undefined method `around_action' for ActionController::Base:Class
```

nateberkopec:
🤷‍♂️

nateberkopec:
@ragesoss Good point. In my head this wasn't a breaking change because the Sentry server accepts culprit *and* transaction, but if you're manually putting a culprit on the event and then we add a transaction attribute, Sentry will ignore the former. 

Are you saying it raised an exception in production? I don't see how that would occur.

ragesoss:
@nateberkopec Here's what the call site was (the old version): https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/ccff70d99326e48687fc444e006fb8107d7bb324

And the error (which was itself caught by Sentry and reported):
```
NoMethodError: undefined method `culprit=' for #<Raven::Event:0x007f8008d9ee20>
  from raven/event.rb:39:in `public_send'
  from raven/event.rb:39:in `block in initialize'
  from raven/event.rb:39:in `each_pair'
  from raven/event.rb:39:in `initialize'
  from raven/event.rb:83:in `new'
  from raven/event.rb:83:in `from_message'
  from raven/instance.rb:114:in `capture_type'
  from lib/wiki_edits.rb:35:in `notify_untrained'
  from app/workers/notify_untrained_users_worker.rb:16:in `perform'
  from sidekiq/processor.rb:188:in `execute_job'
  from sidekiq/processor.rb:170:in `block (2 levels) in process'
  from sidekiq/middleware/chain.rb:128:in `block in invoke'
  from new_relic/agent/instrumentation/sidekiq.rb:33:in `block in call'
  from new_relic/agent/instrumentation/controller_instrumentation.rb:368:in `perform_action_with_newrelic_trace'
  from new_relic/agent/instrumentation/sidekiq.rb:29:in `call'
  from sidekiq/middleware/chain.rb:130:in `block in invoke'
  from raven/integrations/sidekiq.rb:9:in `call'
  from sidekiq/middleware/chain.rb:130:in `block in invoke'
  from sidekiq_unique_jobs/lock/until_executed.rb:63:in `after_yield_yield'
  from sidekiq_unique_jobs/lock/until_executed.rb:20:in `execute'
  from sidekiq_unique_jobs/server/middleware.rb:18:in `call'
  from sidekiq/middleware/chain.rb:130:in `block in invoke'
  from sidekiq/middleware/chain.rb:133:in `invoke'
  from sidekiq/processor.rb:169:in `block in process'
  from sidekiq/processor.rb:141:in `block (6 levels) in dispatch'
  from sidekiq/job_retry.rb:97:in `local'
  from sidekiq/processor.rb:140:in `block (5 levels) in dispatch'
  from sidekiq/rails.rb:42:in `block in call'
  from active_support/execution_wrapper.rb:85:in `wrap'
  from active_support/reloader.rb:68:in `block in wrap'
  from active_support/execution_wrapper.rb:85:in `wrap'
  from active_support/reloader.rb:67:in `wrap'
  from sidekiq/rails.rb:41:in `call'
  from sidekiq/processor.rb:136:in `block (4 levels) in dispatch'
  from sidekiq/processor.rb:204:in `stats'
  from sidekiq/processor.rb:131:in `block (3 levels) in dispatch'
  from sidekiq/job_logger.rb:7:in `call'
  from sidekiq/processor.rb:130:in `block (2 levels) in dispatch'
  from sidekiq/job_retry.rb:72:in `global'
  from sidekiq/processor.rb:129:in `block in dispatch'
  from sidekiq/logging.rb:44:in `with_context'
  from sidekiq/logging.rb:38:in `with_job_hash_context'
  from sidekiq/processor.rb:128:in `dispatch'
  from sidekiq/processor.rb:168:in `process'
  from sidekiq/processor.rb:85:in `process_one'
  from sidekiq/processor.rb:73:in `run'
  from sidekiq/util.rb:16:in `watchdog'
  from sidekiq/util.rb:25:in `block in safe_thread
```

nateberkopec:
Ah, crap. You're right. I should have left an alias in. Sorry about that.

If you want errors like this to surface in your test suite, Raven needs to be configured to actually capture events in that environment. Of course you probably don't want messages and stuff from the test environment to show up on your Sentry dashboard, so this may be a good time to use the `dummy` transport, which does everything except actually send the event anywhere.

Here's an example, which is how we test Raven in the gem:

```ruby
require "raven/transports/dummy"

Raven.configure do |config|
  config.dsn = "dummy://12345:67890@sentry.localdomain/sentry/42"
  config.encoding = "json"
  logger = Logger.new(nil)
  config.logger = logger
end
```

You can then make assertions against the state of the dummy transport:

```ruby
evt = Raven.client.transport.events.first[1]
assert_equal "test", JSON.parse(evt)["environment"]
```

ragesoss:
@nateberkopec awesome, thanks! I did it in the Raven initializer, like this: https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/a785d7683ed6abeefdae6b4662af1e4ea2000b23

ragesoss:
Whoops. Just in case anyone reading this issue tries to follow that example... `dummy:` transport doesn't work for raven-js. https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/794f2254297567ead9c909c569fff0daa5739e6e

nateberkopec:
Ah yeah, `dummy` is just a thing I implemented in the Ruby client, it isn't a standard part of the client SDK (though maybe it should be?)

PR #750, opened 2017-08-15
Add `after_send_event` post send hook

tijmenb:
This adds a hook that is called after an event is successfully sent to Sentry.

It is mutually exclusive with `transport_failure_callback`. Together with that hook this fulfils the "Post-send hook" feature in the list of [expected features in clients][1]:

> Hook called on success or error, and passed the event and exception so that users can perform actions based on events actually being sent.

I've looked in the other client libraries for a good name for this and found 2 similar configs: `after_send_event` is used by the [elixir library][2], and `send_callback` is used by [PHP][3]. I liked `after_send_event` slightly better.

[1]: https://docs.sentry.io/clientdev/features/#post-send-hook
[2]: https://docs.sentry.io/clients/elixir/config/
[3]: https://docs.sentry.io/clients/php/config/

Closes https://github.com/getsentry/raven-ruby/issues/737.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,7 +7,7 @@ AllCops:
      - 'vendor/**/*'
 
 Metrics/ClassLength:
-  Max: 262
+  Max: 270
   CountComments: false
 
 Metrics/AbcSize:
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -238,6 +238,16 @@ Optional settings
           AdminMailer.email_admins("Oh god, it's on fire!").deliver_later
         }
 
+.. describe:: after_send_event
+
+    After the event is sent to Sentry successfully, this Proc or lambda will be called.
+
+    .. code-block:: ruby
+
+        config.after_send_event = lambda { |event|
+          Redis.current.incr('errors_reported')
+        }
+
 Environment Variables
 ---------------------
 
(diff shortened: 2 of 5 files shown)


nateberkopec:
Looks good, not 100% sure about the name. 

Here's our present and future callbacks:

| Name             | When?           | Implemented? |
| ------------- | ------------- | -------------- |
| async             | After capture  | Yes                |
| should_capture  | Before capture  | Yes           |
| transport_failure_callback | After failed HTTP | Yes |
| before_send_event | After capture, before send | No |
| after_send_event | After send | No |

nateberkopec:
It feels like transport_failure_callback and should_capture are really before_send and after_send with less functionality.

tijmenb:
Sorry for the late reply, @nateberkopec. What are you thinking for the name? `before_send` & `after_send`? 

I agree that that these hooks are the same as the `transport_failure_callback` and `should_capture`. In fact, we're misusing `should_capture` in our apps to report things to statsd here:

https://github.com/alphagov/govuk_app_config/blob/1c16a63710962c3bf4db15a689cfa2faacdf86b0/lib/govuk_app_config/configure.rb#L9-L19

Not sure if it's worth replacing them though.

nateberkopec:
I still want to add these functions to the `v3` branch.

tijmenb:
Cool, anything I can do to help?

PR #761, opened 2017-09-11
Add ActionController::UnknownFormat to ignore exceptions

iguchi1124:
ActionController::UnknownFormat is rescued 406 http response from rails 5.

https://github.com/rails/rails/issues/20666

so, I added this exception to default ignore exceptions. what do you think?  @nateberkopec

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,8 +7,7 @@ AllCops:
      - 'vendor/**/*'
 
 Metrics/ClassLength:
-  Max: 262
-  CountComments: false
+  Enabled: false
 
 Metrics/AbcSize:
   Max: 58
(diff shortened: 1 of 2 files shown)


nateberkopec:
Hmmm... sometimes I like receiving these because they're usually bad bots that I want to block. I guess I'm not sure.

iguchi1124:
I don't want to notify this error because this is not a server error. 
It has better to make configure in the application's `Rave.config.excluded_exceptions` if someone want to notify this error for security reasons.

I think this proposal is same as ignoring `ActionController::InvalidAuthenticityToken`. 💭

nateberkopec:
Closing in favor of #769, but thanks for opening (otherwise wouldn't have had the idea!)

iguchi1124:
thanks!

PR #769, opened 2017-10-09
Exclude all 4xx Rails errors

nateberkopec:
All Rails 4xx errors will be excluded with this change. I think this generally makes sense. Sentry's behavior, by default, should be to report programmer errors (bugs). A program rejecting invalid input (4xx) is not really an exception, and the program is behaving as intended.

+ Alphabetized the list
+ Added some comments

### DIFF
--- a/lib/raven/configuration.rb
+++ b/lib/raven/configuration.rb
@@ -157,16 +157,30 @@ module Raven
     # Errors object - an Array that contains error messages. See #
     attr_reader :errors
 
+    # Most of these errors generate 4XX responses. In general, Sentry clients
+    # only automatically report 5xx responses.
     IGNORE_DEFAULT = [
       'AbstractController::ActionNotFound',
+      'ActionController::BadRequest',
       'ActionController::InvalidAuthenticityToken',
+      'ActionController::InvalidAuthenticityToken',
+      'ActionController::InvalidCrossOriginRequest',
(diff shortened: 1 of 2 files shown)


cesc1989:
We use Sentry in production and excluding 4xx would be kind of removing noise from the issues list. Those 4xx could be found by tests or QA prior to going to production. I think this would be good.

Came from twitter.

nateberkopec:
Thanks for the feedback, all 👍

lib/raven/configuration.rb
```diff
@@ -157,16 +157,30 @@ class Configuration
     # Errors object - an Array that contains error messages. See #
     attr_reader :errors
 
+    # Most of these errors generate 4XX responses. In general, Sentry clients
+    # only automatically report 5xx responses.
     IGNORE_DEFAULT = [
       'AbstractController::ActionNotFound',
+      'ActionController::BadRequest',
       'ActionController::InvalidAuthenticityToken',
+      'ActionController::InvalidAuthenticityToken',
```
tijmenb: This one is already being ignored.

nateberkopec: TODO: fix that, also see if there's any common ancestors to these exception classes so I can just ignore those instead.

gingerlime:
Any chance of backporting this to `v2` ? (or when is `v3` due?)

PR #782, opened 2017-10-21
Deep merge in #annotate_exception

janklimo:
#### Changes
1. I'd like to propose to perform a deep merge inside `annotate_exception` so no existing context is lost by accident.
2. Added a spec to make sure the Sidekiq error handler does perform a deep merge on the existing context.


nateberkopec:
Simple and reasonable. Thanks!

Issue #781, opened 2017-10-20
Don't send an exception until retries are exhausted if the exception is retryable with Sidekiq


brettburley:
We tend to have jobs that raise exceptions on occasion due to issues like network timeouts, etc. that we expect to happen regularly. Given a list of `retryable_exceptions`, we might want jobs that raise one of these exceptions to not send an error notification until their last retry is exhausted. This way we can eliminate extra noise from errors that we expect to happen, but still allow these jobs to retry (and eventually succeed), while still receiving immediate notifications for unexpected errors.

This could be implemented in the Sidekiq integration to not send an error report if the exception is in the list of `retryable_exceptions` and the job has remaining retries.

I'm happy to take a stab at a PR that integrates this, but curious if this seems like it is within the scope of an integration, or if the integrations are intended to be very simple, and if there's a precedence for adding configuration options like `retryable_exceptions` that's specific to a particular integration.

nateberkopec:
Agreed. I think this is within the scope of the Sidekiq integration.

kornate:
@st0012 could you give a status update on this? Is it expected to be integrated? As a new customer, we have the same issue whereby timeouts and other commonly retried errors, are costing us a fortune!

st0012:
@brettburley @kornate sorry for the delay. I've implemented this feature in https://github.com/getsentry/sentry-ruby/pull/1532, which is scheduled to be a part of 4.7.0 release 🙂

travisp:
@st0012 this update is useful, but would you consider something that also includes some of the original request to be able to specify which exceptions should be sent? There are many cases where it's useful to ignore certain exceptions that will be retried (until exhaustion), but still report other exceptions.

st0012:
@travisp I can understand the use case, but I won't put that in high priority for now. but if anyone can propose an implementation for it, I'll consider it.

PR #784, opened 2017-10-28
Allow Sidekiq jobs with remaining retries to not notify. (#781)

brettburley:
This prevents specific exceptions from being notifying when there are remaining retries left on a job in the Sidekiq integration. For example, you might configure with:
```
Raven.configure do |config|
  config.retryable_exceptions = [HTTP::TimeoutError, Errno::ECONNRESET]
end
```
See #781 for more details.

Any suggestions as to how I could test with the "Sidekiq full-stack integration" part of the sidekiq spec? In particular, it doesn't appear that the Raven configuration is reset between tests, and other tests don't perform a `Raven.configure`, so I'm not sure how bad it would be to set configuration in the spec to add a test `retryable_exception`.

### DIFF
--- a/lib/raven/configuration.rb
+++ b/lib/raven/configuration.rb
@@ -91,6 +91,10 @@ module Raven
     # We automatically try to set this to a git SHA or Capistrano release.
     attr_accessor :release
 
+    # Array of exception classes that should not be reported until all retries
+    # are exhausted for a Sidekiq job.
+    attr_accessor :retryable_exceptions
+
     # The sampling factor to apply to events. A value of 0.0 will not send
     # any events, and a value of 1.0 will send 100% of events.
     attr_accessor :sample_rate
@@ -198,6 +202,7 @@ module Raven
       self.project_root = detect_project_root
       self.rails_activesupport_breadcrumbs = false
       self.rails_report_rescued_exceptions = true
+      self.retryable_exceptions = []
       self.release = detect_release
       self.sample_rate = 1.0
       self.sanitize_credit_cards = true
@@ -307,6 +312,10 @@ module Raven
       else
         true
       end
+    end
+
+    def retryable_exception?(exc)
+      retryable_exceptions.any? { |x| get_exception_class(x) === exc }
     end
 
     private
--- a/lib/raven/integrations/sidekiq.rb
+++ b/lib/raven/integrations/sidekiq.rb
@@ -1,5 +1,10 @@
 require 'time'
 require 'sidekiq'
+begin
+  # Sidekiq 5 introduces JobRetry and stores the default max retry attempts there.
+  require 'sidekiq/job_retry'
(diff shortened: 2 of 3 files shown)


lib/raven/configuration.rb
```diff
@@ -198,6 +202,7 @@ def initialize
       self.project_root = detect_project_root
       self.rails_activesupport_breadcrumbs = false
       self.rails_report_rescued_exceptions = true
+      self.retryable_exceptions = []
```
brettburley: These changes fail two lint rules:
1. `Assignment Branch Condition size for initialize is too high`, can we just increase this from 30 to ~32? Otherwise, I'm not sure if there's a better place to put this configuration.
2. `Class has too many lines`, not sure if you're OK increasing this limit as well, or if we should try pulling out some of the common functionality. For example, `retryable_exception?` and `excluded_exception?` share logic using `get_exception_class`, which we could pull out into a `util` or similar.

nateberkopec:
I think this sort of behavior can already be accomplished with `should_capture`. I'll write up an example and put it in the docs.

brettburley:
@nateberkopec good point. Looking at the implementation of `should_capture`, it looks like it's only currently passed the exception itself, and not the additional options added to `capture_exception` in the Sidekiq exception handler. However, that information is tagged on `Raven.extra_context[:sidekiq]` -- would you recommend just reading it from there within the capture handler?

gingerlime:
@brettburley thanks for the PR. curious how you managed to resolve this eventually? I bumped into the same problem, and want to find a way to log errors to sentry, but not every retry, but perhaps only if it fails for at least X retries... I guess with `should_capture` and digging into the extra sidekiq context, I can do something, but it feels a bit cumbersome for something that should evaluate quickly, and also, as @nateberkopec [commented on](https://github.com/getsentry/raven-ruby/issues/781#issuecomment-338714995) before, shouldn't this really be part of the sidekiq integration?

gingerlime:
Rollbar seems to do this with a configurable [sidekiq_threshold](https://github.com/rollbar/rollbar-gem/blob/8a6d7dac528ee7b3855430b1135c588897e1f5b2/lib/rollbar/plugins/sidekiq/plugin.rb#L41)

krisleech:
I too think it would be preferable to have non-reporting of retried exceptions built in to the sidekiq integration.

marcsasson:
> I think this sort of behavior can already be accomplished with `should_capture`. I'll write up an example and put it in the docs.

@nateberkopec Can you link to where you wrote an example or can provide one?

seanlinsley:
Is there a reason not to use [`death_handlers`](https://www.rubydoc.info/github/mperham/sidekiq/Sidekiq.death_handlers)? It seems to do all the heavy lifting, and since 2018 has covered all job types: https://github.com/mperham/sidekiq/pull/3980

st0012:

PR #790, opened 2017-11-03
Version 3.0

nateberkopec:


nateberkopec:
Linking https://github.com/jruby/jruby/issues/5099, which causes tests to fail (but shouldn't be a problem in production JRuby use).

PR #798, opened 2017-11-21
Convert globalids to their text representation in sidekiq context

louim:
This fixes https://github.com/getsentry/raven-ruby/issues/797

I went for the simplest implementation, let me know what you think.

### DIFF
--- a/lib/raven/integrations/sidekiq.rb
+++ b/lib/raven/integrations/sidekiq.rb
@@ -41,6 +41,8 @@ module Raven
         context.map { |arg| filter_context(arg) }
       when Hash
         Hash[context.map { |key, value| filter_context_hash(key, value) }]
+      when GlobalID
+        context.to_s
       else
         context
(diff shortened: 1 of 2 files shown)


nateberkopec:
This looks fine. However, you'll need to deal with the case where GlobalID is not defined (e.g. someone using Sidekiq without Rails).

louim:
I made some tweaks, let me know if there's a better way of doing this.

lib/raven/integrations/sidekiq.rb
```diff
@@ -14,6 +14,7 @@ def call(_worker, job, queue)
 
   class SidekiqErrorHandler
     ACTIVEJOB_RESERVED_PREFIX = "_aj_".freeze
+    HAS_GLOBALID = const_defined?('GlobalID')
```
nateberkopec: 👍

PR #802, opened 2017-12-12
Suppress git errors and refactor Configuration class

alexandrz:
Sometimes `.git` directory exists, but `git` is not installed and raven gem logs `No such file or directory - git` on loading. This PR suppresses this error.

Also, it moves release detecting code from a `Configuration` class to `Util::Release`

### DIFF
--- a/lib/raven/configuration.rb
+++ b/lib/raven/configuration.rb
@@ -198,7 +198,7 @@ module Raven
       self.project_root = detect_project_root
       self.rails_activesupport_breadcrumbs = false
       self.rails_report_rescued_exceptions = true
-      self.release = Utils::Release.new(logger).detect_release
+      self.release = detect_release
       self.sample_rate = 1.0
       self.sanitize_credit_cards = true
       self.sanitize_fields = []
@@ -394,6 +394,10 @@ module Raven
 
     def running_on_heroku?
       File.directory?("/etc/heroku")
+    end
+
+    def detect_release
+      Utils::Release.new(logger).detect_release
     end
   end
 end
--- a/lib/raven/configuration.rb
+++ b/lib/raven/configuration.rb
@@ -198,7 +198,7 @@ module Raven
       self.project_root = detect_project_root
       self.rails_activesupport_breadcrumbs = false
       self.rails_report_rescued_exceptions = true
-      self.release = Utils::Release.detect_release
+      self.release = Utils::Release.new(logger).detect_release
       self.sample_rate = 1.0
       self.sanitize_credit_cards = true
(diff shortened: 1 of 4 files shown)


lib/raven/configuration.rb
```diff
@@ -375,7 +375,10 @@ def detect_release_from_capistrano
     end
 
     def detect_release_from_git
-      `git rev-parse --short HEAD`.strip if File.directory?(".git") rescue nil
+      release = `git rev-parse --short HEAD 2>/dev/null`.strip if File.directory?('.git')
+      return nil if release == ''
```
olleolleolle: You also need to return `release` as the last statement.

lib/raven/configuration.rb
```diff
@@ -375,7 +375,10 @@ def detect_release_from_capistrano
...
+    rescue
```
olleolleolle: New RuboCop rules would say "Ah, you ought to catch specific errors, such as `StandardError`, in this case".

olleolleolle: Oh, and `C: Class has too many lines.` just happened.

lib/raven/configuration.rb
olleolleolle: 👍

olleolleolle:
In order to pass RuboCop's ClassLength rule, you may need to find a way to extract classes out of Configuration.

It _seems_ that there's a theme of inspecting the filesystem and environment for information in these methods:

- detect_release_from_heroku
- detect_release_from_capistrano
- detect_release_from_git

lib/raven/configuration.rb
```diff
@@ -198,7 +198,7 @@ def initialize
       self.project_root = detect_project_root
       self.rails_activesupport_breadcrumbs = false
       self.rails_report_rescued_exceptions = true
-      self.release = detect_release
+      self.release = Utils::Release.new(logger).detect_release
```
olleolleolle: Perhaps the 30 AbcSize does not allow getting rid of the `detect_release` method.

olleolleolle:
@nateberkopec Oh, hi!

Could you review this change?

olleolleolle:
@alexandrz Maybe you can improve the Descriptoin and the title of this PR, now that it does a little more (to the code).

alexandrz:
@olleolleolle Updated it. Thank you for reviewing the changes.

nateberkopec:
I went with the simplest thing that could possibly work and used `Raven.sys_command`. I like the idea of a Release object though, I'll add it to the 3.0 branch.

PR #811, opened 2018-01-24
last_event_id doesn't need to be a thread local

### LINKED ISSUE #805: Raven::Instance#capture_type memory leak
Hey, I believe there's a memory leak in Raven on this [line here](https://github.com/getsentry/raven-ruby/blob/master/lib/raven/instance.rb#L128). As I understand, nothing will clean up these thread variables. Let me know if I misunderstood anything. 

Consider this example & reproduction below.

```
2.3.6 :001 > Thread.current.keys.select {|x| x.to_s.include?("sentry")}
 => []
2.3.6 :002 > Raven::CLI.test(dsn)
 => true
2.3.6 :003 > Thread.current.keys.select {|x| x.to_s.include?("sentry")}
 => [:sentry_context, :sentry_breadcrumbs, :sentry_70244520082880_last_event_id]
2.3.6 :004 > Raven::CLI.test(dsn)
 => true
2.3.6 :005 > Thread.current.keys.select {|x| x.to_s.include?("sentry")}
 => [:sentry_context, :sentry_breadcrumbs, :sentry_70244520082880_last_event_id, :sentry_70244513342000_last_event_id]
```

nateberkopec:
Each Instance has a last_event_id, and the "global" last_event_id will
simply be the "global" instance's (Raven.instance).

Fix #805

### DIFF
--- a/lib/raven/instance.rb
+++ b/lib/raven/instance.rb
@@ -19,11 +19,8 @@ module Raven
   #     end
   #   end
   class Instance
-    # See Raven::Client.
     attr_writer :client
-
-    # See Raven::Configuration.
-    attr_accessor :configuration
+    attr_accessor :configuration, :last_event_id
 
     def initialize(context = nil, config = nil)
       @context = @explicit_context = context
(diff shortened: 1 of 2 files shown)


nateberkopec:
Would break multithreaded webservers or Sidekiq...hmph.

Maybe each thread could have it's own Instance? 

I really don't want to resort to finalizers.

nateberkopec:
> Maybe each thread could have it's own Instance?

This wouldn't work because people can/should create their own instances, which may then run in multithread environments such as Sidekiq.

PR #815, opened 2018-02-26
Prevent duplicated events when AJ adapter has a specific integration.

BrentWheeldon:
In rails 5, the job class’s `queue_adapter` property is not a class (such as `ActiveJob::QueueAdapters::SidekiqAdapter`) like it was in rails 4, but an instance (of a class such as `ActiveJob::QueueAdapters::SidekiqAdapter`).

The specs were previously not testing the logic in `already_supported_by_specific_integration?` which is why there wasn’t a failing spec.


nateberkopec:
👍 Well done!

PR #817, opened 2018-03-16
Sentry calls rescue_from handler twice if exception occurs in rescue_from

quinn:
I've added a failing test case to show what I'm talking about. Basically, because the sentry gem relies on `rescue_with_handler` there is an edge case where a legitimate error that Sentry should log, but instead the error gets rescued and sent back to the rescue_from handler.

### DIFF
--- a/lib/raven/integrations/rails/active_job.rb
+++ b/lib/raven/integrations/rails/active_job.rb
@@ -17,7 +17,10 @@ module Raven
       def capture_and_reraise_with_sentry(job, block)
         block.call
       rescue Exception => exception # rubocop:disable Lint/RescueException
-        return if rescue_with_handler(exception)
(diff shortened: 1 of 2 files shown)


quinn:
I Updated the integration to fix the double call error, but Sentry is unable to catch the second exception that was not caught by `rescue_from`. Ideally this would be logged, but I think this functionality is closer to what would be expected in this scenario.

quinn:
I think it is not ideal, but also acceptable that Sentry would be unable to catch an exception within a rescue_from handler. If you are using rescue_from, you are sort-of saying "I am going to handle this error" rather than letting the framework handle it (i.e. Sentry).

nateberkopec:
Can you fix the style errors identified with Rubocop, please?

PR #822, opened 2018-04-11
Add option to disable logging of IP addresses

mrhead:
We are preparing our APP for [GDPR][1] and we do not want to send any personalized information to Sentry, but raven-ruby sends user's IP address to Sentry by default. This commit introduces an option to disable this functionality.

[1]: https://en.wikipedia.org/wiki/General_Data_Protection_Regulation

### DIFF
--- a/lib/raven/configuration.rb
+++ b/lib/raven/configuration.rb
@@ -46,6 +46,9 @@ module Raven
     # the disk. See Raven::LineCache for the required interface you must implement.
     attr_accessor :linecache
 
+    # Log user's IP address? Default: true.
+    attr_accessor :log_ip_address
+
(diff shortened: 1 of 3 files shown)


jeremy:
Could implement as a Processor, too! https://github.com/getsentry/raven-ruby/blob/a3972d94a7b6667d3191d4c160e03fd369d6b6d7/docs/processors.rst

mrhead:
@jeremy Thank you for pointing me to the good direction! I've implemented this as a processor for our app. Also, I'm closing this PR as implementation in the PR isn't good enough.

If anyone is interested in my processor, it's here: https://gist.github.com/mrhead/643dc047a7733b4db801da02953e59b3

PR #824, opened 2018-04-12
Fixes #823 context attributes with nil raises error

### LINKED ISSUE #823: Context attributes with nil raise error
If you send a message with any of the attributes `user`, `extra` or `tags` being `nil` an error is thrown because you try to merge ìt later here

https://github.com/getsentry/raven-ruby/blob/master/lib/raven/event.rb#L212

I think you should skip any attributes which are `nil`.

[Here](https://github.com/getsentry/raven-ruby/blob/master/lib/raven/event.rb#L42) you assign all context attributes no matter what value they have.

joker-777:
https://github.com/getsentry/raven-ruby/issues/823

* skipps context attributes with value nil

### DIFF
--- a/lib/raven/event.rb
+++ b/lib/raven/event.rb
@@ -39,7 +39,7 @@ module Raven
 
       # Allow attributes to be set on the event at initialization
       yield self if block_given?
-      init.each_pair { |key, val| public_send("#{key}=", val) }
+      init.each_pair { |key, val| public_send("#{key}=", val) unless val.nil? }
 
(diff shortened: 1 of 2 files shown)


nateberkopec:
Needs a test but otherwise 👍

gingerlime:
@joker-777 can you add a test?

joker-777:
Yes, I will add a test.Is on my todo-list.

joker-777:
@nateberkopec Added the requested test.

PR #825, opened 2018-04-18
Use a circular buffer for breadcrumbs

xfix:
This avoids accidentally quadratic behaviour for adding breadcrumbs. Ruby arrays [aren't double ended queues](https://stackoverflow.com/questions/7310015/ruby-array-internals).

### DIFF
--- a/lib/raven/breadcrumbs.rb
+++ b/lib/raven/breadcrumbs.rb
@@ -40,6 +40,7 @@ module Raven
 
     def initialize(size = 100)
       @buffer = Array.new(size)
+      @last = 0
     end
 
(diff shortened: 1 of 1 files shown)


xfix:
Oh wait, I noticed that there is an accessor for a buffer, which pretty much means this is here to stay :(.

xfix:
I may look into having accidental O(N^2) complexity only when `buffer` is ever accessed if that is an issue.

nateberkopec:
Not sure this is an issue at the default buffer size. Feel free to reopen with a benchmark to change my mind.

PR #830, opened 2018-04-26
adding stdout transport

boris-fiverr:
Our developers see a value in using Sentry platform for application errors processing.
But we can't use HTTP/S transport to propagate application errors.
Our system runs on Kubernetes platform using Docker containers.
All components obey to `12factor` rules.
All logs are written to `stdout` stream, picked up by `fluentd` plugins and sent to the corresponding receivers.

### DIFF
--- a/spec/raven/transports/stdout_spec.rb
+++ b/spec/raven/transports/stdout_spec.rb
@@ -1,0 +1,11 @@
+require 'spec_helper'
+
+RSpec.describe Raven::Transports::Stdout do
+  let(:config) { Raven::Configuration.new.tap { |c| c.dsn = 'stdout://12345:67890@sentry.localdomain/sentry/42' } }
+  let(:client) { Raven::Client.new(config) }
+
+  it 'should write to stdout' do
+    event = JSON.generate(Raven::Event.from_message("stdout test").to_hash)
+    expect { client.send(:transport).send_event("test", event) }.not_to raise_error
+  end
+end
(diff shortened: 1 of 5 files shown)


nateberkopec:
Mind describing your use case?

boris-fiverr:
Our developers see a value in using Sentry platform for application errors processing.
But we can't use HTTP/S transport to propagate application errors.
Our system runs on Kubernetes platform using Docker containers.
All components obey to 12factor rules.
All logs are written to stdout stream, picked up by fluentd plugins and sent to the corresponding receivers.

erikash:
As an old time user I would love this functionality as well!

spec/raven/transports/stdout_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'spec_helper'
+
+RSpec.describe Raven::Transports::Stdout do
+  let(:config) { Raven::Configuration.new.tap { |c| c.dsn = 'stdout://12345:67890@sentry.localdomain/sentry/42' } }
+  let(:client) { Raven::Client.new(config) }
+
+  it 'should write to stdout' do
+    event = JSON.generate(Raven::Event.from_message("stdout test").to_hash)
+    expect { client.send(:transport).send_event("test", event) }.not_to raise_error
```
nateberkopec: This isn't testing what the behavior is. The test needs to listen to STDOUT for the desired output.

erikash:
@nateberkopec, any update?

PR #832, opened 2018-05-03
Proposal: add Faraday middleware breadcrumb recorder

tilsammans:
The breadcrumb middleware for Faraday from the documentation is quite useful. So useful in fact that it seems more DRY to offer this from the Raven gem, as opposed to writing this for every separate Ruby app.

If you think it's a useful addition, I will add tests and register the middleware with Faraday and add documentation on how to use it.


nateberkopec:
Neat. I like this. Let me think a bit more about how this should be included/enabled in the gem.

st0012:

PR #836, opened 2018-05-18
`Raven.capture(silent: true)` option

slayer:
Added `:silent` option to `Raven.capture` method
It is useful on network requests when further execution should not be interrupted 
```
Raven.capture(silent: true) do
  SomeNetworkRequest.call
end
```

It is a short version of
```
begin
  SomeNetworkRequest.call
rescue => exception
  Raven.capture_exception(exception)
end
```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,7 +7,7 @@ AllCops:
      - 'vendor/**/*'
 
 Metrics/ClassLength:
-  Max: 268
+  Max: 270
   CountComments: false
 
 Metrics/AbcSize:
--- a/lib/raven/instance.rb
+++ b/lib/raven/instance.rb
@@ -87,6 +87,10 @@ module Raven
     #   Raven.capture do
     #     MyApp.run
     #   end
+    #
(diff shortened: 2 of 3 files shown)


nateberkopec:
I don't think I want to add this feature. I don't really see what's wrong with adding your own `rescue` block if you really feel the exception should be swallowed and not re-raised.

slayer:
There are many code in `raven-ruby` that can be moved into the main program.

nateberkopec:
I think the shortest possible way to get the behavior you want would just be:

```ruby
Raven.capture { SomeNetworkRequest.call } rescue nil
```

right?

slayer:
Sure, but it looks like 
```
Raven.capture { SomeNetworkRequestor.new(host: 'example.com').call(params: {a: 1, b: 2} )  } rescue nil
```
It looks a little messy.

Ok, if you think this idea is ugly, you can close the issue

nateberkopec:
I just don't think it's something we need to bake into the library itself. `rescue` is adequate to fulfill your use case here.

PR #843, opened 2018-05-31
Fix infinite backoff under pressure

### LINKED ISSUE #841: Backoff stuck until 36s of idle time
https://github.com/getsentry/raven-ruby/blob/120a416cf0df0a08803992a7c643741862f000ad/lib/raven/client.rb#L28

Because failed_send() is called right after should_try?(), last_check is updated to Time.now, meaning no messages can ever be sent until at least 36s (6 power of 2) has elapsed between two calls to send_event().

Expected behavior: backoff should try to send a message up to once every 36s in the worst case, until a successful send.
Observed behavior: After a single error, no further messages will be sent until there is a gap of 36s between calls to send_event(). On busy systems like I have, this means no more messages will be sent until the process is restarted.

bombela:
Calling failed_send when should_try? returns false was resetting last_check, effectively disabling all logs until enough time passes between two call to send_event (36s here).

Fixes #841

### DIFF
--- a/lib/raven/client.rb
+++ b/lib/raven/client.rb
@@ -25,7 +25,7 @@ module Raven
       event = event.to_hash
 
       unless @state.should_try?
-        failed_send(nil, event)
+        configuration.logger.error "Not sending event due to previous failure(s)."
         return
       end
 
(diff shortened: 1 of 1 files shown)


lib/raven/client.rb
```diff
@@ -25,7 +25,7 @@ def send_event(event)
       event = event.to_hash
 
       unless @state.should_try?
-        failed_send(nil, event)
+        configuration.logger.error "Not sending event due to previous failure(s)."
```
nateberkopec: It would be cool if we could log here how long we have until the next attempt.

nateberkopec: It's also important that the backtrace (essentially all the info on line 96) is still printed here.

For people that don't use i.e. Sidekiq to send their Sentry events (so, automated retry), they're going to lose this event forever.

bombela: Ok will look into it.

bombela:
superseded by #886

PR #845, opened 2018-06-12
This commit suppresses a warning

yui-knk:
```
NOTE: Gem::Specification#has_rdoc= is deprecated with no replacement. It will be removed on or after 2018-12-01.
Gem::Specification#has_rdoc= called from raven-ruby/sentry-raven.gemspec:15.
```


nateberkopec:
Thanks!

PR #848, opened 2018-06-15
Add SemVer badge to README

greysteil:
Adds a SemVer compatibility badge that displays the percentage of CI runs that pass when updating `sentry-raven` between SemVer compatible versions. Data comes from Dependabot (disclosure: I built it).

Here's what the badge looks like:

[SemVer](https://dependabot.com/compatibility-score.html?dependency-name=sentry-raven&package-manager=bundler&version-scheme=semver)

(Any feedback really appreciated - I only just built this.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@
 [![Gem Version](https://img.shields.io/gem/v/sentry-raven.svg)](https://rubygems.org/gems/sentry-raven)
 [![Build Status](https://img.shields.io/travis/getsentry/raven-ruby/master.svg)](https://travis-ci.org/getsentry/raven-ruby)
 [![Gem](https://img.shields.io/gem/dt/sentry-raven.svg)](https://rubygems.org/gems/sentry-raven/)
+[![SemVer](https://api.dependabot.com/badges/compatibility_score?dependency-name=sentry-raven&package-manager=bundler&version-scheme=semver)](https://dependabot.com/compatibility-score.html?dependency-name=sentry-raven&package-manager=bundler&version-scheme=semver)
 
 
 [Documentation](https://docs.getsentry.com/hosted/clients/ruby/) | [Bug Tracker](https://github.com/getsentry/raven-ruby/issues) | [Forum](https://forum.sentry.io/) | IRC: irc.freenode.net, #sentry


olleolleolle:
@greysteil It does not yet render, on my Firefox. (I tried reloading.)

greysteil:
Ooh, that's bad! Will fix asap.

olleolleolle:
@greysteil Update: now visible

greysteil:
👍 - what do you reckon? Any feedback very much appreciated!

olleolleolle:
Oh, the numbers shown after the click are interesting, esp. that they've got a bit of history showing. 

Perhaps there's an even more specific way of saying "this is a number over time, not a snapshot"?

greysteil:
That would be rad, although perhaps hard to do in a badge. Perhaps a bigger embeddable or something like that.

olleolleolle:
@greysteil I was more opining on what the screen _after the click_ looks like.

greysteil:
Ah interesting. Maybe as simple as adding `so far` to the end of the first sentence:

> Dependabot has updated sentry-raven between SemVer compatible versions 320 times across 102 projects **so far**.
93% of those updates passed CI.

What do you reckon about including the badge on this repo?

nateberkopec:
This is super cool. I like it because it reflects what we *do*, rather than what we *say we'll do*. I've always been in the camp of "I'll try my best but sometimes things happen, I'm not a robot". 93% seems like a fair score for that outlook 😆

dentarg:
@greysteil did your service catch the breaking change in [2.7.0](https://github.com/getsentry/raven-ruby/blob/master/changelog.md#270)

See https://github.com/getsentry/raven-ruby/pull/743#issuecomment-342541035

dentarg:
Ah, you are using the info from the projects that are using dependabot, I get it. I thought it was only using CI for this project.

greysteil:
Yep - badge for just that update is [here](https://dependabot.com/compatibility-score.html?dependency-name=sentry-raven&package-manager=bundler&previous-version=2.6.3&new-version=2.7.0) (we include it on PRs).

PR #858, opened 2018-07-31
Separate sidekiq integration into separate files

jasondoc3:
## What
Separates the sidekiq integration into separate files

## Why
I'd like to configure `Raven` without `sidekiq`, but use a custom error handler that inherits from `Raven::SidekiqErrorHandler`. Currently  the `Raven::SidekiqErrorHandler` class cannot be required separately from the `Sidekiq.configure_server` call.


lib/raven/integrations/sidekiq/sidekiq_error_handler.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    # this will change in the future:
+    # https://github.com/mperham/sidekiq/pull/3161
+    def transaction_from_context(context)
+      classname = (context["wrapped"] || context["class"] ||
+                    (context[:job] && (context[:job]["wrapped"] || context[:job]["class"]))
+                  )
+      if classname
+        "Sidekiq/#{classname}"
+      elsif context[:event]
```
martin-jaime: What about assigning `context[:event]` to a variable so that later we don't access it twice (line 50 and 51)?

st0012:
Closing this due to inactivity. Feel free to reopen it if you feel like continuing it though 😄

PR #860, opened 2018-08-21
Enable configurable rack environment recorded parameters. Fix #854.

### LINKED ISSUE #854: Configurable Rack environment formatting
It would be really useful if the `ENVIRONMENT` desired fields for rack env could be set through a config or an env variable. ( The allowed environment fields for rack are currently hardcoded [here](https://github.com/getsentry/raven-ruby/blob/master/lib/raven/integrations/rack.rb#L131) )

If an empty formatter is provided, all fields in the environment hash are to be recorded.

0xdiba:
This commit adds a configuration option allowing rack env parameters
that are sent to sentry in an event to be configured by the user.

The default parameters remain the as they were so no breaking changes
are introduced.

### DIFF
--- a/lib/raven/configuration.rb
+++ b/lib/raven/configuration.rb
@@ -193,6 +193,9 @@ module Raven
     # the dsn value, whether it's set via `config.dsn=` or `ENV["SENTRY_DSN"]`
     attr_reader :dsn
 
+    # Array of rack env parameters to be included in the event sent to sentry.
+    attr_accessor :rack_env_whitelist
+
     # Most of these errors generate 4XX responses. In general, Sentry clients
     # only automatically report 5xx responses.
     IGNORE_DEFAULT = [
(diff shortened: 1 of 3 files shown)


0xdiba:
Hey @nateberkopec sorry for spamming. Is this feat something you would like added ?
If it is, should I remove the docs from the current PR and open one in the sentry-docs repo?

We would really like to have this ability on one of the future raven releases.

st0012:
@0xdiba there seems to be a rubocop error, can you fix it?

0xdiba:
Fixed and rebased on latest master.
Everything should be ok!

st0012:
@0xdiba awesome, thanks 👍 
this will be included in `v3.1.1` and be ported to master once the new 4.0 version comes stable.

st0012 (approved):

PR #861, opened 2018-08-26
Note that should_capture does not replace excluded_exceptions

iloveitaly:
I run into issues with other libraries where using two exception filters disables another filter which you have enabled. Noting this in the docs would be helpful.


st0012:
@iloveitaly thanks for the PR! but I'm closing this because the file doesn't exist anymore.

PR #869, opened 2018-10-02
Sanitize credit card with whitespace

garno:
Sometimes, users will add additional spaces to their credit card number. If it is not sanitized properly on the frontend, it might find its way into the Sentry event which is a breach. To put all the chances on our side, we should ignore those spaces.

All of those should be filtered out now :

* `4560 1234 1234 1234`
* ` 4560123412341234`
* `4560123412341234 `


st0012:
Thanks for the fix! I think it looks good. But the benchmark result shows your change could increase memory allocation by 50~80%:

**Ruby 2.6 & Rails 6.0**

```
Calculating -------------------------------------
              branch   811.648k memsize (    17.514k retained)
                         5.874k objects (    12.000  retained)
                        50.000  strings (     4.000  retained)

Comparison:
              master:     444073 allocated
              branch:     811648 allocated - 1.83x more
```

**Ruby 2.7 & Rails 6.0**

```
Calculating -------------------------------------
              branch   669.615k memsize (    17.265k retained)
                         5.827k objects (    12.000  retained)
                        50.000  strings (     4.000  retained)

Comparison:
              master:     429823 allocated
              branch:     669615 allocated - 1.56x more
```

Unfortunately, we can't accept this amount of increase in memory allocation just for data sanitizing. So unless you can find a way to reduce the allocation impact, I'll need to close this PR.

--- 
I suspect it's because you use `gsub` instead of `gsub!`. And I guess this could reduce the memory allocation:

```ruby
    def strip_whitespace(value)
       value.frozen? ? value.gsub(/\s/, '') : value.gsub!(/\s/, '')
     end
```

hongtron:
👋 I put together a little benchmark that seems to confirm your suspicion about `gsub!` having a smaller memory footprint than `gsub` (also tested `tr` and `delete`, and their `!` variants): https://gist.github.com/hongtron/6b09f2dac28bc66618c358e006e0a1df

```
       user     system      total        real
tr  0.073900   0.000000   0.073900 (  0.073927)
tr!  0.057925   0.000000   0.057925 (  0.057929)
gsub  0.032254   0.000000   0.032254 (  0.032264)
gsub!  0.021508   0.000000   0.021508 (  0.021510)
delete  0.062869   0.000000   0.062869 (  0.062873)
delete!  0.055036   0.000000   0.055036 (  0.055040)


Calculating -------------------------------------
                  tr    17.008M memsize (     0.000  retained)
                       405.000k objects (     0.000  retained)
                        50.000  strings (     0.000  retained)
                 tr!     8.928M memsize (     0.000  retained)
                       203.000k objects (     0.000  retained)
                         2.000  strings (     0.000  retained)
                gsub    12.968M memsize (     0.000  retained)
                       304.000k objects (     0.000  retained)
                        50.000  strings (     0.000  retained)
               gsub!     4.888M memsize (     0.000  retained)
                       102.000k objects (     0.000  retained)
                         1.000  strings (     0.000  retained)
              delete    12.968M memsize (     0.000  retained)
                       304.000k objects (     0.000  retained)
                        50.000  strings (     0.000  retained)
             delete!     4.888M memsize (     0.000  retained)
                       102.000k objects (     0.000  retained)
                         1.000  strings (     0.000  retained)
```

st0012:
Because we're going to drop the data sanitization functionality from SDK after v4.0, I won't make any changes on the processors anymore. So I'm closing this now. Still thanks for the work though 😄

PR #875, opened 2018-11-08
Ensure Rails transaction name removed after action completes

bjeanes:


st0012:
After some investigation, I think there's a reason that we don't put an `ensure` block there. Here's how it works:

- If the controller action (`block.call`) works fine, we don't keep the transaction name. So we `pop` it from the `transaction` array.
- If the controller action triggers an `exception`, we want to keep the transaction in that array. And we'll later use that to send a event to Sentry. Which means we can't `pop` it in this case.

Therefore, I'm closing this now 🙂

bjeanes:
OK. This PR came out of a discussion (with support, or in chat somewhere I don't remember) hence the lack of PR description. The behaviour I was observing is that the controller name was present in subsequent requests. In any case, this was 2 years ago, I am no longer on the project, and so I have no personal investment in seeing this change happen. I only leave this comment to make sure the intended behaviour is examined carefully. I opened this PR due to unintended behaviour, but it may have since been fixed elsewhere.

PR #879, opened 2018-11-29
Fix crash when the fork fails

natemueller:
If the fork itself fails (not the process you exec later)
$CHILD_STATUS is not set and this throws an exceptions.  We saw this
when trying to send log messages to sentry while a server was out of
memory.


st0012:
Thanks for the PR! But the issue has been fixed in https://github.com/getsentry/raven-ruby/pull/895. So I'm closing this.

PR #887, opened 2019-02-19
Prepend the transaction around_action so libraries with controllers can alter the value

drcapulet:
For gems that define controllers, the following code:

```ruby
initializer 'raven.action_controller' do
  ActiveSupport.on_load :action_controller do
    ...
    include Raven::Rails::ControllerTransaction
    ...
  end
end
```

Runs after the controller (and any related `*_actions`) are defined and prevents that controller from being able to change the transaction. By prepending the `around_action`, the transaction is correctly set before any code in that controller runs and allows that controller to change the transaction.


HazAT:
Thx!

PR #895, opened 2019-04-11
Fix crash when Process.wait is used

asBrettisay:
Make `sys_command` tolerate a missing $CHILD_STATUS.

Hi 👋 in my code I was using `Process.wait` and found that it was causing this error in sentry-raven: `undefined method `exitstatus' for nil:NilClass`

My understanding is that calling `Process.wait` clears out the ruby global for last child process, and this is causing the error. I think that sentry-raven should be able to work if I use `Process.wait` or something similar like `trap('CLD', 'IGNORE') so I made a PR that I think will fix the issue. Hope that this helps 👍.

### DIFF
--- a/lib/raven/base.rb
+++ b/lib/raven/base.rb
@@ -100,7 +100,7 @@ module Raven
 
     def sys_command(command)
       result = `#{command} 2>&1` rescue nil
-      return if result.nil? || result.empty? || $CHILD_STATUS.exitstatus != 0
+      return if result.nil? || result.empty? || ($CHILD_STATUS && $CHILD_STATUS.exitstatus != 0)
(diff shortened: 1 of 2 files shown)


dennis-ef:
Can this please be implemented. It's also fixing crashes for me. It was difficult to figure out what was the problem until I found this PR. Thank you.

asBrettisay:
@HazAT can we see if this will work?

HazAT:
Oh, @asBrettisay Can you please fix the tests before merging? 
Thx

asBrettisay:
@HazAT Will do

asBrettisay:
@HazAT Tests are passing

PR #898, opened 2019-04-17
Remove ActiveJob keys for both Sidekiq and DelayedJob

alissonbrunosa:
It's as same as https://github.com/getsentry/raven-ruby/pull/386 but for both Sidekiq and DelayedJob
> Once an ActiveJob is queued, ActiveRecord references get serialized into
some internal reserved keys, such as _aj_globalid.
The problem is, if this job in turn gets queued back into ActiveJob with
these magic reserved keys, ActiveJob will throw up and error. We want to
capture these and mutate the keys so we can sanely report it.

The difference is instead of rename the key I removed it, because this key will not be used.

### DIFF
--- a/lib/raven/integrations/delayed_job.rb
+++ b/lib/raven/integrations/delayed_job.rb
@@ -1,4 +1,5 @@
 require 'delayed_job'
+require 'raven/utils/context_filter'
 
 module Delayed
   module Plugins
@@ -29,7 +30,7 @@ module Delayed
             extra[:handler] = job.handler[0...1000] if job.handler
 
             if job.respond_to?('payload_object') && job.payload_object.respond_to?('job_data')
-              extra[:active_job] = job.payload_object.job_data
+              extra[:active_job] = Utils::ContextFilter.filter_context(job.payload_object.job_data)
             end
             ::Raven.capture_exception(e,
                                       :logger => 'delayed_job',
--- a/lib/raven/integrations/sidekiq/error_handler.rb
+++ b/lib/raven/integrations/sidekiq/error_handler.rb
@@ -1,4 +1,4 @@
-require 'raven/integrations/sidekiq/context_filter'
+require 'raven/utils/context_filter'
 
 module Raven
   module Sidekiq
@@ -6,7 +6,7 @@ module Raven
       SIDEKIQ_NAME = "Sidekiq".freeze
 
       def call(ex, context)
-        context = ContextFilter.filter_context(context)
+        context = Utils::ContextFilter.filter_context(context)
         Raven.context.transaction.push transaction_from_context(context)
         Raven.capture_exception(
           ex,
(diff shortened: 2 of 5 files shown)


cantino:
Looks like a minor Rubocop issue on this.

lib/raven/integrations/delayed_job.rb
```diff
@@ -1,8 +1,12 @@
 require 'delayed_job'
 
+require 'raven/context_filter'
+
 module Delayed
   module Plugins
     class Raven < ::Delayed::Plugin
+      include Raven::ContextFilter
```
st0012, line 8: If you want to call `Raven::ContextFilter.filter` directly, this include is unnecessary

st0012 (changes requested):
Thanks for the work! It looks good in general, just added some small suggestions. Can you rebase it against master as well?

lib/raven/context_filter.rb
```diff
@@ -0,0 +1,40 @@
...
+          Hash[context.map { |key, value| filter_context_hash(key, value) }]
+        else
+          format_globalid(context)
+        end
+      end
+
+      private
+
+      def filter_context_hash(key, value)
+        (key = key[3..-1]) if key[0..3] == ACTIVEJOB_RESERVED_PREFIX
```
alissonbrunosa, line 27: This will remove only the first 3 chars from the key. Like so `_aj_symbol_keys` => `_symbol_keys`. Should I remove this `_` in the beginning as well?

st0012: can we do it with substitution instead? like

```ruby
ACTIVEJOB_RESERVED_PREFIX_REGEX = /^_aj_/.freeze

key = key.sub(ACTIVEJOB_RESERVED_PREFIX_REGEX, "") if key.match?(ACTIVEJOB_RESERVED_PREFIX_REGEX)

File: lib/raven/context_filter.rb [hunk 3]
```diff
@@ -0,0 +1,40 @@
+module Raven
+  module ContextFilter
```
st0012, line 2: think we can move it under `lib/utils/context_filter` and name it `Raven::Utils::ContextFilter`

st0012 (changes requested):
overall looks good 👍  could be merged into 3.1, just need some updates

st0012:
@alissonbrunosa sorry that I forgot to change the base of this PR 🤦‍♂️ because we have started the restructuring work of version 4.0, all the previously-opened contributions should be merged into `v3-1-stable`.

so I wonder if you can revert your merge of the master and merge/rebase `v3-1-stable` instead? I've tried squash merging this PR locally, but that'll erase you from the merge commit (we do squash commit). very sorry for this 🙇

alissonbrunosa:
@st0012 don't worry :sweat_smile: 
I tried to revert the merge commit, but it made thing way worse.
I created a branch from `v3-1-stable`, and add the changes.

st0012 (approved):

st0012:
@alissonbrunosa thanks for the work and sorry again for the hassles 😅

PR #909, opened 2019-07-06
Skip ActiveJob integration if there is a better one

fsateler:
When `already_supported_by_specific_integration?` returns true, the context and breadcrumbs are cleared by `capture_and_reraise_with_sentry`. This is wrong, because then the specific integration will report the exception with an empty context.

Bypass the problem by not doing the `capture_and_reraise_with_sentry` if we have a specific integration


codez:
I am very interested in this pull request because we also see that no custom context is sent for delayed job exceptions. What is missing for the pull request to be merged? Is there something we can help?

PR #914, opened 2019-09-03
Logger can take in exceptions so stringify first

halkeye:
According to https://ruby-doc.org/stdlib-2.5.0/libdoc/logger/rdoc/Logger.html\#method-i-add-label-Args

### DIFF
--- a/lib/raven/breadcrumbs/logger.rb
+++ b/lib/raven/breadcrumbs/logger.rb
@@ -36,7 +36,7 @@ module Raven
 
       # some loggers will add leading/trailing space as they (incorrectly, mind you)
       # think of logging as a shortcut to std{out,err}
-      message = message.strip
+      message = message.to_s.strip
 
       last_crumb = Raven.breadcrumbs.peek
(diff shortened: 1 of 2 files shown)


HazAT:
I think the failure doesn't have anything to do with the change.

BTW the link you posted is broken.

halkeye:
I agree master has the same errors.

I think my link has a extra slash in it

https://ruby-doc.org/stdlib-2.5.0/libdoc/logger/rdoc/Logger.html#method-i-add-label-Args

halkeye:
Yea that works

HazAT:
Thx!

halkeye:
Awesome. Now to wait for a release.

Thanks so much

HazAT:
Just to make sure, is this blocking anything?
Since it's such a minor change I wanted to wait until we have more to release.

halkeye:
It's blocking us from using the logger breadcrumb feature. Since we have some.legacy code just logging the exception directly

HazAT:
@halkeye Releaseing it now :)

PR #920, opened 2019-09-21
feat: Allow block on tags_context and extra_context

rivayama:
This makes `tags_context` and `extra_context` block receivable and it defines a scope for the tags/extra.

Before:

```ruby
posts.each do |post|
  Raven.tags_context(post_id: post.id)
  # do something
end
Raven.context.tags.delete(:post_id)
```

After:

```ruby
posts.each do |post|
  Raven.tags_context(post_id: post.id) do
    # do something
  end
end
```

### DIFF
--- a/lib/raven/instance.rb
+++ b/lib/raven/instance.rb
@@ -183,6 +183,9 @@ module Raven
     #   Raven.tags_context('my_custom_tag' => 'tag_value')
     def tags_context(options = nil)
       context.tags.merge!(options || {})
+      yield if block_given?
+    ensure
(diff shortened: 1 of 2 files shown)


HazAT:
THX!

tycooon:
Hi, I added a comment here: https://github.com/getsentry/raven-ruby/commit/0b257441ec37ffbf8a39a7c97a496570cd9d28f7. Please take a look.

philipbjorge:
@rivayama -- Is there a specific reason that we didn't add this capability for user context?
Just wondering or if this is a good opportunity for me to make a PR :wink:

PR #939, opened 2019-12-10
fix #938

### LINKED ISSUE #938: rake raven:test does not work if environments are specified in config
Example:

```
Raven.configure do |config|
  config.dsn = ...
  config.environments = %w(production staging)
end
```

```
rake raven:test 
divided by 0 excluded from capture: Not configured to send/capture in environment 'staging'
```

This seems to be due to ```.pop``` here:
https://github.com/getsentry/raven-ruby/blob/master/lib/raven/cli.rb#L21
It mutates config.environments so it no longer includes the environment we are sending for

glebtv:


lib/raven/cli.rb
```diff
@@ -18,7 +18,7 @@ def self.test(dsn = nil, silent = false, config = nil) # rubocop:disable all
 
       # wipe out env settings to ensure we send the event
       unless config.capture_allowed?
-        env_name = config.environments.pop || 'production'
+        env_name = config.environments.first || 'production'
```
HazAT: Shouldn't this be `last` instead of `first` to maintain the same behavior?

st0012:
closing because the issue has been fixed in https://github.com/getsentry/raven-ruby/pull/989

PR #940, opened 2019-12-12
Fix bugs with send_event in asynchronous mode

cstyles:
When running Raven asynchronously, we call `to_json_compatible` on `event` which results in all of the keys of the `event` hash being Strings, not Symbols. So `event[:event_id]` will return `nil` and the log won't have the `event_id`. Similarly, `send_event` can (indirectly) call ` get_log_message` which also assumes the keys to the hash are Symbols. This PR adds checks for when the keys are Strings.

### DIFF
--- a/lib/raven/client.rb
+++ b/lib/raven/client.rb
@@ -35,7 +35,8 @@ module Raven
         return
       end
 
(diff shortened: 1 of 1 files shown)


cstyles:
I think this fixes https://github.com/getsentry/raven-ruby/issues/800

Issue #943, opened 2020-01-10
Please provide methods to not depend on shell exec in raven-ruby gem


jsjohnst:
Mature organizations usually run security management software (such as a security siem) which alerts to issues with their code. One of the common features of these types of packages is letting you know about insecure operations the code running performs. Generally speaking, having production Ruby code invoke shell commands is considered to be an anti-pattern in security.

There are multiple places in the `raven-ruby` gem which invoke system commands:

1. To get the current Git release version as performed in [configuration.rb](https://github.com/getsentry/raven-ruby/blob/master/lib/raven/configuration.rb#L412). Thankfully in the case where an organization does not want to perform this shell exec, you provide an override via the environment as performed [here](https://github.com/getsentry/raven-ruby/blob/master/lib/raven/configuration.rb#L349-L356).
2. In [context.rb](https://github.com/getsentry/raven-ruby/blob/master/lib/raven/context.rb#L38), it is used to pull in the ruby version via `ruby -v`. As the exact same information is available via `RUBY_DESCRIPTION` in all supported versions of Ruby (at least since Ruby 1.9.3 anyway, maybe earlier, which is long since end of life), can you please replace the invocation of the `sys_command` with a check to see if that exists, then fall back on the sys_command in the case it doesn't exist (if you prefer, honestly I don't see the reason to not just depend on `RUBY_DESCRIPTION`)?
3. Also in [context.rb](https://github.com/getsentry/raven-ruby/blob/master/lib/raven/context.rb#L28-L31), it is used to pull in information about the current operating os via repeated calls to `uname` with different flags being passed in. Can you please provide a way to pass this in via the environment as done in example 1 (with regards to the current release version), falling back on the shell exec for backwards compatibility and/or for those who want to enable auto-detection.

I'd be happy to provide a pull request if you prefer, but these changes should hopefully be fairly straight forward and I am surprised this hasn't come up before with other customers.

st0012:
I have added https://github.com/getsentry/raven-ruby/pull/1017 to address point 2 and 3

jsjohnst:
Wow... nine months to write five lines of code to address a security issue. Good to know how serious you take customer reported security issues!

PR #945, opened 2020-01-13
Replace unnecessary shell exec to get Ruby version info

jsjohnst:
In [issue #943 sub-point #2](https://github.com/getsentry/raven-ruby/issues/943), it was identified that using RUBY_DESCRIPTION returns the same thing as `ruby -v` does for all currently supported Ruby versions. This change fixes that sub-point and provides a fallback to the old method in the event it fails as a safety precaution.


jsjohnst:
None of the build failures were related to my change. :|

st0012:
Thanks for the fix! Can you rebase it against the latest master? I have fixed the build.

PR #951, opened 2020-01-29
Circular references processor should not change user data

estepnv:
I understand that it's not memory-efficient, but you are working with business rules which might rely on references.
Consider opting out this behavior, or display a warning somewhere, and this behavior should be documented. 
I spent 6 hours of debugging code because of the sentry event. Imagine my feelings when I figure out it's caused by raven.

### DIFF
--- a/lib/raven/processor/removecircularreferences.rb
+++ b/lib/raven/processor/removecircularreferences.rb
@@ -6,9 +6,9 @@ module Raven
 
       case value
       when Hash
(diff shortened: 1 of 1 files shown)


st0012:
@estepnv thanks for the fix! 
but this approach could increase memory allocation by 16% or more, which is something we should avoid if possible (you can see the detail about this analysis at https://github.com/getsentry/raven-ruby/pull/994's description). so I will close this once #994 is proved working and merged.

PR #952, opened 2020-01-30
Drop Failing Spec for Circular References in Hashes

ksylvest:
Fix for https://github.com/getsentry/raven-ruby/issues/947.

The `JSON` 2.3.0 gem changed how parsing errors are raised. For example:

    require 'json'
    data = {}
    data['cycle'] = data
    JSON.dump(data)

With json v2.2.0:
This raises a `SystemStackError`.

With json v2.3.0:
This causes `fatal (machine stack overflow in critical region)`


ksylvest:
@HazAT wanted to followup from the change introduced here: https://github.com/getsentry/raven-ruby/pull/946

I see a few different options:

1. pin the version of `json` to 2.2.0 (probably not a great idea since it is a system library)
2. drop the failing test here (https://github.com/getsentry/raven-ruby/blob/master/spec/raven/json_spec.rb#L64-L77)
3. change the test to handle ruby version changes
4. try and get this fixed in ruby/ruby

Do you have any thoughts?

ksylvest:
@HazAT do you know if anyone on the sentry team is available to help with issues / PRs?

HazAT:
Hey, sorry about that.
Currently, we have no Ruby expert on the team, it's mostly me trying to keep up with everything that's going on.
So if you are interested in helping out, we are always looking for contractors, if you are interested, shoot me a mail my first name @ sentry.io :)

WRT to this PR, I would say #2 is fine.

ksylvest:
Adjusted this spec to use no. 2.

PR #959, opened 2020-05-12
CI: Use Ruby 2.4.10, 2.5.8, 2.6.6, and 2.7.1

hosamaly:
As per the README:
> We test on Ruby 2.3, 2.4, 2.5, 2.6 and 2.7 at the latest patchlevel/teeny version.


st0012:
@hosamaly I just removed the patch part from version specifications in https://github.com/getsentry/raven-ruby/pull/970, so we won't need to bump Ruby versions that often now. still thanks for the PR 😄

PR #960, opened 2020-05-18
Serialize ActiveJob arguments to report global ids

hosamaly:
Previously, a job initialized with an ActiveRecord object was reported using its string representation, which often looked like `"#<MyClass:0x00007fa3b7cc37f8>"`.

This change allows us to report ActiveJob arguments in a more useful way. Although the reported arguments make a private implementation detail visible (i.e. ActiveJob's serialization keys), it does so to be future-proof by relying on the same serialization logic as ActiveJob.

### DIFF
--- a/lib/raven/integrations/rails/active_job.rb
+++ b/lib/raven/integrations/rails/active_job.rb
@@ -40,7 +40,7 @@ module Raven
       def raven_context(job)
         ctx = {
           :active_job => job.class.name,
-          :arguments => job.arguments,
+          :arguments => ActiveJob::Arguments.serialize(job.arguments),
           :scheduled_at => job.scheduled_at,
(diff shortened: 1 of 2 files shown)


hosamaly:
I based this commit on the tag v2.13.0 because I hope to see it in an upcoming release of v2.x. Unfortunately, I had to create the pull request against `master` because there is no branch for v2.x yet.

hosamaly:
I discovered a problem with this pull request. If an [async job](https://docs.sentry.io/clients/ruby/config/#optional-settings) is used to send the event to Sentry, then enqueuing the job will fail because ActiveJob doesn't accept arguments with the same key as its private identification keys.

I'm closing this for now until I can think of a better design.

PR #964, opened 2020-07-01
Add an option for delayed_job to report errors to Sentry until after …

juankiz:
…every retry has failed

I couldn't find contribution guidelines so please let me know if something needs to be added or changed.

### DIFF
--- a/lib/raven/configuration.rb
+++ b/lib/raven/configuration.rb
@@ -169,6 +169,10 @@ module Raven
     # }
     attr_reader :before_send
 
+    # Set this option to true if you want Sentry to only capture the last job
+    # retry if it fails.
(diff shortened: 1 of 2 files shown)


lib/raven/configuration.rb
```diff
@@ -169,6 +169,11 @@ class Configuration
     # }
     attr_reader :before_send
 
+    # Job queues are retried when an exception is raised.
+    # Setting this option to true will wait untill all retries have passed before
+    # reporting the exception to Sentry
+    attr_accessor :report_after_job_retries
```
enortham, line 175: I think the following is clearer: Set this option to true if you want Sentry to only capture the last job retry if it fails.

st0012:

PR #993, opened 2020-08-21
Fix broken links in doc

tricknotes:
These links show "Page Not Found".
It seems https://docs.getsentry.com/hosted/clients was moved to https://docs.sentry.io/platforms.


st0012:
@tricknotes awesome! thanks for the fix 😄

PR #1004, opened 2020-08-28
[v3 Branch Pick] Exclude all 4xx Rails errors

st0012:
This PR picks the changes of https://github.com/getsentry/raven-ruby/pull/769

### DIFF
--- a/lib/raven/configuration.rb
+++ b/lib/raven/configuration.rb
@@ -172,16 +172,29 @@ module Raven
     # Errors object - an Array that contains error messages. See #
     attr_reader :errors
 
+    # Most of these errors generate 4XX responses. In general, Sentry clients
+    # only automatically report 5xx responses.
     IGNORE_DEFAULT = [
       'AbstractController::ActionNotFound',
+      'ActionController::BadRequest',
       'ActionController::InvalidAuthenticityToken',
+      'ActionController::InvalidCrossOriginRequest',
+      'ActionController::MethodNotAllowed',
+      'ActionController::NotImplemented',
+      'ActionController::ParameterMissing',
       'ActionController::RoutingError',
       'ActionController::UnknownAction',
+      'ActionController::UnknownFormat',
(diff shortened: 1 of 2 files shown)


lib/raven/configuration.rb
```diff
@@ -172,16 +172,29 @@ class Configuration
...
+      'ActionController::InvalidCrossOriginRequest',
+      'ActionController::MethodNotAllowed',
+      'ActionController::NotImplemented',
+      'ActionController::ParameterMissing',
       'ActionController::RoutingError',
       'ActionController::UnknownAction',
+      'ActionController::UnknownFormat',
+      'ActionController::UnknownHttpMethod',
+      'ActionDispatch::Http::Parameters::ParseError',
+      'ActionView::MissingTemplate',
```
agrobbin, line 190: @st0012 I'm not sure what the current state of the `v3` branch is, but the inclusion of `ActionView::MissingTemplate` here feels like a bug. While Rails does render a `406 Not Acceptable` when [`ActionView::MissingExactTemplate`](https://github.com/rails/rails/blob/v6.0.3.4/actionpack/lib/action_dispatch/middleware/exception_wrapper.rb#L16) is raised, it renders a `500 Internal Server Error` when `ActionView::MissingTemplate` is raised (via its [default error handling](https://github.com/rails/rails/blob/v6.0.3.4/actionpack/lib/action_dispatch/middleware/exception_wrapper.rb#L8)).

st0012: yeah I think it's too aggressive to exclude the `ActionView::MissingTemplate` by default. will remove it in the upcoming version. thanks for the opinion 👍

st0012: I've added https://github.com/getsentry/sentry-ruby/pull/1092 to remove it

agrobbin: Thanks @st0012!!

PR #1006, opened 2020-08-28
Deprecate dasherized filenames

### LINKED ISSUE #677: Fix inconsistent camelcasing in filenames
everything_should_be_camel_cased.rb

I'll make the change in 3.0, but the old filenames will still work with a deprecation notice for 4.0.

st0012:
We have 2 files with dasherized names:
- `sentry-raven-without-integrations` 
- `raven/integrations/rack-timeout`

This PR allows users to require them with underscored names: 
- `sentry_raven_without_integrations` 
- `raven/integrations/rack_timeout`

Requiring with dasherized names will still be allowed before version `4.0` but it'll print deprecation warnings.

`sentry-raven` is not touched because that's the gem's name.

This closes #677


jeremy:
Note that `rack-timeout` is the gem name. Integrations are [loaded automatically based on gem name](https://github.com/getsentry/raven-ruby/blob/f3edc6e458e121b1fc9de63fbf3153c1da0e1692/lib/raven/base.rb#L78-L89), so any rack-timeout users now get a deprecation warning they can't fix.

PR #1007, opened 2020-08-28
Fix merge error from rack-timeout raven_context on old releases

jonmast:
Originally nils returned from raven_context were permitted, but with the
refactor from #741 this gets passed directly to merge and raises an
error. Switch to returning an empty hash rather than nil on old
rack-timeout releases which don't have "env" set to avoid this problem.

### DIFF
--- a/lib/raven/integrations/rack-timeout.rb
+++ b/lib/raven/integrations/rack-timeout.rb
@@ -10,7 +10,11 @@ module RackTimeoutExtensions
     # Only rack-timeout 0.3.0+ provides the request environment, but we can't
     # gate this based on a gem version constant because rack-timeout does
     # not provide one.
-    { :fingerprint => ["{{ default }}", env["REQUEST_URI"]] } if defined?(env)
+    if defined?(env)
+      { :fingerprint => ["{{ default }}", env["REQUEST_URI"]] }
+    else
+      {}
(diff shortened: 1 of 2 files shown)


st0012:
Thanks for the PR. This PR indeed makes the SDK more robust, which is good 👍 

However, it also means we'll miss the chance to spot bugs that hidden in other parts of this gem. And since the issue has been there for a while, I guess it's not too critical to most of the users. So we can spend more time addressing the issue from the source.

I wonder if this change can fix the issue you mentioned?

```ruby
module RackTimeoutExtensions
  def raven_context
    if defined?(env)
      { :fingerprint => ["{{ default }}", env["REQUEST_URI"]] }
    else
      {}
    end
  end
end
```

jonmast:
@st0012 Yes, simply returning an empty hash in the rack-timeout extension will also solve my problem.

I went with the fallback approach to match the original implementation since I wasn't sure what the intended contract for the method was. Requiring it to always be hash does seem more maintainable, I'm happy to update this.

st0012:
@jonmast thanks for the change, can you also update the PR title and description 😄

jonmast:
@st0012 Updated

st0012 (approved):
thanks for the fix 👍

PR #1010, opened 2020-09-03
Refactor event initialization

st0012:
Currently, `Raven::Event` can be initialized without any options because it can fetch configuration & context directly from the singleton instance. 

While this might be easier for testing setup, it makes the code more error-prone as we don't get full control on the event's attributes even when initializing it directly:

```ruby
options = { #... }
event_1 = Raven::Event.new(options)

# somewhere in the app swaps the configuration
Raven.configuration = another_configuration

# now we have a new event with a completely different setup, even if we pass the same options
event_2 = Raven::Event.new(options)
```

So this PR avoids such problem by enforcing passing `configuration`, `context` and `breadcrumbs` when initializing `Raven::Event`.

### DIFF
--- a/lib/raven/instance.rb
+++ b/lib/raven/instance.rb
@@ -77,7 +77,7 @@ module Raven
     # Send an event to the configured Sentry server
     #
     # @example
-    #   evt = Raven::Event.new(:message => "An error")
+    #   evt = Raven::Event.new(:message => "An errore)
     #   Raven.send_event(evt)
     def send_event(event, hint = nil)
       client.send_event(event, hint)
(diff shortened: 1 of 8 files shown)


lib/raven/instance.rb
```diff
@@ -77,7 +77,7 @@ def configure
     # Send an event to the configured Sentry server
     #
     # @example
-    #   evt = Raven::Event.new(:message => "An error")
+    #   evt = Raven::Event.new(:message => "An errore)
```
Sija, line 80: Sth went wrong with this change.

PR #1011, opened 2020-09-03
Support cleaning up exception backtrace with customized backtrace_cleaner

### LINKED ISSUE #957: Ungroupable error stacktraces: can Raven-Ruby leverage `Rails.backtrace_cleaner`?
Switching to Sentry 10 recently, we started having big issues with error grouping in our Rails application. It didn't take us long to notice that the problem seems to come form same errors having different stacktraces.

The difference is located in the name of methods generated from template files.
I couldn't find better documentation or Rails code extracts, so instead allow me to share this RailsConf video extract instead 😬: [_templates are translated into methods_](https://youtu.be/8Dld5kFWGCc?t=2359).

The stacktrace would look something like that:
image

Focus on `_app_views_home_index_html_erb___2412767935268124114_70177239986900`. This method name is generated from the `app/views/home/index.html.erb` template. I could not pinpoint in Rails code where those strings of digits (fingerprints?) are generated and appended to the method name, but they seem to vary depending on multiple factors (at least the file's content, but it looks like for a given file we've produced multiple values). This generated in Sentry multiple errors that would not be grouped together.

Here's what a stacktrace diff looks like between two errors I would expect to get grouped together on Sentry, but that did not:

This diff was displayed in Sentry's "Similar issues" tab.

There might be settings on Sentry that would allow those similar issues to eventually look close enough to be merged by Sentry, but I wondered what tools were available on my end to clean that up.

That's when I found about [ActiveSupport::BacktraceCleaner](http://api.rubyonrails.org/classes/ActiveSupport/BacktraceCleaner.html). This is what Rails uses to display cleaner, more focused backtraces when an error is displayed.

In fact, [Rails::BacktraceCleaner](https://github.com/rails/rails/blob/b66235d432d0505857ff667e0a1ddecfb5640a56/railties/lib/rails/backtrace_cleaner.rb#L19) introduces a filter that cleans up the template method names from the backtrace, which is exactly what I was looking for. It also removes a lot of noise which is not really what we're looking for when first debugging the application on Sentry. (Silencers are what splits the "application trace" from the "framework trace".)

I though it would be great if I could just apply `Rails.backtrace_cleaner` to the stack traces before they get sent to Sentry, but the best thing I found myself able to do (without rewriting that logic), was to write a small monkey patch. 🙈 

```rb
module Raven
  class Backtrace
    class << self
      alias_method :orig_parse, :parse
      def parse(backtrace, opts = {})
        orig_parse(::Rails.backtrace_cleaner.clean(backtrace), opts)
      end
    end
  end
end
```

I hijacked the `Raven::Backtrace.parse` method in order to clean up the backtrace while it's still a raw array of lines, before Raven works on it and produces frames it will send to Sentry. The result, for the same error I screenshot at the top, can be seen in this other screenshot:
image

That's quite satisfying, and should fix our problems with errors not getting merged together (also should help making the stacktrace displayed on Sentry more workable, by [focusing](https://github.com/rails/rails/blob/0196551e6039ca864d1eee1e01819fcae12c1dc9/activesupport/lib/active_support/backtrace_cleaner.rb#L4-L8) on the parts of the code that are most likely to be producing errors.

Now the part I'm not satisfied about, obviously, is that I had to write a monkey patch. 🐒

My question is: is there any better mechanism (provided by Raven, or else) that would allow an application to apply a `BacktraceCleaner` to the raw backtrace before it gets parsed by Raven? (I am aware of Raven's processors, but those happen _after_ the stacktrace has been parsed...)

st0012:
This PR adds 2 new features:

## New config option: `backtrace_cleanup_callback`
It takes a lambda/proc object (default is `nil`) and will be called with exception's backtrace. Here's an example usage:

```ruby
Raven.configure do |config|
  config.backtrace_cleanup_callback = lambda do |backtrace|
    Rails.backtrace_cleaner.clean(backtrace)
  end
end
```

## Automatically clean Rails app's backtrace

With the Rails integration, it'll automatically use a customized `Raven::Rails::BacktraceCleaner` to clean up exception's backtrace. This can be disabled by:

```ruby
Raven.configure do |config|
  config.backtrace_cleanup_callback = nil
end
```

### Raven::Rails::BacktraceCleaner

This cleaner is pretty much like Rails 6's [backtrace cleaner](https://github.com/rails/rails/blob/master/railties/lib/rails/backtrace_cleaner.rb) but without silencers. The main reason to add this cleaner is to remove template methods from the trace, e.g.

```
app/views/welcome/view_error.html.erb in _app_views_welcome_view_error_html_erb__2807287320172182514_65600 at line 1
```

will become

```
app/views/welcome/view_error.html.erb at line 1
```

This can help Sentry group issues more accurately. See  #957  for more information about this.

Closes #957

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -55,6 +55,9 @@ Style/CommentedKeyword:
 Style/RescueModifier:
   Enabled: false
 
+Style/RegexpLiteral:
+  Enabled: false
+
 Style/StringLiterals:
   Enabled: false
 
--- a/lib/raven/backtrace.rb
+++ b/lib/raven/backtrace.rb
@@ -94,6 +94,8 @@ module Raven
     def self.parse(backtrace, opts = {})
       ruby_lines = backtrace.is_a?(Array) ? backtrace : backtrace.split(/\n\s*/)
 
+      ruby_lines = opts[:configuration].backtrace_cleanup_callback.call(ruby_lines) if opts[:configuration]&.backtrace_cleanup_callback
+
       filters = opts[:filters] || []
       filtered_lines = ruby_lines.to_a.map do |line|
         filters.reduce(line) do |nested_line, proc|
--- a/lib/raven/configuration.rb
+++ b/lib/raven/configuration.rb
@@ -118,6 +118,19 @@ module Raven
     # Otherwise, can be one of "http", "https", or "dummy"
     attr_accessor :scheme
 
+    # a proc/lambda that takes an array of stack traces
+    # it'll be used to silence (reduce) backtrace of the exception
+    #
+    # for example:
+    #
+    # ```ruby
+    # Raven.configuration.backtrace_cleanup_callback = lambda do |backtrace|
+    #   Rails.backtrace_cleaner.clean(backtrace)
+    # end
+    # ```
+    #
+    attr_accessor :backtrace_cleanup_callback
+
     # Secret key for authentication with the Sentry server
     # If you provide a DSN, this will be set automatically.
     #
--- a/lib/raven/event.rb
+++ b/lib/raven/event.rb
@@ -175,7 +175,7 @@ module Raven
     end
 
     def stacktrace_interface_from(backtrace)
-      Backtrace.parse(backtrace).lines.reverse.each_with_object([]) do |line, memo|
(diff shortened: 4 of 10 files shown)


davidstosik (approved):
👏  Thanks a lot, this is looking great, and provides a nice modularity!

PR #1021, opened 2020-09-11
allow changing the default logger for events

ashkulz:
Can't use `logger` as that key is already in use.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - REFACTOR: Refactor event initialization ([#1010](https://github.com/getsentry/raven-ruby/pull/1010))
 - REFACTOR: Unify breadcrumb loggers activation ([#1016](https://github.com/getsentry/raven-ruby/pull/1016))
 - FIX: Fix merge error from rack-timeout raven_context on old releases ([#1007](https://github.com/getsentry/raven-ruby/pull/1007))
+- FEATURE: allow changing the default logger in events
 
 ## 3.0.4
 
--- a/lib/raven/configuration.rb
+++ b/lib/raven/configuration.rb
@@ -168,6 +168,9 @@ module Raven
     # Should the SSL certificate of the server be verified?
     attr_accessor :ssl_verification
 
+    # Default logger for events. String.
+    attr_accessor :default_logger
+
     # Default tags for events. Hash.
     attr_accessor :tags
 
@@ -270,6 +273,7 @@ module Raven
       self.timeout = 2
       self.transport_failure_callback = false
       self.before_send = false
+      self.default_logger = :ruby
     end
 
     def server=(value)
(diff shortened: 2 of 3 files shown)


ashkulz:
Can you review this, @st0012?

st0012:
I don't think the use case (configurable event logger) is important enough to add a new global config. However, I do think every event's logger should be configurable. So a change like would be better to me

```ruby
# event.rb
def initialize(options)
  # ....
  self.configuration = options[:configuration]
  # ....
  self.logger        = options[:logger] || configuration.logger || :ruby
end
```

ashkulz:
`configuration.logger` is a ruby `Logger` so I don't think what you say will work.

Background: we often just set up the context and use the default exception handler i.e. there's no explicit call to `Raven.capture_exception` -- hence the above approach. Not sure if you want to introduce some methods in the context instead?

st0012:
hmm, I see. what's the reason to specify a different logger though?

ashkulz:
We want the logger to be set the default logger to `web` (or `qa` for QA servers) while we use specific loggers for areas e.g. `api`, `mobile`, `background_job` etc -- effectively the "nature" of the crash.

st0012:
So does the `web` and `qa` servers...etc. actually use different logging setups? Or the goal is just to identify the source of the events with the `logger` tag? If it's the latter, I think you can achieve the goal with customized tags? Like

```ruby
# in api app's config
config.tags.merge!(source: "api")

# in background job worker's config
config.tags.merge!(source: "background_job")
```

If you can do

```ruby
# in api app's config
config.default_logger = "api"

# in background job worker's config
config.default_logger = "background_job"
```

adding customized tags should be pretty similar to it?

ashkulz:
Sure, I can do that -- but using `logger` is the recommended way for the [delayed_job](https://github.com/getsentry/raven-ruby/blob/b9d150ec1d7b1ba070550c91d3b694c519f132eb/lib/raven/integrations/delayed_job.rb#L35) and [rake](https://github.com/getsentry/raven-ruby/blob/21669e97ee704efe40432ae66a1078cf74030ca6/lib/raven/integrations/rake.rb#L12) integrations, so I wanted to use that same field (instead of just another tag). What is the use of the `logger` otherwise -- just to indicate the language? That's already covered by the `platform`, I think.

ashkulz:
Are you okay with adding a field `logger` to `Raven::Context` instead? This would then be used in `Raven::Instance#set_core_attributes_from_context.

st0012:
@ashkulz so, an update: the `logger` tag isn't really used nowadays. we'll remove it in v4.0, so you should be able to add it as a customized tag then

ashkulz:
Thanks, already decided to go with a tag-based approach as you'd suggested :+1:

PR #1023, opened 2020-09-11
Allow blocks on user_context

philipbjorge:
This makes user_context block receivable and it defines a scope for the tags/extra.

Before:

```
Raven.user_context id: user.id
Raven.capture_message("foo")
Raven.user_context
```

After:
```
Raven.user_context id: user.id do
  Raven.capture_message("foo")
end
```

---

I believe I updated this in a way that did not break anyone's expectations of the return value.
*Note* The method documentation for `user_context` says that it `Merges with existing context (if any).` but that is not actually the case, nor have I made it work as documented.

### DIFF
--- a/spec/raven/instance_spec.rb
+++ b/spec/raven/instance_spec.rb
@@ -247,9 +247,19 @@ RSpec.describe Raven::Instance do
 
   describe "#user_context" do
     context "without a block" do
-      it "empties the user context when called" do
+      it "empties the user context when called without options" do
         subject.context.user = { id: 1 }
         expect(subject.user_context).to eq({})
+      end
+      
+      it "empties the user context when called with nil" do
+        subject.context.user = { id: 1 }
+        expect(subject.user_context(nil)).to eq({})
+      end
+      
+      it "empties the user context when called with {}" do
+        subject.context.user = { id: 1 }
+        expect(subject.user_context({})).to eq({})
       end
 
       it "returns the user context when set" do
--- a/spec/raven/instance_spec.rb
+++ b/spec/raven/instance_spec.rb
@@ -245,6 +245,20 @@ RSpec.describe Raven::Instance do
     end
   end
 
+  describe "#user_context" do
+    context "without a block" do
+      it "empties the context when called" do
+        subject.context.user = { id: 1 }
+        expect(subject.user_context).to eq({})
+      end
+
+      it "returns the context when set" do
+        expected = { id: 1 }
+        expect(subject.user_context(expected)).to eq(expected)
+      end
+    end
+  end
+
   describe "#tags_context" do
     let(:default) { { :foo => :bar } }
     let(:additional) { { :baz => :qux } }
--- a/spec/raven/instance_spec.rb
+++ b/spec/raven/instance_spec.rb
@@ -251,12 +251,12 @@ RSpec.describe Raven::Instance do
         subject.context.user = { id: 1 }
         expect(subject.user_context).to eq({})
       end
-      
+
       it "empties the user context when called with nil" do
         subject.context.user = { id: 1 }
         expect(subject.user_context(nil)).to eq({})
       end
-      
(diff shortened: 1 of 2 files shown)


st0012:
thanks for the PR!

I think it's a good idea to align `user_context` with other context APIs. But we need to consider the following things before making this change:
- Will it cause great impact on the existing clients?
- How much benefit can we get from this change?
- Does it comply Sentry's SDK development guideline?

### Will it cause great impact on the existing clients?

I know some projects (like my company's) have adopted the fact that `user_context` mutates the context's user information and doesn't change it back. So this change will definitely break them in an implicit way. Which means they'll find the user context is missed from Sentry, and only find the cause after searching around the repo and find this PR. I think this will create a great impact on some of the clients.

### How much benefit can we get from this change?

How about the benefit though? We do get a less confusing interface when compared with others. But one thing I'm not sure about is that from my understanding, we generally don't switch between users within the same request. So unlike `extra_context` or `tags_context`, the need like below should be quite rare:

```ruby
Raven.user_context id: user.id do
  Raven.capture_message("foo") # sends with user
end

Raven.capture_message("foo") # sends without user....why?
```

### Does it comply Sentry's SDK development guideline?

According to Sentry's [SDK development document](https://develop.sentry.dev/sdk/unified-api/)

> scope.set_user(user): Shallow merges user configuration (email, username, …). Removing user data is SDK-defined, either with a remove_user function or by passing nothing as data.

API for user data is meant to merge the given value into the current context. So I think the current behavior is fine.

Of course, our current APIs don't align the guideline very well, like the `extra_context` and `tags_context` behave differently than it. And in the following version 4.0 (coming in later this year), this should be well-addressed.


### Conclusion

So based on the above points, I don't plan to make this change right now. When building the upcoming 4.0 version, I will consider adding similar APIs though. Thanks again for the idea & PR 👍

philipbjorge:
@st0012 

> I know some projects (like my company's) have adopted the fact that user_context mutates the context's user information and doesn't change it back. So this change will definitely break them in an implicit way. Which means they'll find the user context is missed from Sentry, and only find the cause after searching around the repo and find this PR. I think this will create a great impact on some of the clients.

I introduced tests to ensure that the current behavior of `user_context` is maintained.

`Raven.user_context('id' => 1, 'email' => 'foo@example.com')` will continue to mutate the user context and not restore it.

So I don't think this will cause any change for existing clients? Am I missing something?

> But one thing I'm not sure about is that from my understanding, we generally don't switch between users within the same request.

We use Sentry to record errors in background jobs where we do switch user contexts between and within jobs. Because the user context is thread local, it leaks between executions of sidekiq jobs.

I understand that this doesn't make much sense in the request context, but we wouldn't be using this there.

----

All that being said, it's fine if this doesn't make it upstream but I'm not sure I agree with your assessment of this PR.

Thanks for all the hard work on Sentry :) ✌️

st0012:
> I introduced tests to ensure that the current behavior of user_context is maintained.

I see, sorry that I didn't notice that. I assumed the goal is to completely match other API's behavior.

> We use Sentry to record errors in background jobs where we do switch user contexts between and within jobs. 
thanks for bringing this up, I think it makes sense in such contexts

If it doesn't break current usages, I think it's worth considering adding a support for different usages. I'll give it a closer look later this week and can probably merge it then.

spec/raven/instance_spec.rb
```diff
@@ -245,6 +245,40 @@ def ivars(object)
     end
   end
 
+  describe "#user_context" do
+    context "without a block" do
+      it "empties the user context when called" do
```
st0012, line 250: nitpick: think this should be `"empties the user context when called without options"`

st0012 (changes requested):

st0012 (approved):

st0012:
@philipbjorge thanks for the great work 👍

PR #1026, opened 2020-09-16
Revert rack-timeout deprecation

### LINKED ISSUE #1025: Unfixable deprecation warning with rack-timeout integration
Using Sentry with rack-timeout now results in a deprecation warning that the wrong file was required.

#1006 got rid of all dasherized filenames. Integrations are automatically loaded based on gem name, however, and rack-timeout is dasherized. So using `rack-timeout` will always require `raven/integrations/rack-timeout`, triggering the deprecation warning.

The workaround is to explicitly specify integrations, omitting rack-timeout, and manually require `raven/integrations/rack_timeout` instead.

I think we should just revert this portion of #1006. Keep the dasherized integration filename that matches the gem name.

st0012:
The rack-timeout integration is automatically required along with the gem itself. So there's no need to deprecate the original require name, which is causing noise to users.

Closes #1025


jeremy:
Thanks @st0012!

PR #1027, opened 2020-09-16
Return value of `rescue_with_handler` when intercepting ActiveJob exceptions

bensheldon:
## Description

When using ActiveJob's `rescue_from` (or `retry_on` / `discard_on` which are implemented with `rescue_from`), ActiveJob will return the captured exception. This PR fixes Sentry's ActiveJob instrumentation to preserve that behavior.


st0012 (approved):

st0012:
@bensheldon thanks 😄

Issue #891, opened 2019-02-24
Automatically add request_id tag


tilsammans:
The Python library now detects and tags a request id. This technique was recently [in the Sentry blog](https://blog.sentry.io/2019/01/31/using-nginx-sentry-trace-errors-logs)

Seems like a good addition for Rails too. Need a pull request?

IvanChou:
I need this, too

st0012:
this has been added in #1120 so closing now

PR #1033, opened 2020-09-19
Add request id to headers if exists

berniechiu:
## Description

- This should add enhancement for https://github.com/getsentry/sentry-ruby/issues/891, which allows Sentry server to auto parse request id from default Rails stack
- Follow what https://github.com/skylightio/skylight-ruby/blob/3e742ea00f41b3c57c3f0305ed82c794b396d7ab/lib/skylight/middleware.rb#L124 might retrieve from the ID from Rails Middleware

### DIFF
--- a/lib/raven/integrations/rack.rb
+++ b/lib/raven/integrations/rack.rb
@@ -67,6 +67,8 @@ module Raven
   end
 
   module RackInterface
+    REQUEST_ID_HEADERS = %w(action_dispatch.request_id HTTP_X_REQUEST_ID).freeze
+
     def from_rack(env_hash)
       req = ::Rack::Request.new(env_hash)
 
@@ -81,6 +83,15 @@ module Raven
     end
 
     private
+
+    # Request ID based on ActionDispatch::RequestId
+    def read_request_id_from(env_hash)
+      REQUEST_ID_HEADERS.each do |key|
+        request_id = env_hash[key]
+        return request_id if request_id
+      end
+      nil
+    end
 
     # See Sentry server default limits at
     # https://github.com/getsentry/sentry/blob/master/src/sentry/conf/server.py
@@ -101,6 +112,7 @@ module Raven
         begin
           key = key.to_s # rack env can contain symbols
           value = value.to_s
+          next memo['X-Request-Id'] ||= read_request_id_from(env_hash) if REQUEST_ID_HEADERS.include?(key)
           next unless key.upcase == key # Non-upper case stuff isn't either
 
           # Rack adds in an incorrect HTTP_VERSION key, which causes downstream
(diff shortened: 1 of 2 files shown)


lib/raven/integrations/rack.rb
```diff
@@ -67,6 +67,8 @@ def call(env)
   end
 
   module RackInterface
+    REQUEST_ID_HEADERS = Set.new(%w(HTTP_X_REQUEST_ID action_dispatch.request_id)).freeze
```
st0012, line 70: If this is not mutable to users, why does it need to be a set instead of just an array?

lib/raven/integrations/rack.rb
```diff
@@ -82,6 +84,11 @@ def from_rack(env_hash)
 
     private
 
+    # Request ID based on ActionDispatch::RequestId
+    def read_request_id_from(env_hash)
+      env_hash['action_dispatch.request_id'] || env_hash['HTTP_X_REQUEST_ID']
```
st0012, line 89: We should make sure the keys we use here syncs with the `REQUEST_ID_HEADERS` constant automatically. So I prefer something like

```ruby
def read_request_id_from(env_hash)
  REQUEST_ID_HEADERS.each do |key|
    request_id = env_hash[key]
    return request_id if request_id
  end
  nil
end
```

st0012 (changes requested):
thanks for the PR 👍  just added some small suggestions
also, can you rebase this PR  (or cherry-picking the commits) against  [v3-1-stable](https://github.com/getsentry/raven-ruby/tree/v3-1-stable) instead of master? 
sorry I didn't make it clear but we currently only accept contributions to the `3.1.x` versions before the v4 is finished. (I have updated the readme & PR template for this though)

lib/raven/integrations/rack.rb
berniechiu: single source of truth makes sense~ : )

lib/raven/integrations/rack.rb
berniechiu: oh~ I was thinking of a faster `.include?` through `Set`, which could be faster with O(1) instead of O(n), but I think only two elements I guess it's fine here, change back to array

berniechiu:
> thanks for the PR 👍 just added some small suggestions
also, can you rebase this PR (or cherry-picking the commits) against v3-1-stable instead of master?
sorry I didn't make it clear but we currently only accept contributions to the 3.1.x versions before the v4 is finished. (I have updated the readme & PR template for this though)

No prob~~

st0012 (approved):

st0012:
@berniechiu thanks for the work 👍

PR #1042, opened 2020-10-01
fix: correct extra_context method name and ussage for README

salzig:
This is a Documentation fix. 

The README stated that `Raven.tags` (`Raven.respond_to? :tags => false`) and `Raven.extra` (`Raven.respond_to? :extra => false`) should be available to merge in additional details, but that seems to be false. Instead there are two methods that merge a passed in hash (defined at: `lib/raven/instance.rb` Line 194 and 209).


st0012:
Thanks for spotting that!

PR #1048, opened 2020-10-02
Add Scope setters

st0012:
This PR adds the [standard setter APIs](https://develop.sentry.dev/sdk/unified-api/#scope) to the `Scope` class.

### DIFF
--- a/sentry-ruby/lib/sentry/scope.rb
+++ b/sentry-ruby/lib/sentry/scope.rb
@@ -2,7 +2,9 @@ require "sentry/breadcrumb_buffer"
 
 module Sentry
   class Scope
-    attr_accessor :transactions, :contexts, :extra, :tags, :user, :level, :breadcrumbs, :fingerprint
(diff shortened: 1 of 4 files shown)


sentry-ruby/lib/sentry/scope.rb
```diff
@@ -47,6 +48,69 @@ def dup
...
+      @tags = tags_hash
+    end
+
+    def set_tag(key, value)
+      @tags.merge!(key => value)
+    end
+
+    def set_contexts(contexts_hash)
+      check_argument_type!(contexts_hash, Hash)
+      @contexts = contexts_hash
```
fsateler, line 76: Hi! is there a specific reason this method overwrites `@contexts` instead of merging? Tags and extras (which are conceptually similar) are both merged

st0012: I believe this is a bug. thanks for raising this, I'll fix it 🙂

PR #1049, opened 2020-10-03
Use latest sdk name format

bruno-garcia:
I noticed a few sample events with `sentry-ruby`.
The new SDK name format is: `sentry.lang`. With more specific stuff under it like `sentry.lang.framework`.


st0012:
hey thanks, I just made the change in https://github.com/getsentry/sentry-ruby/commit/97b90908c1adc17eabef4cd049b63cd54eb22f55 with the test update as well

PR #1050, opened 2020-10-04
Rack integration

st0012:
usage:

```ruby
require 'rack'
require 'sentry-ruby'
  
Sentry.init do |config|
  config.dsn = 'https://my_dsn'
end
  
app = Rack::Builder.app do
  use Sentry::Rack::CaptureException
  run lambda { |env| raise "Rack down" }
end
```


sentry-ruby/lib/sentry/rack/capture_exception.rb
```diff
@@ -0,0 +1,51 @@
...
+      end
+
+      def call(env)
+        # this call clones the main (global) hub
+        # and assigns it to the current thread's Sentry#get_current_hub
+        # it's essential for multi-thread servers (e.g. puma)
+        Sentry.clone_hub_to_current_thread unless Sentry.get_current_hub
+        # this call creates an isolated scope for every request
+        # it's essential for multi-process servers (e.g. unicorn)
+        Sentry.with_scope do |scope|
```
st0012, line 15: because there are multi-thread/multi-process servers in the Ruby ecosystem, we need to support both of them.

for multi-thread servers, like `puma` or `webrick`, `Sentry.clone_hub_to_current_thread` sets up the a cloned hub in every new thread. so each thread's sentry scope, client...etc. won't interfere each other

for multi-process servers, like `unicorn`, the above line is redundant as every request will all run on the process' main thread. what it needs is an isolated scope for that particular request. so this is why we use `Sentry.with_scope` here.

PR #1052, opened 2020-10-08
Add Layer class to contain both client and scope in the same stack

st0012:


sentry-ruby/lib/sentry/hub.rb
```diff
@@ -6,21 +6,26 @@ class Hub
...
+    end
+
+    def current_scope
+      current_layer&.scope
     end
 
     def bind_client(client)
-      @clients << client
+      layer = Layer.new(client, current_scope)
+      @stack << layer
```
HazAT, line 24: bind_client should replace the client in the current layer

sentry-ruby/lib/sentry/hub.rb
```diff
@@ -42,15 +47,15 @@ def push_scope
           Scope.new
         end
 
-      @scopes.push(new_scope)
+      @stack << Layer.new(current_client, new_scope)
```
HazAT, line 50: configure_scope shouldn't push a layer

HazAT (changes requested):

sentry-ruby/lib/sentry/hub.rb
st0012: this is inside `push_scope` actually

HazAT: oh

HazAT (approved):

Issue #1051, opened 2020-10-07
WIP: Unified API


HazAT:
Legend: 
(L) = Low Priority
(M) = Medium Priority
(H) = High Priority

sentry-ruby/lib/sentry/client.rb

- [x] (M) def encode(event)
    Should be in transport (or in an abstraction)

- [x] (M) def generate_auth_header
    Should also be in transport

    Same: 

    ```
    PROTOCOL_VERSION = '5'
    USER_AGENT = "sentry-ruby/#{Sentry::VERSION}"
    CONTENT_TYPE = 'application/json'
    ```

- [x] (H) functions in the Client should receive a Scope that the Hub passes.
https://github.com/getsentry/sentry-php/blob/master/src/Client.php#L126

The reason we want to do this is to provide this functionallity:

Capture with provided a block to change the scope only for one capture call
```
SentrySDK.capture(error: error) { (scope) in
    // Changes in here will only be captured for this event
    // The scope in this callback is a clone of the current scope
    // It contains all data but mutations only influence the event being sent
    scope.setTag(value: "value", key: "myTag")
}
```

Capture Call by passing a new Scope instance:
```
// By explicity just passing the scope, only the data in this scope object will be added to the event
// The global scope (calls to configureScope) will be ignored
// Only do this if you know what you are doing, you loose a lot of useful info
// If you just want to mutate what's in the scope use the callback, see: captureError
SentrySDK.capture(exception: exception, scope: scope)
```

Or the normal way of using `configure_scope`

sentry-ruby/lib/sentry/configuration.rb

- [x] (L) Can we rename this to Options? Or is this not idiomatic?
It's called Options in most other SDKs

- [x] (M) attr_accessor :project_id, attr_accessor :public_key , attr_accessor :secret_key, attr_accessor :host, attr_accessor :path, attr_accessor :port
Can we remove this?

- [x] (H) Remove all `attr_accessor :sanitize*` fields

- [x] (H) attr_accessor :scheme
This should become a transport option, where users can provide their own instance of a transport to send stuff to Sentry
Also the transport should just switch depending on if the DSN has http or https

- [x] (H) attr_accessor :ssl* Should all be TransportOptions, see JavaScript
attr_accessor :timeout -> TransportOptions
attr_reader :encoding -> TransportOptions
attr_accessor :open_timeout -> TransportOptions
attr_accessor :proxy -> TransportOptions

- [x] (M) attr_reader :transport_failure_callback 
Remove

## Envelopes

[...]

PR #1054, opened 2020-10-08
Extract dsn processing out of Configuration class

st0012:
This addresses

> (M) attr_accessor :project_id, attr_accessor :public_key , attr_accessor :secret_key, attr_accessor :host, attr_accessor :path, attr_accessor :port
Can we remove this?

in #1051

### DIFF
--- a/sentry-ruby/lib/sentry/client.rb
+++ b/sentry-ruby/lib/sentry/client.rb
@@ -7,7 +7,9 @@ module Sentry
 
     def initialize(configuration)
       @configuration = configuration
-      @transport = case configuration.scheme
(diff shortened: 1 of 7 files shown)


sentry-ruby/lib/sentry/client.rb
```diff
@@ -7,7 +7,9 @@ class Client
 
     def initialize(configuration)
       @configuration = configuration
-      @transport = case configuration.scheme
+      scheme = configuration.scheme || configuration.dsn.scheme
```
st0012, line 10: I'll remove the scheme concept from configuration in another PR, along with other Transport options

Issue #1041, opened 2020-10-01
uninitialized constant Delayed::Plugins::Raven::Utils after upgrading to vs 3.1.1


atstockland:
**Describe the bug**
Background jobs are failing with error "uninitialized constant Delayed::Plugins::Raven::Utils"

**To Reproduce**
I'm not sure how to reproduce this.  I was surprised to not find an existing bug on the matter.

**Environment**
 - Ruby Version: 2.7.1
 - SDK Version: 3.1.1
 - Integration Versions (if any):
  - Rails: 6.0.3.3
  - delayed_job: 4.1.8
  - delayed_job_active_record: 4.1.4

**Raven Config**

# require 'raven-ruby'

Raven.configure do |config|

  config.dsn = 'https://xxx.ingest.sentry.io/xxx'
  config.environments = %w[staging production]
  config.sanitize_fields = Rails.application.config.filter_parameters.map(&:to_s)
  config.excluded_exceptions += ['SignalException']
  config.async = lambda { |event|
    SentryJob.perform_later(event)
  }

end

atstockland:
After pushing out an unrelated update on my app, I reloaded the failed job and it processed as expected.  Seems this was an anomaly.

toao:
I have the same issue, looks like a loading/namespace issue in https://github.com/getsentry/sentry-ruby/blob/v3-1-stable/lib/raven/integrations/delayed_job.rb#L33

The ```Utils::ContextFilter``` is used without an additional namespace, that's why it gets extended to ```Delayed::Plugins::Raven::Utils``` which then throws the error.

I guess a fix would be to simply use ```::Raven::Utils::ContextFilter.filter_context(job.payload_object.job_data)``` instead.

PR #1057, opened 2020-10-10
Fix Delayed::Plugins::Raven when job raises exception

eitoball:
## Description

This PR should fix `Delayed::Plugins::Raven` from raising `NameError: uninitialized constant Delayed::Plugins::Raven::Utils
` when job raises an exception during executing.

See #1041


st0012 (approved):
this fix and the test looks awesome! thanks 😄

PR #1078, opened 2020-10-23
Add send_default_pii config option

st0012:
Because the new SDK won't perform data sanitization anymore, by default it shouldn't send any sensitive data. So this PR removes some information from the request information:

1. the user's IP address
2. the request's body/form data
3. the cookies

But this behavior can be changed with the new `send_default_pii` config option. If the SDK users still want to receive sensitive data for debugging purpose, `config.send_default_pii = true` will resume the behavior.

### DIFF
--- a/sentry-ruby/spec/sentry/event_spec.rb
+++ b/sentry-ruby/spec/sentry/event_spec.rb
@@ -77,6 +77,85 @@ RSpec.describe Sentry::Event do
     end
   end
 
+  context 'rack context specified' do
+    require 'stringio'
+
+    before do
+      Sentry.init do |config|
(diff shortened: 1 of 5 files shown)


sentry-ruby/spec/sentry/event_spec.rb
```diff
@@ -77,6 +77,85 @@
...
+        server_name: 'foo.local'
+      )
+
+      Sentry::Event.new(
+        options: options,
+        configuration: Sentry.configuration,
+      )
+    end
+
+    context "without config.send_default_pii = true" do
```
st0012, line 120: @HazAT can you review this and the below scenarios to see if I missed any field/information to ignore?

Issue #1061, opened 2020-10-12
[4.0] capture* with explicit Scope


HazAT:
Follow up: https://github.com/getsentry/sentry-ruby/issues/1051

From our internal discussion:

# Explicit Scope aka. second capture argument/kwargs

Approver: Daniel Griesser
Contributors: Kamil Ogórek, Armin Ronacher
Created: May 15, 2020 3:04 PM
Created by: Daniel Griesser
Decision Date: May 15, 2020
Driver: Daniel Griesser
Informed: Markus Unterwaditzer, Árpád Borsos, Rodolfo Carvalho, Katie Byers, Bruno Garcia, Manoel Aranda, Philipp Hofmann
Status: Completed

# Decision: Option A

Note: We will implement and add this to SDKs on a per need basis. Meaning the next time we touch it we should implement this.

tl;dr

```tsx
// The global scope state will never be mutated
// the second arg is used as a replacement for `withScope`

captureException(new Error("test"), (scope) => {
   // Used with a callback clones the current scope and mutates the scope
});

// Used with kwargs clones the current scope and mutates the scope
captureException(new Error("test"), {level: "debug", tags={a: "b"}});

// Used with Scope instance
const scope = new Scope();
scope.setTags({a: "b"});
captureException(new Error("test"), scope);
```

---

**Goals:**

We want to enable one-off scope modification for capture calls without having to configure/push scope.

**Motivations:**

- Users find the current API is too complex for single cases
- If a hub is shared between threads then push/pop scope is unsafe. Sharing hubs is already permitted between threads.

**Github References:**
[https://github.com/getsentry/sentry-javascript/issues/1607](https://github.com/getsentry/sentry-javascript/issues/1607#issuecomment-629223346)

# Option A: Explicit Scopes with Well Known Data

This proposal proposes to let users instantiate and modify scopes and pass these to the global and per-hub capture methods.

Capture methods can also take well-known data keyword arguments if the language supports that and create an implicit scope. This can also be a plain old data object in javascript.

For instance `user={}` sets the user attribute on a new scope.

### Example

```python
# Make a new scope and pass it to a single capture.  In this example an
# attachment is added to the scope
scope = Scope()
scope.add_attachment(...)
capture_exception(Error(...), scope=scope)

# Here we use the scope in a way that sets the user for a single call.
scope = Scope()
scope.set_user({'id': 42})
capture_message("Whatever", scope=scope)

# Additionally `user` is well-known data and can be set directly.  This is
[...]

PR #1086, opened 2020-10-31
Allow passing custom scope to Hub#capture* helpers

st0012:
This PR implements and closes https://github.com/getsentry/sentry-ruby/issues/1061. See `sentry-ruby/spec/sentry/hub_spec.rb`'s `#capture_*` specs for examples

### DIFF
--- a/sentry-ruby/spec/sentry/hub_spec.rb
+++ b/sentry-ruby/spec/sentry/hub_spec.rb
@@ -4,60 +4,120 @@ RSpec.describe Sentry::Hub do
   let(:configuration) do
     config = Sentry::Configuration.new
     config.dsn = DUMMY_DSN
+    config.transport.transport_class = Sentry::DummyTransport
     config
   end
   let(:client) { Sentry::Client.new(configuration) }
+  let(:transport) { client.transport }
   let(:scope) { Sentry::Scope.new }
 
   subject { described_class.new(client, scope) }
 
-  before do
-    allow(client).to receive(:send_event)
-  end
+  shared_examples "capture_helper" do
+    context "with custom attributes" do
+      it "updates the event with custom attributes" do
+        subject.send(capture_helper, capture_subject, tags: { foo: "bar" })
 
+        event = transport.events.last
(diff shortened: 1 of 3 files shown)


sentry-ruby/spec/sentry/hub_spec.rb
```diff
@@ -4,71 +4,228 @@
...
+        scope
+      end
+
+      it "accepts a custom scope" do
+        subject.capture_message(message, scope: new_scope)
+
+        event = transport.events.last
+        expect(event.tags).to eq({ custom_scope: true })
+      end
+    end
```
HazAT, line 73: Can we also set something here on the scope to make sure the passed scope instance is used and not the global one?

sentry-ruby/spec/sentry/hub_spec.rb
```diff
@@ -4,71 +4,228 @@
...
+      it 'yields the scope to a passed block' do
+        subject.capture_message(message) do |scope|
+          scope.set_tags({ foo: "bar" })
+        end
 
-      expect(event.tags).to eq({ foo: "bar" })
-      expect(event.server_name).to eq("foo.local")
+        event = transport.events.last
+        expect(event.tags).to eq({ foo: "bar" })
+      end
```
HazAT, line 83: Same here, using the block should be a "merge" with the existing global scope.

HazAT (changes requested):
Also I think we can de-duplicate the tests a bit.
`captureException`, `captureMessage` and `captureEvent` all have the same tests with content.
Maybe we can switch it so  `captureException` and `captureMessage` only test that they accept all parameter, where `captureEvent` really tests all the possible combinations.

sentry-ruby/spec/sentry/hub_spec.rb
st0012: this is exactly this test for: only the `new_scope` has the `custom_scope: true` tag and the global one doesn't

HazAT (approved):

PR #1093, opened 2020-11-02
Correct Processors configuration and Integration links

calvinhughes:
## Description
Some of the links in the README currently go to a 404, so I've amended these to go to the correct URLs on the documentation site.


st0012 (approved):
thanks for catching them 👍

PR #1094, opened 2020-11-05
meta: Add GHA for artifacts and craft

HazAT:


st0012:
closing this as we already have a working workflow now

PR #1101, opened 2020-11-07
Remove Event options

st0012:
After #1086, all the APIs use a temporary Scope for assigning custom attributes to the Event. So Event's options aren't really used anymore. And this PR just removes the related code.


sentry-ruby/lib/sentry/event.rb
```diff
@@ -113,6 +105,10 @@ def to_json_compatible
     end
 
     def add_exception_interface(exc)
+      if exc.respond_to?(:sentry_context)
+        @extra.merge!(exc.sentry_context)
+      end
```
jeremy, line 110: Subtle regression here.

Previously, `#sentry_context` could return a hash like `{ extra: …, tags: …, fingerprint: … }` so you could enrich every aspect of the event captured by an exception.

Now, the `#sentry_context` method is treated as `extra` context only.

sentry-ruby/lib/sentry/client.rb
```diff
@@ -54,31 +39,16 @@ def capture_event(event, scope)
       event
     end
 
-    def event_from_exception(exception, **options)
-      exception_context =
-        if exception.instance_variable_defined?(:@__sentry_context)
-          exception.instance_variable_get(:@__sentry_context)
-        elsif exception.respond_to?(:sentry_context)
-          exception.sentry_context
```
jeremy, line 62: Note how `#sentry_context` used to be able to provide every aspect of event context, not just `extra`.

sentry-ruby/spec/sentry/client_spec.rb
```diff
@@ -2,10 +2,9 @@
 
 class ExceptionWithContext < StandardError
   def sentry_context
-    { extra: {
-      'context_event_key' => 'context_value',
-      'context_key' => 'context_value'
-    } }
+    {
+      foo: "bar"
+    }
```
jeremy, line 7: Illustrates the breaking change.

sentry-ruby/lib/sentry/event.rb
st0012: Hey I'm sorry for the trouble. It's an undocumented feature that's ported from `sentry-raven`, which I thought no body was using (I didn't mention it in the migration guide and didn't receive any report). So I didn't think much when removing it. Would you mind sharing your usage of it?

jeremy: Sorry, I should have mentioned sooner. We avoided upgrading when this changed and did not alert you at the time.

We used it to "enrich" exceptions with the relevant Sentry context. Here are some examples:
```ruby
# Cross-cutting exceptions like database deadlocks and Redis timeouts can
# crop up anywhere. Sentry groups exceptions by backtrace, though, so we
# end up with tons of separate events for the same underlying cause.
# Fix by overriding the exception fingerprint to group them all together.

module FingerprintExceptionByClassName
  def raven_context
    { fingerprint: [ self.class.name ] }
  end
end

require 'unicorn'
Unicorn::ClientShutdown.include FingerprintExceptionByClassName
Redis::TimeoutError.include FingerprintExceptionByClassName
Faraday::TimeoutError.include FingerprintExceptionByClassName
Faraday::ConnectionFailed.include FingerprintExceptionByClassName

Faraday::ClientError.class_eval do
  def raven_context
    response = self.response || {}
    { fingerprint: [ self.class.name, response[:status] ],
      tags: { response_status: response[:status] },
      extra: { response_body: response[:body] }
    }
  end
end

ActiveRecord::StatementInvalid.class_eval do
  def raven_context
    case message
    when /\A(.*) mysql_errno=(\d+) sqlstate=(\d+)/
      { fingerprint: [ 'Mysql2', $2.to_i ],
        tags: { mysql_errno: $2.to_i, sqlstate: $3.to_i }
      }
    else
      defined?(super) ? super : {}
    end
  end
end

require 'rmagick'
Magick::ImageMagickError.class_eval do
  def raven_context
    case message
    when /Unsupported marker type/
      { fingerprint: [ "Magick: Unsupported marker type" ] }
    when /\A(.*)`/
      { fingerprint: [ "Magick: #{$1}" ] }
    else
      defined?(super) ? super : {}
    end
  end
end

require 'resque'
Resque::DirtyExit.class_eval do
  def raven_context
    case message
    when /pid (\d+) ([A-Z]+) \(signal (\d+)\)/
      { fingerprint: [ 'Resque::DirtyExit', $3.to_i ],
        tags: { pid: $1.to_i, signame: $2, signo: $3.to_i }
      }
    else
      defined?(super) ? super : {}
    end
  end
end

SignalException.class_eval do
  def raven_context
    { fingerprint: [ 'SignalException', signo ],
      tags: { signo: signo }
    }
  end
end
```

st0012: I think you can replicate the same behavior with an addition to the SDK's `before_send` callback:

```rb
    if exception = hint[:exception]
      exception.raven_context.each do |key, value|
        # here I assume the event would be a Sentry::Event object
        # it'll be a hash if you use the async callback though (which will be removed in version 5.0)
        event.send("#{key}=", value)
      end
    end

    event
  end
```

Just curious, if there're other issues that blocked you from migrating to `sentry-ruby`?

jeremy: Ah, that's a nice approach! We did use `before_send` but didn't consider retaining the existing `#raven_context` methods.

No other blocker issues. The migration was straightforward enough that I wondered whether providing a Raven compatibility layer with deprecations would be feasible & helpful. Most client usage is fairly basic and most API breakage are simple name changes like `Raven.user_context(…)` → `Raven.set_user(…)` and `Raven.configure` → `Sentry.init`.

The only other notable bit is that `config.tags = …` within the `Raven.configure` block was a nice and obvious way to set default tags for all events. That can be done with `Sentry.set_tags(…)` at app boot as well so it's achievable, but it's less obvious and discoverable.

/cc @lewispb

st0012: > I wondered whether providing a Raven compatibility layer with deprecations would be feasible & helpful.

Some APIs actually have subtle behavioral changes, like `Sentry.set_user` doesn't take a block anymore. So I thought an extra layer could be confusing at times. Also as you said, most of them are name changes and shouldn't be hard to our clients, so it's probably not worth the effort for the extra layer.

> The only other notable bit is that config.tags = … within the Raven.configure block was a nice and obvious way to set default tags for all events. 

Do you use the same DSN in multiple projects? Why do you need default tags that are the same in all events?

benoittgt: Following https://github.com/getsentry/sentry-ruby/pull/1101#discussion_r774807097 we still use an enriched Exception object with a method specific for Sentry event's tags like with `sentry_context` and we use it like this

```ruby
    exception = hint[:exception]
    if exception.respond_to?(:sentry_tags_from_issue)
      event.tags = exception.sentry_tags_from_issue
    end

    event
  end

PR #1108, opened 2020-11-20
Support transaction (part 1)

st0012:
This PR builds the groundwork for the performance monitoring feature. It can already send the new "transaction" type data to the server now.

This feature can be enabled by setting:

```ruby
Sentry.init do |config|
  # any float larger than 0.0 to activate the feature
  # but it doesn't do the sampling now and sends every transaction
end
```

Every event happened inside a transaction will now have a new `Trace Details` section:

截圖 2020-11-20 下午3 48 15

---

By clicking `Search By Trace`, users can see all the events associated with that transaction, including the transaction event itself.

截圖 2020-11-20 下午3 48 48

---

And this is how a transaction event looks like:

截圖 2020-11-20 下午3 49 14

### Upcoming Features

- Sampling mechanism
- Different tracing events (currently only supports rack request and active record's SQL query)

### DIFF
--- a/sentry-ruby/lib/sentry-ruby.rb
+++ b/sentry-ruby/lib/sentry-ruby.rb
@@ -3,6 +3,9 @@ require "sentry/core_ext/object/deep_dup"
 require "sentry/configuration"
 require "sentry/logger"
 require "sentry/event"
+require "sentry/transaction_event"
+require "sentry/span"
+require "sentry/transaction"
 require "sentry/hub"
 require "sentry/rack"
 
@@ -60,6 +63,10 @@ module Sentry
       Thread.current[THREAD_LOCAL] || clone_hub_to_current_thread
     end
 
+    def get_transaction
+      get_current_hub.get_transaction
+    end
+
     def clone_hub_to_current_thread
       Thread.current[THREAD_LOCAL] = get_main_hub.clone
     end
@@ -90,6 +97,14 @@ module Sentry
 
     def capture_message(message, **options, &block)
       get_current_hub.capture_message(message, **options, &block)
+    end
+
+    def start_span(**options)
+      get_current_hub.start_span(**options)
+    end
+
+    def start_transaction(**options)
+      get_current_hub.start_transaction(**options)
     end
 
     def last_event_id
--- a/sentry-ruby/lib/sentry-ruby.rb
+++ b/sentry-ruby/lib/sentry-ruby.rb
@@ -63,10 +63,6 @@ module Sentry
       Thread.current[THREAD_LOCAL] || clone_hub_to_current_thread
     end
 
-    def get_transaction
-      get_current_hub.get_transaction
-    end
-
     def clone_hub_to_current_thread
       Thread.current[THREAD_LOCAL] = get_main_hub.clone
     end
@@ -97,10 +93,6 @@ module Sentry
 
     def capture_message(message, **options, &block)
       get_current_hub.capture_message(message, **options, &block)
-    end
-
-    def start_span(**options)
-      get_current_hub.start_span(**options)
     end
 
     def start_transaction(**options)
--- a/sentry-ruby/lib/sentry/hub.rb
+++ b/sentry-ruby/lib/sentry/hub.rb
@@ -67,6 +67,23 @@ module Sentry
       @stack.pop
     end
 
+    def get_transaction
+      current_scope.span
+    end
+
+    def start_span(**options)
+      if span = current_scope.span
+        span.start_child(**options)
+      else
+        Span.new(**options)
+      end
+    end
+
+    def start_transaction(transaction: nil, **options)
+      transaction ||= Transaction.new(**options, sampled: true)
+      # TODO: Add sampling logic
+    end
+
     def capture_exception(exception, **options, &block)
       return unless current_client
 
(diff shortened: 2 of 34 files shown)


sentry-ruby/lib/sentry-ruby.rb
```diff
@@ -60,6 +63,10 @@ def get_current_hub
       Thread.current[THREAD_LOCAL] || clone_hub_to_current_thread
     end
 
+    def get_transaction
+      get_current_hub.get_transaction
+    end
```
HazAT, line 68: I would remove this function, we don't want to have too much surface.
Use `Sentry.get_current_hub.get_scope.get_transaction` instead

sentry-ruby/lib/sentry-ruby.rb
```diff
@@ -92,6 +99,14 @@ def capture_message(message, **options, &block)
       get_current_hub.capture_message(message, **options, &block)
     end
 
+    def start_span(**options)
+      get_current_hub.start_span(**options)
+    end
```
HazAT, line 104: I would remove this function, we don't want to have too much surface.
Also, start_span alone shouldn't exist, this should only work on a transaction with `start_child`

sentry-ruby/lib/sentry/hub.rb
```diff
@@ -67,6 +67,23 @@ def pop_scope
       @stack.pop
     end
 
+    def get_transaction
+      current_scope.span
+    end
```
HazAT, line 72: We have both `get_span` and `get_transaction` on the scope. It could make sense to sometimes get the span to create a tree.
On the other hand. `get_transaction` should always return the root span (using the span recorder to return the first entry) see other SDKs how it's implemented there.

sentry-ruby/lib/sentry-ruby.rb
HazAT: I just check, in JS it's still there but not used so it's deprecated.

sentry-ruby/lib/sentry/hub.rb
```diff
@@ -67,6 +67,23 @@ def pop_scope
...
+
+    def start_span(**options)
+      if span = current_scope.span
+        span.start_child(**options)
+      else
+        Span.new(**options)
+      end
+    end
```
HazAT, line 80: Shouldn't exists anymore, spans are only created as child of transactions or other spans.

sentry-ruby/lib/sentry/span.rb
```diff
@@ -0,0 +1,134 @@
...
+
+module Sentry
+  class Span
+    TRACEPARENT_REGEXP = Regexp.new(
+      "^[ \t]*" +  # whitespace
+      "([0-9a-f]{32})?" +  # trace_id
+      "-?([0-9a-f]{16})?" +  # span_id
+      "-?([01])?" +  # sampled
+      "[ \t]*$"  # whitespace
+    )
```
HazAT, line 12: Should move into `transaction.rb`

sentry-ruby/lib/sentry/span.rb
```diff
@@ -0,0 +1,134 @@
...
+    end
+
+    def finish
+      # already finished
+      return if @timestamp
+
+      @timestamp = Time.now.utc.to_f
+    end
+
+    def to_traceparent
```
HazAT, line 54:
```suggestion
    def to_sentry_trace
```

sentry-ruby/lib/sentry/transaction.rb
```diff
@@ -0,0 +1,34 @@
+module Sentry
+  class Transaction < Span
```
HazAT, line 2: In this function we need a static function called `from_sentry_trace` that gets a header string and creates a Transaction with `traceid` and so on, see: https://github.com/getsentry/sentry-php/pull/1134/files#diff-b778af5cbead054606b23aed4835bd172616ebd9ab923275fe5845e6b45b4b9bR77

HazAT (changes requested):

sentry-ruby/lib/sentry/span.rb
st0012: should this be moved to `Transaction` too?

sentry-ruby/lib/sentry/span.rb
```diff
@@ -0,0 +1,127 @@
...
+      return if @timestamp
+
+      @timestamp = Time.now.utc.to_f
+    end
+
+    def to_sentry_trace
+      sampled_flag = ""
+      sampled_flag = @sampled ? 1 : 0 unless @sampled.nil?
+
+      "#{@trace_id}-#{@span_id}-#{sampled_flag}"
```
HazAT, line 51: I think here is a nit, if `sampled` is nil which is a valid value we would emit `#{@trace_id}-#{@span_id}-` with a trailing `-`

we should change that (eventho I think it would still work)

HazAT (approved):

sentry-ruby/lib/sentry/span.rb
st0012: if `@samped` is `nil`, `sampled_flag` will be an empty string, which will add nothing to the string and become

```
"#{@trace_id}-#{@span_id}-"
```

so exactly the expected behavior 😄 

here's the spec for such case https://github.com/getsentry/sentry-ruby/pull/1108/files#diff-a54e868d59f29d1f5369002784782b73b90c18c6493aab761e0230a51329fb6bR52-R61

PR #1122, opened 2020-12-03
Support event hint

st0012:
This PR adds support for event hints. It's useful to user's custom callbacks like `before_send` or event processors.

### DIFF
--- a/sentry-ruby/spec/sentry/scope_spec.rb
+++ b/sentry-ruby/spec/sentry/scope_spec.rb
@@ -198,14 +198,16 @@ RSpec.describe Sentry::Scope do
     end
 
     it "applies event processors to the event" do
-      subject.add_event_processor do |event|
+      subject.add_event_processor do |event, hint|
         event.tags = { processed: true }
+        event.extra = hint
         event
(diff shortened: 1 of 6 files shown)


sentry-ruby/spec/sentry/scope_spec.rb
```diff
@@ -198,14 +198,16 @@
...
         event.tags = { processed: true }
+        event.extra = hint
         event
       end
 
-      subject.apply_to_event(event)
+      subject.apply_to_event(event, { foo: "bar" })
 
       expect(event.tags).to eq({ processed: true })
+      expect(event.extra).to eq({ foo: "bar" })
```
HazAT, line 210: I guess this is just for testing but the intent of the hint is different ^^

HazAT (approved):

sentry-ruby/spec/sentry/scope_spec.rb
st0012: yeah it's for testing only. I think hints are for users to distinguish events in custom callback/processors?

PR #1126, opened 2020-12-07
Add configuration option for trusted proxies

jeschneiderhan:
## Description

When Sentry calculates the IP address for a user, it will skip addresses that fall in a predetermined list of private IP address ranges: https://github.com/getsentry/sentry-ruby/blob/master/sentry-ruby/lib/sentry/utils/real_ip.rb#L10-L17. Some people run services behind proxies that are not part of these ranges, which currently results in the IP address of the proxy being used as the user IP address.

For example, Let's assume I have a service running behind Cloudflare. When a request from a client (`2.2.2.2`) reaches my application, it has headers like the following:

```
X-Forwarded-For: 2.2.2.2,172.68.54.247     # 172.68.54.247 is an IP address owned by Cloudflare
X-Real-Ip: 172.68.54.247
```

Right now, if the request throws an error, the Sentry will have a user IP address of `172.68.54.247` since it is the first address encountered that is not in the default list of IPs to skip.

ActionDispatch allows the specification of `trusted_proxies` via the `config.action_dispatch.trusted_proxies` setting. Any IP ranges specified will be added to default list of private address ranges (See [remote_ip.rb](https://github.com/rails/rails/blob/fe76a95b0d252a2d7c25e69498b720c96b243ea2/actionpack/lib/action_dispatch/middleware/remote_ip.rb) for more information). I've attempted to provide a similar setting to the `sentry-ruby` configuration. It also updates the `sentry-rails` railtie to automatically add any `config.action_dispatch.trusted_proxies` values to the sentry-ruby `trusted_proxies`.

With these settings, it is possible to add the IP ranges of the non-private proxies that you trust. These IPs will be skipped, and in the example above, will result in Sentries using the correct user IP (`2.2.2.2` in the above example).

I also noticed that the `X-Forwarded-For` header is being iterated from the beginning of the string. Although this works a lot of the time, I think the usual way to process this header is to iterate from right to left, taking the first IP that is not in the trusted list. The [remote_ip](https://github.com/rails/rails/blob/fe76a95b0d252a2d7c25e69498b720c96b243ea2/actionpack/lib/action_dispatch/middleware/remote_ip.rb#L118) code reverses the list prior to processing, for example. I added a `forwarded_ips.reverse` call below to match that behavior.

I hope that all makes sense!


st0012:
@jeschneiderhan thanks for the PR, I think the idea makes sense 👍 
but currently, we're busy preparing for the official release of the new SDKs. so I can't accept new feature changes until we roll out the `4.0.0` version and will put this into the `4.1.0` milestone. I hope that's ok to you?

jeschneiderhan:
@st0012 sounds good! Thanks!

st0012:
@jeschneiderhan sorry for the delay, we have several bug fixes needed to be shipped this week, which contains a few breaking changes. so we cut the `4.1.0` earlier than expected. as a result, this PR has been moved to `4.2.0` 🙇 

I'll review and merge it once we confirm the `4.1.0` doesn't have major issues. in the meantime, can you resolve the conflicts too? I updated the structure of `railtie.rb` so you'll need to adjust the code a bit.

jeschneiderhan:
@st0012 thanks for the update. I've rebased off of master and resolved the conflicts.

sentry-rails/spec/support/test_rails_app/app.rb
```diff
@@ -74,6 +74,7 @@ def self.name
 
   app.config.hosts = nil
   app.config.secret_key_base = "test"
+  app.config.action_dispatch.trusted_proxies = ["5.5.5.5"]
```
st0012, line 77: on the current master you can specify app configuration with `make_basic_app` as well, so I'd prefer to configure this option only when necessary. you should be able to do something like:

```ruby
before do
  make_basic_app do |config, app|
    app.config.action_dispatch.trusted_proxies = ["5.5.5.5"]
  end
end

# the test case
```

sentry-ruby/lib/sentry/utils/real_ip.rb
```diff
@@ -13,21 +13,23 @@ class RealIp
...
         real_ip: nil,
-        forwarded_for: nil
+        forwarded_for: nil,
+        trusted_proxies: []
       )
         @remote_addr = remote_addr
         @client_ip = client_ip
         @real_ip = real_ip
         @forwarded_for = forwarded_for
+        @trusted_proxies = (LOCAL_ADDRESSES + Array(trusted_proxies)).map { |proxy| IPAddr.new(proxy) }
```
st0012, line 32: I'd add a `uniq` here to avoid duplications.

sentry-ruby/lib/sentry/utils/real_ip.rb
```diff
@@ -37,7 +39,7 @@ def calculate_ip
         # Could be a CSV list and/or repeated headers that were concatenated.
         client_ips    = ips_from(@client_ip)
         real_ips      = ips_from(@real_ip)
-        forwarded_ips = ips_from(@forwarded_for)
+        forwarded_ips = ips_from(@forwarded_for).reverse
```
st0012, line 42: can you add a comment on why we need to call `reverse` here?

sentry-ruby/lib/sentry/utils/real_ip.rb
```diff
@@ -63,7 +65,7 @@ def ips_from(header)
       end
 
       def filter_local_addresses(ips)
```
st0012, line 67: this method should be renamed too, as it doesn't only filter local addresses now.

sentry-ruby/spec/sentry/utils/real_ip_spec.rb
```diff
@@ -13,12 +13,13 @@
     subject do
       Sentry::Utils::RealIp.new(
         :forwarded_for => "192.168.0.2, 2.2.2.2, 3.3.3.3, 4.4.4.4",
+        :forwarded_for => "2.2.2.2, 3.3.3.3, 4.4.4.4, 192.168.0.2",
```
st0012, line 16: why do we need these changes here? I think they have something to do with the `reverse` change, but this line looks like a duplicated entry?

st0012 (changes requested):
some small changes needed + some questsions. otherwise looks good 👍

jeschneiderhan:
Thanks for the comments, @st0012. I'm hoping to find time to address them this afternoon.

sentry-ruby/lib/sentry/utils/real_ip.rb
jeschneiderhan: 👍  Comment added. 

From the Mozilla [docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For):
> If a request goes through multiple proxies, the IP addresses of each successive proxy is listed. This means, the right-most IP address is the IP address of the most recent proxy and the left-most IP address is the IP address of the originating client.

When processing these values for a client IP address, one typically starts from the rightmost value (the last proxy) and walks backward until the first IP not in a trusted IP range is found. This is used as the client IP address since nothing before that can be "trusted" to have recorded legitimate values.

The ActiveDispatch [code](https://github.com/rails/rails/blob/fe76a95b0d252a2d7c25e69498b720c96b243ea2/actionpack/lib/action_dispatch/middleware/remote_ip.rb#L118) reverses the list prior to processing, for example.

sentry-ruby/spec/sentry/utils/real_ip_spec.rb
jeschneiderhan: I think this ordering more accurately reflects what you would see in a request - the farther right you go in the list the closer you get to the destination. It makes more sense for the proxy ini a private IP address range to be all the way at the end (since it would have been the last proxy in the chain).

jeschneiderhan:
@st0012 I _think_ I got to everything 👍

sentry-ruby/spec/sentry/utils/real_ip_spec.rb
st0012: I see, thanks for explaining 👍

sentry-ruby/spec/sentry/utils/real_ip_spec.rb
```diff
@@ -50,6 +50,19 @@
...
 
+  context "when custom proxies are provided" do
+    subject do
+      Sentry::Utils::RealIp.new(
+        :forwarded_for => "2.2.2.2, 3.3.3.3, 4.4.4.4",
+        :trusted_proxies => ["4.4.4.4"]
+      )
+    end
+
+    it "should return REMOTE_ADDR" do
```
st0012, line 61: this description should be updated I think?

sentry-ruby/spec/sentry/event_spec.rb
```diff
@@ -107,6 +107,30 @@
           cookies: {}
         )
 
+        expect(event.to_hash[:tags][:request_id]).to eq("abcd-1234-abcd-1234")
+        expect(event.to_hash[:user][:ip_address]).to eq("2.2.2.2")
+      end
+    end
+
+    context "with config.trusted_proxies = [\"2.2.2.2\"]" do
```
st0012, line 115: this block should be placed inside the `"with config.send_default_pii = true"` context.  and then you won't need to set `Sentry.configuration.send_default_pii = true` again in the before block.

st0012 (commented):
I left some comments on updating the test cases. otherwise all look good now 👍

sentry-ruby/spec/sentry/utils/real_ip_spec.rb
jeschneiderhan: Yes, sir!

sentry-ruby/spec/sentry/event_spec.rb
jeschneiderhan: I _think_ I understood. Let me know if I was mistaken 👍

st0012 (approved):
@jeschneiderhan thanks for the great work ❤️  I'll resolve the conflicts 😉

jeschneiderhan:
I think the check that failed may have been a transient error. Can we re-run it?

st0012:
@jeschneiderhan the failure is actually inherited from the `4-2` branch, so merging now.

PR #1132, opened 2020-12-10
ActiveSupportLogger should only record events that has a started time

### LINKED ISSUE #1131: Sentry 4.0 incompatibility with `ActiveSupport::Notifications.publish`
**Describe the bug**

When testing our upgrade to Sentry 4.0, we ran into an issue with another dependency calling `ActiveSupport::Notifications.publish` and Sentry's `ActiveSupportLogger` trying to call `to_i` on a non-`Time` object. In this case, it is Active Admin's [`load!` method](https://github.com/activeadmin/activeadmin/blob/v2.8.1/lib/active_admin/application.rb#L110-L120), though I'd imagine any use of `ActiveSupport::Notifications.publish` will cause a similar issue.

**To Reproduce**

Call `ActiveSupport::Notifications.publish`.

**Expected behavior**

`ActiveSupportLogger` should not raise an exception.

**Actual behavior**

`ActiveSupportLogger` raises an `undefined method 'to_i'` exception.

**Environment**
 - Ruby Version: 2.7.2
 - SDK Version: 4.0.0
  - Rails 6.0.3.4
  - Active Admin 2.8.1

**Raven Config**

This is not necessary but could be helpful.

```ruby
Sentry.init do |config|
  # Send Sentry notifications asynchronously by default
  config.async = ->(event) { SentryJob.perform_later(event) }

  config.breadcrumbs_logger = [:active_support_logger]
end
```

st0012:
Because `ActiveSupport::Notifications.publish` can be used to publish anything, whether it's a standard Rails instrumentation payload, a string, or even a custom object, it's dangerous for the logger to accept anything that's being published.

So this PR filters out events that are not compatible with the default Rails instrumentation format to avoid issues.

closes #1131

### DIFF
--- a/sentry-rails/lib/sentry/rails/breadcrumb/active_support_logger.rb
+++ b/sentry-rails/lib/sentry/rails/breadcrumb/active_support_logger.rb
@@ -14,7 +14,10 @@ module Sentry
 
           def inject
             @subscriber = ::ActiveSupport::Notifications.subscribe(/.*/) do |name, started, finished, unique_id, data|
-              add(name, started, finished, unique_id, data)
+              # we only record events that has a started timestamp
+              if started.is_a?(Time)
+                add(name, started, finished, unique_id, data)
+              end
             end
           end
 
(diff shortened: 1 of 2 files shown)


st0012:
fyi the fix has been released in `4.0.1`

PR #1155, opened 2020-12-17
Implement non-blocking event sending

### LINKED ISSUE #1130: Internal Transport queue to limit backpressure
We should have an internal transport queue that at one point stops accepting events
e.g. on JS we have a queue that in max can hold 30 events in queue that should be sent to Sentry.
If the queue is full because we are sending so much the transport is not accepting anything anymore.

Python
https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/transport.py#L130

st0012:
For users who don't use `config.async`, sending events is currently a blocking operation. This means a request won't complete before some SDK operations like serializing or making requests to Sentry are also finished. So this PR introduces non-blocking event sending functionality to the SDK.

The functionality works like this: 
1. When the SDK is initialized, a `Sentry::BackgroundWorker` will be initialized too.
2. When an event is passed to `Client#capture_event`, instead of sending it directly with `Client#send_event`, we'll let the worker do it.
3. The worker will have a number of threads. And one of the idle threads will pick the job and call `Client#send_event`.
  - If all the threads are busy, new jobs will be put into a queue, which has a limit of 30.
  - If the queue size is exceeded, new events will be dropped.

### About Sentry::BackgroundWorker

- The worker is built on top of the [concurrent-ruby](https://github.com/ruby-concurrency/concurrent-ruby) gem's [ThreadPoolExecutor](http://ruby-concurrency.github.io/concurrent-ruby/master/Concurrent/ThreadPoolExecutor.html), which is also used by Rails ActiveJob's async adapter. This should minimize the risk of messing up client applications with our own thread pool implementation.
- It respects users' async implementation. So if the users have `config.async` set, the worker won't initialize a thread pool and won't be used either.
- The worker will be initialized at the end of the `Sentry.init` call.

#### New Configuration Option - `background_worker_threads`

This change also introduces a new `background_worker_threads` config option. It allows users to decide how many threads should the worker hold. By default, the value will be the number of the processors the user's machine has. For example, if the user's machine has 4 processors, the value would be 4.

Of course, users can always override the value to fit their use cases, like

```ruby
config.background_worker_threads = 5 # the worker will have 5 threads for sending events
```

Users can also disable this new non-blocking behaviour by giving a `0` value:

```ruby
config.background_worker_threads = 0 # all events will be sent synchronously
```

closes #1130

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,10 @@ The official Ruby-language client and integration layer for the [Sentry](https:/
 
 We test on Ruby 2.4, 2.5, 2.6 and 2.7 at the latest patchlevel/teeny version. We also support JRuby 9.0.
 
+## Migrate From sentry-raven
+
+If you're using `sentry-raven`, we recommend you to migrate to this new SDK. You can find the benefits of migrating and how to do it in our [migration guide](https://docs.sentry.io/platforms/ruby/migration/).
+
 ## Getting Started
 
 ### Install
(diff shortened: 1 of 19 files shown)


HazAT (approved):
LGTM, I'll add @bruno-garcia for a second pair of eyes

bruno-garcia (approved):
LGTM. Very nice and simple

PR #1159, opened 2020-12-18
Use stricter dependency declaration

st0012:


agrobbin:
Nice, thanks @st0012!

Issue #1173, opened 2020-12-28
sentry-sidekiq gem swallows exceptions, preventing worker retries


jackrr:
**Describe the bug**

After installing the `sentry-sidekiq` gem, jobs that raise exceptions are not added to the retry queue.

**To Reproduce**

Add the `sentry-sidekiq` gem to a runtime using `sidekiq`. Raise some exception in your worker's `perform` method.

**Expected behavior**

The raised exception appears in sentry and the job is added to the retry queue.

**Actual behavior**

The raised exception appears in sentry, but the job is considered "processed" and not retried.

**Environment**
 - Ruby Version: 2.7.2
 - SDK Version: 4.1.1
 - Integration Versions (if any):
  - Rails 6.0.3.4
  - Sidekiq 6.1.2

**Sentry Config**

```ruby
Sentry.init do |config|
  config.dsn = ENV['SENTRY_DSN']
  config.breadcrumbs_logger = [:active_support_logger, :sentry_logger]
  config.environment = ENV.fetch('SENTRY_ENV', Rails.env)
  config.release = ENV['BUILD_VERSION']

  # To activate performance monitoring, set one of these options.
  # We recommend adjusting the value in production:
  # config.traces_sample_rate = 0.5
  # or
  # config.traces_sampler = lambda do |context|
  #   true
  # end
end
```

I've gotten around the issue by monkey-patching `Sentry::Sidekiq::CleanupMiddleware` so that it re-raises the caught exception after capturing it on sentry as follows:

```ruby
module Sentry
  module Sidekiq
    class CleanupMiddleware
      def call(_worker, job, queue)
        return yield unless Sentry.initialized?

        Sentry.clone_hub_to_current_thread
        Sentry.with_scope do |scope|
          scope.set_extras(sidekiq: job.merge("queue" => queue))
          scope.set_transaction_name("Sidekiq/#{job["class"]}")

          begin
            yield
          rescue => ex
            Sentry.capture_exception(ex, hint: { background: false })
            # THIS IS THE PATCH
            raise ex
            # THIS IS THE PATCH
          end
        end
      end
    end
  end
end
```

I can open a PR to include the change if that's the right way forward, but wanted to double check here first to make sure there isn't something else I'm missing (such as a configuration flag or other "better practice" way to achieve my desired behavior).

st0012:
@jackrr sorry, that's a bug. I think your fix is correct though 👍

jackrr:
[...]

PR #1174, opened 2020-12-29
fix: retry jobs that raise exceptions

jackrr:

**Other Notes**

## Description
Describe your changes:

CleanupMiddleware was swallowing exceptions raised in jobs, preventing
Sidekiq from adding the job to the retry queue. Re-raising from the
CleanupMiddleware led to a second firing from the ErrorHandler
class. To get around this, this commit narrows the scope of
ErrorHandler to only handle Sidekiq lifecycle events, and renames the
ErrorHandler class to EventErrorHandler accordingly.

Also fixes (I think) a second issue, where active job context key
renaming was not being applied on the CleanupMiddleware class, which
was the class responsible for firing job error events. Ensures that
the context filter is called in the CleanupMiddleware as well.

Addresses https://github.com/getsentry/sentry-ruby/issues/1173

### DIFF
--- a/sentry-sidekiq/lib/sentry-sidekiq.rb
+++ b/sentry-sidekiq/lib/sentry-sidekiq.rb
@@ -1,7 +1,7 @@
 require "sidekiq"
 require "sentry-ruby"
 require "sentry/sidekiq/version"
-require "sentry/sidekiq/error_handler"
+require "sentry/sidekiq/event_error_handler"
 require "sentry/sidekiq/cleanup_middleware"
 # require "sentry/sidekiq/configuration"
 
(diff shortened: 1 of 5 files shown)


st0012:
thanks for the PR!
but I think there's a fundamental issue with the current usages of middlewares/error handlers (my fault). after taking a look at the old SDK and sidekiq's documents, I think `CleanupMiddleware` shouldn't catch exceptions at all. since we also register an error handler, all the error reporting should be done by it alone. I will make a structural change on it in another PR as a demonstration.

jackrr:
@st0012 Yeah, I had a feeling about this solution being undesirable, but was having some issues getting the shared scope context to work as expected. Glad to see you have it working, much better looking end result in your PR. I've opened a PR to your branch with a regression test to prove the bug is fixed: https://github.com/getsentry/sentry-ruby/pull/1176

PR #1175, opened 2020-12-29
Fix sidekiq middleware

### LINKED ISSUE #1173: sentry-sidekiq gem swallows exceptions, preventing worker retries

st0012:
The changes are mainly around the middleware part. Here's the commit message:

```
Since we already register the ErrorHandler, the middleware shouldn't be
responsible for reporting errors anymore.

Its responsibilities are:
1. Clone the main hub to the current thread. This will make sure Sentry
   is accessible from inside a job so the users can add custom contexts.
2. Add some default context data to the scope, like the transaction name
   and queue name.
3. Clear the scope after the job is processed successfully.
```

Other than that are spec restructuring and small improvements on the `ErrorHandler`.

This should fix #1173


st0012:
@jackrr I believe this can solve #1173 as well, would you mind giving it a try too? 
your fix is great but I think it's built around the currently incorrect usages of sidekiq middleware. however, this is just my limited understanding of sidekiq. so I'd love to hear your opinion on this as well 😄

jackrr:
@st0012 looks great! Thanks for responding so quickly. I've added a regression test to prove the fix for https://github.com/getsentry/sentry-ruby/issues/1173 here: https://github.com/getsentry/sentry-ruby/pull/1176

st0012:
@jackrr np, thanks for reporting the issue + all the feedback. that's why I can fix it rather quickly 🙂

PR #1176, opened 2020-12-29
test: add regression test for sidekiq retries

jackrr:
Proves fix for https://github.com/getsentry/sentry-ruby/issues/1173


st0012:
@jackrr awesome! thank you 👍

PR #1189, opened 2021-01-08
Supply event hint to async callback when possible

### LINKED ISSUE #1188: Support event hint in async
Currently, the SDK only passes `event` to the `async` block. So users who use `async` will not have `hint` in their `before_send` block. We should pass `hint` to the `async` block as well so those users can take advantage of the event hint too.

However, adding a new argument will break apps that use lambda for the `async` option.

st0012:
This fixes #1188.

### DIFF
--- a/sentry-ruby/spec/sentry/client_spec.rb
+++ b/sentry-ruby/spec/sentry/client_spec.rb
@@ -29,23 +29,45 @@ RSpec.describe Sentry::Client do
     let(:event) { subject.event_from_message(message) }
 
     context 'with config.async set' do
+      let(:async_block) do
+        lambda do |event|
+          subject.send_event(event)
+        end
+      end
+
       around do |example|
         prior_async = configuration.async
-        configuration.async = proc { :ok }
+        configuration.async = async_block
         example.run
         configuration.async = prior_async
       end
 
-      before do
-        allow(subject).to receive(:send_data)
-      end
-
-      it "doesn't send the event right away" do
-        expect(configuration.async).to receive(:call)
+      it "executes the given block" do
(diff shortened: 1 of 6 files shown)


sentry-ruby/spec/sentry/client_spec.rb
```diff
@@ -29,23 +29,45 @@ def sentry_context
...
         returned = subject.capture_event(event, scope)
 
         expect(returned).to be_a(Sentry::Event)
+        expect(subject.transport.events.first).to eq(event.to_json_compatible)
+      end
+
+      context "with 2 arity block" do
+        let(:async_block) do
+          lambda do |event, hint|
+            event["tags"]["hint"] = hint
```
benoittgt, line 57: I think that something that should maybe be documented. We jump from event object to and hash in async scenario with for example `event.server_name` to `event['server_name']` for example.

I had the surprise yesterday. Maybe add a small notice when speaking about `async` in README.

Also the second args should be documented? Maybe you planned to do it in another PR 😊

st0012: >  I think that something that should maybe be documented. 

yeah I agree with this. thanks for reminding 😄 

> Also the second args should be documented? 

if we consider "not passing event hint" as a bug, this change just fixes it to match the users' expectation. so I don't think it's something that worth document for. unless you're talking about updating the examples? then yes, I'll update them as well.

benoittgt: Yes. More updating the example :)

benoittgt:
Thanks!

Issue #777, opened 2017-10-12
extra_context ignored in rack / grape app


travisp:
We're using Grape API (https://github.com/ruby-grape/grape) inside of Rails and setup sentry for our apps. We followed the setup here: https://github.com/getsentry/raven-ruby/issues/373

In a grape before block we have,
```
    before do
      Raven.extra_context(
        params: params.to_hash,
        url: request.url,
        declared_client: request.headers['User-Agent']
      )
    end
```

However when we `capture_exception` in grape for errors that occur (we do this because we gracefully handle many exceptions), this extra context never gets passed to Sentry. I'm unsure if this is a bug in Sentry, or a problem with what we're doing

nateberkopec:
I don't know how to use Grape. Can you please modify [this script](https://gist.github.com/nateberkopec/030581ead896bbb8f5ccbc5b961929c7) to get a reproduction?

travisp:
@nateberkopec here you go https://gist.github.com/travisp/044b9e296229f90f5bc3d274414db45b

The problem seems to be that Raven::Rack.call actually captures the exception before Grape's `rescue_all` (triggering it more than expected -- which I missed previously), and then calls `Context.clear!` when it re-raises the exception and sends it to `rescue_all`.

The same issue happens if I don't disable the automatic integrations, don't use `use Raven::Rack` and reliable on the rails integration.

The desired behavior would be to capture the exception in Grape _first_, which would also let us decide not to re-raise the exception to Raven::Rack if we wanted. I'm not sure where the appropriate place to get this working is, or if there's a change I can make myself.

nateberkopec:
I think this test is invalid:

```ruby
def test_has_middleware
    assert_includes app.middleware, Raven::Rack
  end
```

App in this context is the Rails application. You added the middleware to the Grape rack app, not the Rails one, so of course this test will fail.

I think something's wrong with your Grape app as well, because:

```ruby
def test_hello
    get "/hello"
    assert last_response.ok?
  end
```

...also fails.

travisp:
That was the setup in https://github.com/getsentry/raven-ruby/issues/373 but I'll update the test

travisp:
[...]

PR #1197, opened 2021-01-12
Fix headers serialization for sentry-ruby

moofkit:
## Description
There is performance issue with `sentry-raven` and `sentry-ruby` gems.

In some cases different middlewares assign to `env` hash some large objects.
I.e. [grape](https://github.com/ruby-grape/grape) expose his routing in `env['grape.routing_args']` https://github.com/ruby-grape/grape/blob/02d7113d09eb9fcb4264c841d1fdd305e3e8adb5/lib/grape/util/env.rb#L22

In large applications it may be havy hash object. So when `sentry` try to cast this with `#to_s` that causes to large CPU consumption. I.e. in our application it takes around 45 sec to perform and almost 100% CPU load).
And the result just throwed away because this is not headers that needs to be in the event.

There are some demo that emulates this issue https://gist.github.com/moofkit/e1bf1ca0d3186aa7f9b89d3c3ee7fc58
You can try it with
```
$ make start
$ make test
> time curl 'http://localhost:9292/exception'
{"error":"An error ocurred"}       10.11 real         0.00 user         0.00 sys
```

Why it takes 10 secs of 5 is covered in this issue https://github.com/getsentry/sentry-ruby/issues/777

So, this little patch postpone `Object#to_s` after all headers checks in both `sentry-raven` and `sentry-ruby` gems

### DIFF
--- a/sentry-ruby/lib/sentry/interfaces/request.rb
+++ b/sentry-ruby/lib/sentry/interfaces/request.rb
@@ -68,7 +68,6 @@ module Sentry
       env.each_with_object({}) do |(key, value), memo|
         begin
           key = key.to_s # rack env can contain symbols
-          value = value.to_s
           next memo['X-Request-Id'] ||= Utils::RequestId.read_from(env) if Utils::RequestId::REQUEST_ID_HEADERS.include?(key)
           next if is_server_protocol?(key, value, env["SERVER_PROTOCOL"])
           next if is_skippable_header?(key)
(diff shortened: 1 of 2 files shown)


st0012:
@moofkit thanks for pointing this out & the PR!  I'd love to merge it right away, but we're managing the release of these 2 gems separately. so can you split this PR into 2?

moofkit:
@st0012 sure

moofkit:
@st0012 thanks for quick response!

I've split this in two PR.
This PR is for raven https://github.com/getsentry/sentry-ruby/pull/1198

st0012:
@moofkit thanks 👍 I'll give them a look again tomorrow and merge them

st0012 (approved):

st0012:
@moofkit great work, thanks 👍

smcgivern:
Thanks so much for this @moofkit, we also hit this issue and it was great to see such a clear PR description to help us understand it was the same problem: https://gitlab.com/gitlab-org/gitlab/-/issues/330729

PR #1198, opened 2021-01-12
fix format headers for sentry-raven

moofkit:
## Description
Postpone headers serialization in `sentry-raven`
More details in PR: https://github.com/getsentry/sentry-ruby/pull/1197

### DIFF
--- a/sentry-raven/lib/raven/integrations/rack.rb
+++ b/sentry-raven/lib/raven/integrations/rack.rb
@@ -100,7 +100,6 @@ module Raven
       env_hash.each_with_object({}) do |(key, value), memo|
         begin
           key = key.to_s # rack env can contain symbols
-          value = value.to_s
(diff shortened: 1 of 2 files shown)


moofkit:
it seems like this spec does not cover the case https://github.com/getsentry/sentry-ruby/blob/master/sentry-raven/spec/raven/integrations/rack_spec.rb#L163
Because from now we trying to cast only values with `http-` prefix.

Should I change a spec that enforces this case with addin `http-` prefix?
Which seems to me something unlikely to put object in http header

Or we don't need this `rescue` code anymore?

st0012:
@moofkit don't worry about the test coverage drop. It's unlikely that we'll do any drastic change to the `sentry-raven` gem anyway 🙂

st0012 (approved):

PR #1199, opened 2021-01-13
Don't include headers & request info in tracing span or breadcrumb

st0012:
Data like `request` or `headers` can be huge in some applications. They're also likely to contain complex data structure that's hard/impossible to serialize (some might cause infinite loop).

Consider those information should have been collected in the `request` interface of the event, storing them in breadcrumbs could cause more harm than benefit.


leonid-shevtsov:
This is not enough, you need to also remove `response` - in particular because it contains a reference to `request` - and many other nasty objects too.

st0012:
@leonid-shevtsov I'm sorry that I missed this comment 🙇 I've added https://github.com/getsentry/sentry-ruby/pull/1463 to fix the issue.

PR #1205, opened 2021-01-14
Support capturing "sentry-trace" header from the middleware

### LINKED ISSUE #1200: Pick up `sentry-trace` header on Rails/Rack
On incoming requests, we should pick up the `sentry-trace` header and derive the transaction from it.
Like here in Laravel
https://github.com/getsentry/sentry-laravel/blob/ee3434f62981faabb1803e5ed57e76bc843dca02/src/Sentry/Laravel/Tracing/Middleware.php#L87

If we don't do this, we can't connect frontend -> backend.

st0012:
When `sentry-trace` is provided in the header, we should start the transaction with the information from the trace, including:

1. trace_id
2. parent_span_id
3. sampled

Closes #1200

### DIFF
--- a/sentry-ruby/CHANGELOG.md
+++ b/sentry-ruby/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Support capturing "sentry-trace" header from the middleware [#1205](https://github.com/getsentry/sentry-ruby/pull/1205)
+
 ## 4.1.3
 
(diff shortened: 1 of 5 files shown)


taylorthurlow:
@st0012 Is this header something that's used internally and not intended to be part of the public API? Sending a request to a rack app with a `sentry-trace` HTTP header is accessible in the rack env with `env["HTTP_SENTRY_TRACE"]`. Am I misunderstanding the use case for this feature?

st0012:
@taylorthurlow ah this you're right 🤦‍♂️ will address this asap!

PR #1209, opened 2021-01-15
Check the argument type of capture_exception and capture_event helpers

st0012:


benoittgt (approved):
Thanks for this addition! It's nice to have is in a `module` <3

sentry-ruby/spec/sentry/hub_spec.rb
```diff
@@ -151,6 +151,12 @@
       end.to change { transport.events.count }.by(1)
     end
 
+    it "raises error when passing a non-exception object" do
+      expect do
+        subject.capture_exception("String")
+      end.to raise_error(ArgumentError)
```
benoittgt, line 157: Maybe you could match the exception message too `raise_error(ArgumentError, "expect the argument")`?

PR #1212, opened 2021-01-17
Fix broken link in README

tricknotes:
The "update-readme" branch has been removed.
It seems "master" branch is correct for here.


st0012 (approved):
thanks!

PR #1213, opened 2021-01-17
Inline global method

tricknotes:
It is a good practice to avoid global method definition.
When method is defined in globally, this method can be called from outside of gem.
It may affect other gems that have the same named method.


st0012 (approved):
yeah I think your approach is much better, thanks for the suggestion 👍

tricknotes:
Thanks for your quick review too!

PR #1215, opened 2021-01-19
Add `ActionDispatch::Http::MimeNegotiation::InvalidType` to the list of default ignored Rails exceptions

agrobbin:
`ActionDispatch::Http::MimeNegotiation::InvalidType` was introduced in rails/rails#40353 to allow services like Sentry to more specifically target this situation and ignore the exception.

It will be introduced in the next version of Rails 6.0.x, and is already released in 6.1.


agrobbin:
I'm pretty sure the errors showing up in CI are not related to my change.

st0012 (approved):
thanks for the change 👍  I've re-run the CI and it's passing now.

agrobbin:
Thanks @st0012!

PR #1223, opened 2021-01-21
Copy request env before Rails' ShowExceptions middleware

### LINKED ISSUE #1219: Rails / Rack integration reporting incorrect URL (500 page)
**Describe the bug**
When the rack middleware catches an unhandled exception in production, the reported URL relative path is always `/500`

In development, it works as expected with the correct URL.

This seems to be a regression since the last major version upgrade.

Screen Shot 2021-01-20 at 15 48 07

**To Reproduce**

Example repo: https://github.com/mcclymont/rails-sentry-url-bug

* Create a new blank rails project with `rails new my_api --api`
* Include the gems 'sentry-ruby' and 'sentry-rails' in the Gemfile
* Create config/initializers/sentry.rb
* Create a basic route and controller with route e.g. GET /testing.
* In the controller action, raise an error intentionally e.g. `raise 'testing sentry'`
* Navigate to /testing

**Expected behavior**
 The exception is raised and reported to Sentry with the correct request URL path: `http://localhost:3000/testing`

**Actual behavior**
When running with RAILS_ENV=production, the reported URL path is `http://localhost:3000/500`

**Environment**
 - Ruby Version: 2.7.1
 - SDK Version: sentry-ruby (4.1.4), sentry-rails (4.1.4)
 - Integration Versions (if any): Rails 6.0.3.4

**Raven Config**

```ruby
Sentry.init do |config|
  config.dsn = ENV.fetch('SENTRY_DSN')
  config.environment = 'dev'

  config.before_send = lambda do |event, _hint|
    puts event.request&.url
    event
  end
end
```

st0012:
When Rails' ShowExceptions middleware calls `#render_exception`, it
    also updates some of the env's headers, like changing `PATH_INFO` to
    `/500`.

    And if the current scope holds the same env hash as ShowExceptions does, this
    behavior will affect the reported event as well. Because
    the CaptureExceptions middleware will always run after ShowExceptions
    is called.

    So the solution to copy current scope's rack env when
    we know the exceptions app is going to be rendered.

    Notes:
    - sentry-raven patches ShowExceptions and reports events before
      `ShowExceptions#render_exception` is called, so it doesn't have such issue.
    - `ShowExceptions#render_exception` only changes first-level env
      attributes. So a simple `dup` should be able to solve the issue, don't
      need to deep-copy the entire env hash.

Closes #1219


mcclymont:
Thanks @st0012 !

PR #1229, opened 2021-01-22
Fix `metadata["changelog_uri"]` in gemspecs

tricknotes:
## Description
All gems have changelog under its own directories.

### DIFF
--- a/sentry-rails/sentry-rails.gemspec
+++ b/sentry-rails/sentry-rails.gemspec
@@ -16,7 +16,7 @@ Gem::Specification.new do |spec|
 
   spec.metadata["homepage_uri"] = spec.homepage
   spec.metadata["source_code_uri"] = spec.homepage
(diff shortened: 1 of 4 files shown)


sentry-rails/sentry-rails.gemspec
```diff
@@ -16,7 +16,7 @@ Gem::Specification.new do |spec|
 
   spec.metadata["homepage_uri"] = spec.homepage
   spec.metadata["source_code_uri"] = spec.homepage
-  spec.metadata["changelog_uri"] = "#{spec.homepage}/blob/master/CHANGELOG.md"
+  spec.metadata["changelog_uri"] = "#{spec.homepage}/blob/master/sentry-ruby/CHANGELOG.md"
```
st0012, line 19: shouldn't this be `sentry-rails`?

tricknotes: Ouch! I fixed and rebased.

st0012 (approved):
thanks for the fix 👍

PR #1231, opened 2021-01-25
Serialize event hint before passing it to the async block

### LINKED ISSUE #1227: Event is not compatible with ActiveJob 6+
**Describe the bug**

Trying to use `Sentry.capture_exception(error)` to capture  a StandardError and send it asynchronously ,  but it failed to enqueue and reported 

> async event sending failed: Unsupported argument type: StandardError. 

The failed reason can be found here: <https://github.com/rails/rails/blob/v6.0.3.4/activejob/lib/active_job/serializers.rb#L28>. It seems like ActiveJob 6 can't serialize an Exception object

**Expected behavior**
Event should be compatible with ActiveJob 6+

**Environment**
 - Ruby Version: 2.7.2
 - SDK Version: 4.1.4
  - Rails 6.0.3.4, Sidekiq 6.1.2

**Sentry-ruby Config**

```ruby
    Sentry.init do |config|
      config.breadcrumbs_logger = [:active_support_logger]
        event.modules.clear
        event
      end
      config.async = lambda { |event, hint| SentryJob.perform_later(event, hint) }
    end
```

st0012:
Event hint should also be serialized into a json-compatible hash to make sure ActiveJob can process it correctly.

Closes #1227

### DIFF
--- a/sentry-rails/examples/rails-6.0/config/application.rb
+++ b/sentry-rails/examples/rails-6.0/config/application.rb
@@ -19,5 +19,7 @@ module Rails60
 
     # https://github.com/getsentry/raven-ruby/issues/494
     config.exceptions_app = self.routes
+
+    config.webpacker.check_yarn_integrity = false
   end
 end
--- a/sentry-ruby/lib/sentry/client.rb
+++ b/sentry-ruby/lib/sentry/client.rb
@@ -32,6 +32,7 @@ module Sentry
           event_hash = event.to_json_compatible
 
           if async_block.arity == 2
+            hint = JSON.parse(JSON.generate(hint))
             async_block.call(event_hash, hint)
           else
             async_block.call(event_hash)
--- a/sentry-ruby/spec/sentry/client_spec.rb
+++ b/sentry-ruby/spec/sentry/client_spec.rb
@@ -80,14 +80,14 @@ RSpec.describe Sentry::Client do
           end
         end
 
-        it "supplies hint as the second argument" do
+        it "serializes hint and supplies it as the second argument" do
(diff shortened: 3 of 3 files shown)


mikewindow:
Can we please merge this?

I received an insensitive upgrade email about food rotting during a pandemic when we're having trouble obtaining and affording food. Then the new gems don't work with Rails 6.

agrobbin:
@st0012 we just ran into an issue that I believe was caused by this. Serializing the `hint` causes the exception object to no longer be accessible to the `async` block, as it's been converted to JSON. We have a (temporary) check in our `async` block that needs to actually inspect the `Exception#backtrace`, and it's no longer accessible via `hint[:exception]` (which is what it was pre-JSON serialization).

I understand wanting to provide a serialized value that can be easily passed to `SendEventJob` (or some other custom Active Job-like class), so I'm wondering if there's another way for those who need to access the exception itself to do so?

st0012:
you can access the backtrace through the event payload like:

```ruby
    if hint["exception"]
      exception = event["exception"]["values"].first
      stacktrace = exception["stacktrace"]["frames"]
      # which looks like this
      #
      # [
      #   {:project_root=>"/Users/st0012/projects/sentry-ruby/sentry-ruby", :abs_path=>"/some/other/path", :function=>"other_function", :lineno=>1412, :in_app=>false, :filename=>"/some/other/path"},
      #   {:project_root=>"/Users/st0012/projects/sentry-ruby/sentry-ruby", :abs_path=>"/path/to/some/file", :function=>"function_name", :lineno=>22, :in_app=>false, :filename=>"/path/to/some/file"}
      # ]
    end
```

PR #1235, opened 2021-01-25
Allow `Sentry.init` without block argument

sue445:
## Description
When `SENTRY_DSN` has been passed from environment variable, `config.dsn` in `Sentry.init` is needless.

But simply call  `Sentry.init` will result in an error

e.g.

```ruby
# example.rb
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "sentry-ruby", "4.1.4"
end

require "sentry-ruby"

Sentry.init
```

```bash
$ SENTRY_DSN=xxxxxx ruby example.rb

/Users/sue445/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/sentry-ruby-4.1.4/lib/sentry-ruby.rb:67:in `init': no block given (yield) (LocalJumpError)
	from example.rb:10:in `<main>'
```

The workaround is below, but it's a bit dirty...

```ruby
Sentry.init do
  # nop
end

# or

Sentry.init {  }
```

So I made it possible to call `Sentry.init` without block

### DIFF
--- a/sentry-ruby/lib/sentry-ruby.rb
+++ b/sentry-ruby/lib/sentry-ruby.rb
@@ -65,7 +65,7 @@ module Sentry
 
     def init(&block)
       config = Configuration.new
-      yield(config)
+      yield(config) if block_given?
       client = Client.new(config)
       scope = Scope.new
       hub = Hub.new(client, scope)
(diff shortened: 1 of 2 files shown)


st0012 (approved):
Thanks for the change 👍

PR #1238, opened 2021-01-26
Extract wrapped Sidekiq job class names in both Sidekiq integrations

### LINKED ISSUE #1237: Active Job + Sidekiq transaction name has reverted back to JobWrapper
**Describe the bug**

I'm not sure when this started happening, maybe with the launch of the v4 SDK, but we have started seeing `Sidekiq/ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper` as the transaction name rather than the wrapped job class name for errors within Active Job/Sidekiq. It seems this is a new incarnation of #702.

**To Reproduce**

```ruby
class FooJob < ApplicationJob
  def perform
    raise 'bar'
  end
end
```

**Expected behavior**

`FooJob` is used as the transaction name.

**Actual behavior**

`Sidekiq/ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper` is used as the transaction name.

**Environment**
 - Ruby Version: 2.7.2
 - SDK Version: 4.1.4
 - Integration Versions:
  - Rails 6.0.3.4
  - Sidekiq 6.1.3

**Sentry Config**

```ruby
Sentry.init do |config|
  # Send Sentry notifications asynchronously by default
  config.async = ->(event) { SentryJob.perform_later(event) }
end
```

agrobbin:
The Active Job `wrapped` job class name logic was only happening in the lower-level `ErrorHandler`, and was actually being ignored since `transaction_name` was already set via `SentryContextMiddleware`.

Now, the logic to properly extract the `transaction_name` has been refactored into 1 `Util` class that is used by `SentryContextMiddleware` and `ErrorHandler`.

Separately, this removes the now-unused `CleanupMiddleware`, after its replacement in #1177.

Fixes #1237.


agrobbin:
I'm not sure why project-level code coverage dropped so significantly, @st0012 let me know if I have something weird going on here that is causing that!

st0012:
@agrobbin thanks for the PR 👍 it looks good but I'll do some testing to see why I was not able to spot the issue before merging this, which will probably be this Friday 🙂

sentry-sidekiq/spec/sentry/sidekiq/util_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      it 'extracts the class' do
+        expect(transaction_name).to eq('Sidekiq/FooJob')
+      end
+    end
+
+    context 'when the context is a wrapped job' do
+      let(:context) { { 'wrapped' => 'FooJob', 'class' => 'WrapperJob' } }
+
+      it 'extracts the class' do
```
st0012, line 18: it'd be better if the case description can be more clear. like `"extracts the wrapped job's name"`.

sentry-sidekiq/spec/sentry/sidekiq/util_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+        expect(transaction_name).to eq('Sidekiq/startup')
+      end
+    end
+
+    context 'when the context is for something else' do
+      let(:context) { { foo: 'bar' } }
+
+      it 'extracts the class' do
```
st0012, line 50: same as above

sentry-sidekiq/spec/sentry/sidekiq/util_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    context 'when the context is a wrapped nested job' do
+      let(:context) { { job: { 'wrapped' => 'FooJob', 'class' => 'WrapperJob' } } }
+
+      it 'extracts the class' do
+        expect(transaction_name).to eq('Sidekiq/FooJob')
+      end
+    end
+
+    context 'when the context is an event' do
+      let(:context) { { event: 'startup' } }
```
st0012, line 40: I'm not too familiar with sidekiq, can `event` and other attributes like `wrapped`, `class` appears at the same time?

sentry-sidekiq/lib/sentry/sidekiq/util.rb
```diff
@@ -0,0 +1,21 @@
+module Sentry
+  module Sidekiq
+    class Util
+      SIDEKIQ_NAME = "Sidekiq".freeze
+
+      def self.transaction_name_from_context(context)
```
st0012, line 6: I'm thinking maybe this can be moved into `ContextFilter` as well? it seems reasonable to be able to extract the transaction name as well. wdyt?

st0012 (commented):
thanks for reporting the issue & the PR 😄 

I can confirm this fixes the issue
截圖 2021-01-29 下午6 25 05

just left a few questions + small suggestions. I think we can merge it today 👍

sentry-sidekiq/spec/sentry/sidekiq/util_spec.rb
agrobbin: My understanding, based on the [Sidekiq `fire_event` source](https://github.com/mperham/sidekiq/blob/02d265aac059644d42550677fdea77a7874f27c4/lib/sidekiq/util.rb#L52-L65), is that these 2 things cannot coexist.

agrobbin:
@st0012 just pushed an update here, let me know if there's anything else I can do!

st0012 (approved):
thanks for the great work 👍

agrobbin:
Thanks, @st0012! Looking forward to getting this released and into production.

PR #1240, opened 2021-01-27
CI: Workaround YAML gotcha in Actions

dentarg:
To avoid unexpectedly stop testing Ruby 3.0 when Ruby 3.1 is released.

See actions/runner#849

You can see that it says just `3`, not `3.0` at https://github.com/getsentry/sentry-ruby/runs/1775107213?check_suite_focus=true#step:4:4


st0012 (approved):
wow I didn't know about this 😳 thanks for the PR!

PR #1251, opened 2021-01-30
Capture exceptions in Action Cable connections and channels

### LINKED ISSUE #470: ActionCable probably doesn't work
Just realized we probably aren't catching ActionCable exceptions, need to confirm/fix

agrobbin:
Hi @st0012! I mentioned this in https://github.com/getsentry/sentry-ruby/issues/470#issuecomment-708533493 (and https://github.com/getsentry/sentry-ruby/issues/470#issuecomment-709291341), and now that 4.x is reasonably stable, thought it worth getting this PR out there.

There are 5 types of hooks that Action Cable provides:

* `Connection#connect`
* `Connection#disconnect`
* `Channel#subscribed`
* `Channel#unsubscribed`
* `Channel` actions

Now, any exceptions raised within those hooks are captured by Sentry, and reported as `ActionCable/[...]` transactions. Additional context is included depending on the hook the exception was raised within.

A note/quirk: the Rack env that's included in the scope is from the `Connection`, and therefore has a URL of the cable `mount_path` (usually `/cable`) as well as the headers from that initial connection request.

Additionally, there is not currently a really clean way to hook in and set `user_context`. I don't know if that is a blocker for this integration, but wanted to make sure it was noted.

Closes #470.

### DIFF
--- a/sentry-rails/CHANGELOG.md
+++ b/sentry-rails/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Add Action Cable exception capturing [#1251](https://github.com/getsentry/sentry-ruby/pull/1251)
+
 ## 4.1.6
 
 - Prevent exceptions app from overriding event's transaction name [#1230](https://github.com/getsentry/sentry-ruby/pull/1230)
--- a/sentry-rails/lib/sentry/rails/action_cable.rb
+++ b/sentry-rails/lib/sentry/rails/action_cable.rb
@@ -1,0 +1,72 @@
+module Sentry
+  module Rails
+    module ActionCable
+      class ErrorHandler
+        ACTION_CABLE_NAME = 'ActionCable'
+
+        def self.capture(env, transaction_name:, extra_context: nil, &block)
+          Sentry.with_scope do |scope|
+            scope.set_rack_env(env)
+            scope.set_extras(action_cable: extra_context) if extra_context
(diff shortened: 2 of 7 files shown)


agrobbin:
Hmm weird, I'm not sure why CI isn't able to load `action_cable/connection/test_case`, it runs as expected locally.

agrobbin:
Ah, mystery solved! Turns out, Rails didn't introduce its Action Cable test helpers until 6.0. I'm not sure what the preference here would be, as I am pretty sure the Action Cable integration itself would still work in 5.2, there's just not a great way to test against both versions.

Let me know what you'd like to do about that @st0012!

st0012:
@agrobbin thanks for the PR 👍 

regarding the testing helper, can you take the necessary part of it to a helper file in `sentry-rails`, and only require it when testing against Rails 5?

also, can you help me setup actioncable in the Rails 6 example app (`/examples/rails-6.0`) we use? the ideal situation is that I can trigger events just by visiting an url. there could be 4 different paths to cover the 4 types of callbacks we support.

finally, I'll change the PR's base to `4-2` branch because it's a new feature and will be introduced in v4.2.0.

st0012:
@agrobbin sorry that this was automatically closed due to I merged the `4-2` branch 🤦. can you reopen it and change the base to `master` again? thx

agrobbin:
Re-opened as #1295.

PR #1261, opened 2021-02-02
Add ability to have many post initialization callbacks

mrexox:
## Description

- [x] Add ability to have many integrations with as many changes to Configuration, as needed (current codebase allows only one post_initialization_callback)

I don't know if there is an issue about it. I faced this problem when I was trying to write my own extension of `Sentry::Configuration`. I suppose, it's not going to work with sentry-rails, because sentry-rails already redefined `post_initialization_callback`, and there was no mechanism for adding another callback to initialization process except for patching `initialize` method. I think this change will make it a bit easier in future.

### DIFF
--- a/sentry-rails/sentry-rails.gemspec
+++ b/sentry-rails/sentry-rails.gemspec
@@ -23,5 +23,5 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency "rails", ">= 5.0"
-  spec.add_dependency "sentry-ruby-core", "~> 4.1.2"
+  spec.add_dependency "sentry-ruby-core", "~> 4.2.0"
 end
--- a/sentry-rails/sentry-rails.gemspec
+++ b/sentry-rails/sentry-rails.gemspec
@@ -23,5 +23,5 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency "rails", ">= 5.0"
-  spec.add_dependency "sentry-ruby-core", "~> 4.2.0"
+  spec.add_dependency "sentry-ruby-core", "~> 4.1.2"
 end
--- a/sentry-ruby/lib/sentry/configuration.rb
+++ b/sentry-ruby/lib/sentry/configuration.rb
@@ -386,7 +386,19 @@ module Sentry
       end
     end
 
-    # allow extensions to extend the Configuration class
-    def post_initialization_callback; end
+    def post_initialization_callback
+      @@post_initialization_hooks ||= []
+      @@post_initialization_hooks.each do |hook|
+        instance_eval(&hook)
+      end
+    end
+
+    protected
+
+    # allow extensions to add their hooks to the Configuration class
+    def self.add_post_initialization_hook(&block)
+      @@post_initialization_hooks ||= []
+      @@post_initialization_hooks << block
+    end
   end
 end
--- a/sentry-ruby/lib/sentry/configuration.rb
+++ b/sentry-ruby/lib/sentry/configuration.rb
@@ -156,6 +156,10 @@ module Sentry
 
     AVAILABLE_BREADCRUMBS_LOGGERS = [:sentry_logger, :active_support_logger].freeze
 
+    # Post initialization callbacks are called at the end of initialization process
+    # allowing extending the configuration of sentry-ruby by multiple extensions
+    @@post_initialization_callbacks = []
+
     def initialize
       self.background_worker_threads = Concurrent.processor_count
       self.breadcrumbs_logger = []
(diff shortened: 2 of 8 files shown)


st0012:
@mrexox we already have a `4-2` branch and the PR here #1232. can you change your PR to base on that? thx

sentry-rails/lib/sentry/rails/version.rb
```diff
@@ -1,5 +1,5 @@
 module Sentry
   module Rails
-    VERSION = "4.1.6"
+    VERSION = "4.2.0"
```
st0012, line 3: can you revert all the version changes? we have a different release procedure here 🙂

sentry-rails/sentry-rails.gemspec
```diff
@@ -23,5 +23,5 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency "rails", ">= 5.0"
-  spec.add_dependency "sentry-ruby-core", "~> 4.1.2"
+  spec.add_dependency "sentry-ruby-core", "~> 4.2.0"
```
st0012, line 26: and please revert dependency changes too

sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -386,7 +386,19 @@ def server_name_from_env
       end
     end
 
-    # allow extensions to extend the Configuration class
-    def post_initialization_callback; end
+    def post_initialization_callback
```
st0012, line 389: since there are now multiple callbacks to be run, this should be renamed to `run_post_initialization_callbacks`

sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -386,7 +386,19 @@ def server_name_from_env
...
+      @@post_initialization_hooks.each do |hook|
+        instance_eval(&hook)
+      end
+    end
+
+    protected
+
+    # allow extensions to add their hooks to the Configuration class
+    def self.add_post_initialization_hook(&block)
+      @@post_initialization_hooks ||= []
```
st0012, line 400: this could be avoided because we should always initialize `@@post_initialization_hooks` with an array with/without anything calls `add_post_initialization_hook`.

sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -386,7 +386,19 @@ def server_name_from_env
...
+      @@post_initialization_hooks.each do |hook|
+        instance_eval(&hook)
+      end
+    end
+
+    protected
+
+    # allow extensions to add their hooks to the Configuration class
+    def self.add_post_initialization_hook(&block)
```
st0012, line 399: there should also be a class method called `post_initialization_callbacks` for retrieving the callbacks.

sentry-ruby/spec/sentry/configuration_spec.rb
```diff
@@ -367,4 +367,27 @@ class MyTestException < RuntimeError; end
       end
     end
   end
+
+  describe '.add_post_initialization_hook' do
+    module Sentry
+      class Configuration
+        attr_reader :hook1, :hook2
```
st0012, line 374: can we test this without patching the original class?

st0012 (changes requested):
thanks for the idea and PR, but I think we still need some polishing here 🙂

sentry-ruby/lib/sentry/configuration.rb
mrexox: Should I define an alias for this method too, so older versions of `sentry-rails` could work with newer version of `sentry-ruby`?

st0012: no it won't be necessary. we've restricted the dependency to `~> 4.1.2` so users will need to upgrade `sentry-rails` for `4.2` as well.

mrexox:
I guess, the problem with CI is a problem with jruby. Don't know why, but it cannot find out a `I18n::HashRefinement` although it is being used by I18n. Tests on I18n pass OK with jruby, while these tests are failing 😕 . I'll investigate on this a bit later. Maybe someone has an idea, how to fix it?

st0012:
no worries, that’s happening on master and the 4-2 branch as well, I’ll take care of that 🙂

sentry-ruby/lib/sentry/configuration.rb
mrexox: Added initialization of it to the class body

st0012:
@mrexox in case you're curious about it, the issue is a bug on jruby https://github.com/jruby/jruby/issues/6547, which was surfaced because of https://github.com/ruby-i18n/i18n/issues/554.

st0012:
@mrexox I think if you rebase the branch against `4-2`, the build will be green

st0012 (approved):
great job 👍 
one last thing before merge - can you add a changelog entry for this change 😄

PR #1265, opened 2021-02-04
Gracefully handle "sentry-trace" headers that don't match the regexp

lfittl:
Fixes the handling of invalid "sentry-trace" header values, such as "null", that otherwise lead to an application crash.

Example error backtrace:

```
Rack app error handling request { POST /action }
#<NoMethodError: undefined method `[]' for nil:NilClass>
.../sentry-ruby-core-4.1.6/lib/sentry/transaction.rb:32:in `from_sentry_trace'
.../sentry-ruby-core-4.1.6/lib/sentry/rack/capture_exceptions.rb:21:in `block in call'
.../sentry-ruby-core-4.1.6/lib/sentry/hub.rb:52:in `with_scope'
.../sentry-ruby-core-4.1.6/lib/sentry-ruby.rb:149:in `with_scope'
.../sentry-ruby-core-4.1.6/lib/sentry/rack/capture_exceptions.rb:14:in `call'
```

### DIFF
--- a/sentry-ruby/CHANGELOG.md
+++ b/sentry-ruby/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Bug Fixes
+
+- Ignore invalid values for sentry-trace header that don't match the required format [#1265](https://github.com/getsentry/sentry-ruby/pull/1265)
+
 ## 4.2.0
 
 ### Features
--- a/sentry-ruby/lib/sentry/rack/capture_exceptions.rb
+++ b/sentry-ruby/lib/sentry/rack/capture_exceptions.rb
@@ -16,12 +16,9 @@ module Sentry
           scope.set_transaction_name(env["PATH_INFO"]) if env["PATH_INFO"]
           scope.set_rack_env(env)
 
-          span =
-            if sentry_trace = env["HTTP_SENTRY_TRACE"]
-              Sentry::Transaction.from_sentry_trace(sentry_trace, name: scope.transaction_name, op: transaction_op)
-            else
-              Sentry.start_transaction(name: scope.transaction_name, op: transaction_op)
-            end
+          sentry_trace = env["HTTP_SENTRY_TRACE"]
+          span = Sentry::Transaction.from_sentry_trace(sentry_trace, name: scope.transaction_name, op: transaction_op) if sentry_trace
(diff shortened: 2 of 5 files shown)


st0012 (approved):
thanks for the fix! can you add a new changelog entry for this as well? thx

st0012 (changes requested):
actually, can you add a test case to `capture_exceptions_spec.rb`?

lfittl:
@st0012 Done - I've also found an adjacent bug, added in a follow-up commit in this PR, since its in the same code area (happy to move to a separate PR if you prefer).

st0012:
@lfittl thanks for the update 👍 some thoughts:

1. I still want to see a test in `spec/sentry/rack/capture_exceptions_spec.rb` for the invalid sentry case. I suspect that this change breaks the middleware because it assumes `span` won't be `nil` in any case.
2. the issue you mentioned is indeed a problem, but I'm not sure if adding the sampling decision is the right answer. I think the correct solution is to change `sentry-rails`'s condition on instrumentation. I'll open a PR for that.

lfittl:
> 1. I still want to see a test in `spec/sentry/rack/capture_exceptions_spec.rb` for the invalid sentry case. I suspect that this change breaks the middleware because it assumes `span` won't be `nil` in any case.

Good catch - fixed now. I assume we'd want to run through the regular logic in this case.

> 2. the issue you mentioned is indeed a problem, but I'm not sure if adding the sampling decision is the right answer. I think the correct solution is to change `sentry-rails`'s condition on instrumentation. I'll open a PR for that.

Ack, removed again from this PR. Thanks!

st0012 (commented):
thanks for catching the issue and proposed the great fix 😄 

because I just released `4.2.0` earlier today, can you rebase the branch and update the changelog again?

```
# Changelog

## Unreleased <- Add this

< and put the entry here >

## 4.2.0
.....
```

lfittl:
@st0012 Done - rebased & updated change log entry.

st0012 (approved):

st0012:
@lfittl I've talked to other maintainers about the other issue you spotted in this PR. and the answer is that the child transactions initialized from `.from_sentry_trace` should just inherit the parent's sampling decision. I've added https://github.com/getsentry/sentry-ruby/pull/1269 to update this.

Issue #1246, opened 2021-01-28
Sentry::Transport swallows network errors


sujh:
**Describe the bug**
When the event submission fails because of a network error, Sentry::Transport will log the error and swallow it. The app can't catch the error by itself, besides, `Sentry.send_event` always return a event object whether the submission is successful or not .  When I use `Sentry.send_event`,  I can't tell whether the sumission is failed from errors or returned value. Sidekiq's retry mechanism won't work because there is no error.

**Expected behavior**
Please raise the error when fail to submit event
**Actual behavior**
No error is raised when fail to submit event

**Environment**
Ruby Version: 2.7.2
SDK Version: 4.1.4
Integration Versions (if any):
Rails 6.0.3.4, Sidekiq 6.1.2

**Raven Config**

This is not necessary but could be helpful.

```ruby
    Sentry.init do |config|
      config.breadcrumbs_logger = [:active_support_logger]
      config.async = lambda { |event, hint| SentryWorker.perform_async(event, hint) }
      config.enabled_environments = %w(production)
      config.send_modules = false
      config.context_lines = 1
      config.transport.open_timeout = 2
    end
```

PR #1268, opened 2021-02-05
fix#1246: raise error when send_event failed

sujh:
## Description
Sentry::Transport will swallow exceptions when send_event failed, so the app include sidekiq can't catch error and retry.
Related issue: [https://github.com/getsentry/sentry-ruby/issues/1246]()

### DIFF
--- a/sentry-ruby/lib/sentry/transport.rb
+++ b/sentry-ruby/lib/sentry/transport.rb
@@ -33,7 +33,7 @@ module Sentry
       event
     rescue => e
       failed_for_exception(e, event)
-      nil
+      raise e
     end
 
     def generate_auth_header
--- a/sentry-ruby/spec/sentry/transport_spec.rb
+++ b/sentry-ruby/spec/sentry/transport_spec.rb
@@ -125,12 +125,12 @@ RSpec.describe Sentry::Transport do
         allow(subject).to receive(:send_data).and_raise(StandardError)
       end
 
-      it "returns nil" do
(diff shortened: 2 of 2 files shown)


st0012:
@sujh thanks for the PR! I agree that `send_event` should raise exceptions when it failed. but that means `Sentry.capture_*` would also fail if there's a networking issue. so we need to consider 2 cases:

1. when `send_event` is used alone, perhaps in a background job, it should raise exceptions when failed.
2. when `send_event` is called directly from the chain of `Sentry.capture_*` calls, it should not raise exceptions.

I'm actually working on a solution right now and will open a PR later this week 🙂

sujh:
@st0012 glad to know the progress! I am kinda confused why Sentry.capture_* shouldn't raise error. If there is an error in Sentry.capture_*, it will be swallowed by ThreadPoolExecutor.

st0012:
@sujh although events are sent by workers by default now, there still are users who fall back to the old behavior by setting

```ruby
config.background_worker_threads = 0
```

in those cases, the error will be raised in the same thread too. so we also need to maintain the same behavior for them.

sujh:
@st0012 Oh, got it. Thanks for your patiant explain🙂

st0012:
@sujh I just added https://github.com/getsentry/sentry-ruby/issues/1290 to provide a guideline on exception handling and list all the scenarios I can come up with. It'd be great if you can give it a look and provide your opinions on it 🙂

st0012:
closing because the functionality has been implemented in #1298

PR #1273, opened 2021-02-05
Add sentry-delayed_job integration

### LINKED ISSUE #1129: Add sentry-delayed_job

st0012:
Closes #1129


sentry-delayed_job/lib/sentry/delayed_job/plugin.rb
```diff
@@ -0,0 +1,42 @@
...
+require "delayed_job"
+
+module Sentry
+  module DelayedJob
+    class Plugin < ::Delayed::Plugin
+      callbacks do |lifecycle|
+        lifecycle.around(:invoke_job) do |job, *args, &block|
+          Sentry.with_scope do |scope|
+            scope.set_extras(**generate_extra(job))
+            scope.set_tags("delayed_job.queue" => job.queue, "delayed_job.id" => job.id)
```
dan-jensen, line 11: With Ruby 2.6 this line raises `TypeError: hash key "delayed_job.id" is not a Symbol` (and for the similar string keys below).  [Apparently](https://bugs.ruby-lang.org/issues/16853#note-9) "Keyword splat would not allow anything but symbols as keys pre-2.7".  To get it to work on Ruby 2.6 I changed them all to symbols like `scope.set_tags(:delayed_job_queue =>...`.  Admittedly not as nice as your dot version, but is there any other option for backward compatibility?

dan-jensen: Just for compmarison: it looks like `sentry-sidekiq` [doesn't set a prefix](https://github.com/getsentry/sentry-ruby/blob/master/sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb#L13), but I think the prefix is a good idea.

st0012: because Sentry doesn't display nested data structure, so I'm trying the prefix approach in this gem to make things look nicer.

`not prefixed`

截圖 2021-02-06 上午9 48 34

`prefixed`

截圖 2021-02-06 上午9 48 03

dan-jensen: @st0012 yes, I understand – but is it possible for prefixes to be compatible with Ruby 2.6 and lower?

Issue #1274, opened 2021-02-05
Issue with 4.2.0 -> IPAddr::AddressFamilyError: address family must be specified


johnernaut:
**Describe the bug**
When running our application in production (via ECS and Fargate), we're seeing an application error which causes the process to die immediately.  Here's an example taken from the logs:
```
Feb 05 13:04:52.630		rails	/var/www/homebase/vendor/bundle/ruby/2.6.0/gems/puma-3.12.6/lib/puma/thread_pool.rb:135:in `block in spawn_thread'
Feb 05 13:04:52.630		rails	/var/www/homebase/vendor/bundle/ruby/2.6.0/gems/puma-3.12.6/lib/puma/server.rb:334:in `block in run'
Feb 05 13:04:52.630		rails	/var/www/homebase/vendor/bundle/ruby/2.6.0/gems/puma-3.12.6/lib/puma/server.rb:476:in `process_client'
Feb 05 13:04:52.630		rails	/var/www/homebase/vendor/bundle/ruby/2.6.0/gems/puma-3.12.6/lib/puma/server.rb:706:in `handle_request'
Feb 05 13:04:52.630		rails	/var/www/homebase/vendor/bundle/ruby/2.6.0/gems/puma-3.12.6/lib/puma/configuration.rb:227:in `call'
Feb 05 13:04:52.630		rails	/var/www/homebase/vendor/bundle/ruby/2.6.0/gems/newrelic_rpm-6.8.0.360/lib/new_relic/agent/instrumentation/middleware_tracing.rb:99:in `call'
Feb 05 13:04:52.630		rails	/var/www/homebase/vendor/bundle/ruby/2.6.0/gems/railties-6.0.3.3/lib/rails/railtie.rb:190:in `method_missing'
Feb 05 13:04:52.630		rails	/var/www/homebase/vendor/bundle/ruby/2.6.0/gems/railties-6.0.3.3/lib/rails/railtie.rb:190:in `public_send'
Feb 05 13:04:52.629		rails	/var/www/homebase/vendor/bundle/ruby/2.6.0/gems/railties-6.0.3.3/lib/rails/engine.rb:527:in `call'
Feb 05 13:04:52.629		rails	/var/www/homebase/vendor/bundle/ruby/2.6.0/gems/newrelic_rpm-6.8.0.360/lib/new_relic/agent/instrumentation/middleware_tracing.rb:99:in `call'
[...]

PR #1288, opened 2021-02-15
always convert trusted proxies to string

gagalago:
fix https://github.com/getsentry/sentry-ruby/issues/1274

## Description
`IPAddr.new` only accept the address as a string but not as an `IPAddr` as suggested in the [Rails documentation](https://api.rubyonrails.org/classes/ActionDispatch/RemoteIp.html).

### DIFF
--- a/sentry-rails/lib/sentry/rails/railtie.rb
+++ b/sentry-rails/lib/sentry/rails/railtie.rb
@@ -41,7 +41,7 @@ module Sentry
     end
 
     def configure_trusted_proxies
-      Sentry.configuration.trusted_proxies += Array(::Rails.application.config.action_dispatch.trusted_proxies)
+      Sentry.configuration.trusted_proxies += Array(::Rails.application.config.action_dispatch.trusted_proxies).map(&:to_s)
     end
 
     def extend_active_job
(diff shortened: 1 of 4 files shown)


woidda (approved):

mihaic195 (approved):

st0012 (changes requested):
thanks for the PR, but I think the solution can still be improved 🙂 

although `sentry-rails` is where most users would register the trusted proxies, `config.trusted_proxies` is also a public configuration option. so even though the PR can stop most of the problematic cases, it doesn't prevent users from manually injecting the converted ip addresses. imo, it's the `Utils::RealIP`'s responsibility to check every ip address' type and whether to convert them.

also, there should be test cases to cover the situation this PR wants to address.

woidda (approved):

st0012 (approved):
thanks for the fix! can you also add a changelog entry?

gagalago:
> thanks for the fix! can you also add a changelog entry?

I did it by hand. As it's not possible to do it with craft when it's not the master branch. 

Can be interesting for next time :wink:  to add on the `CONTRIBUTING.md` or in the pull request template who is responsible to add the entry in the changelog and how to do it in a feature branch (to be compatible later with craft).

st0012:
@gagalago thanks for the advice 😄 I also add changelogs by hand tbh. but I'll mention the changelog entry in the pr template 👍

PR #1314, opened 2021-03-04
Compress event payload by default

st0012:
Rules:
- If `config.transport.encoding == "gzip"` and the event envelope's  size is larger than 30kb, the envelope will be compressed.
- If `config.transport.encoding == "gzip"` but the event envelope's  size is smaller than 30kb, the envelope won't be compressed.
- If `config.transport.encoding != "gzip"`, the event won't be  compressed regardless how but it is.

### DIFF
--- a/sentry-ruby/lib/sentry/transport/http_transport.rb
+++ b/sentry-ruby/lib/sentry/transport/http_transport.rb
@@ -1,8 +1,12 @@
 require 'faraday'
+require 'zlib'
 
 module Sentry
   class HTTPTransport < Transport
-    CONTENT_TYPE = 'application/json'
+    GZIP_ENCODING = "gzip"
+    GZIP_THRESHOLD = 1024 * 30
+    CONTENT_TYPE = 'application/x-sentry-envelope'
+
     attr_reader :conn, :adapter
 
     def initialize(*args)
@@ -13,8 +17,16 @@ module Sentry
     end
 
     def send_data(data)
+      encoding = ""
+
+      if should_compress?(data)
+        data = Zlib.gzip(data)
+        encoding = GZIP_ENCODING
+      end
+
       conn.post @endpoint do |req|
         req.headers['Content-Type'] = CONTENT_TYPE
+        req.headers['Content-Encoding'] = encoding
         req.headers['X-Sentry-Auth'] = generate_auth_header
         req.body = data
       end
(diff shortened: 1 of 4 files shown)


sentry-ruby/lib/sentry/transport/http_transport.rb
```diff
@@ -13,6 +18,10 @@ def initialize(*args)
     end
 
     def send_data(data)
+      if should_compress?(data)
+        data = Base64.strict_encode64(Zlib.deflate(data))
```
jan-auer, line 22: This uses our legacy compression format, which uses deflate + base64, not gzip. While we technically support it on the envelope endpoint, it has a few drawbacks. Could you instead set the `Content-Encoding` encoding header and apply actual gzip compression?

If gzip is not an option, you can also set the content encoding to `deflate` and then do a full `zlib` compression (including the zlib structure header). 

See [Request Compression](https://develop.sentry.dev/sdk/overview/#request-compression) and [`Content-Encoding`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding) for more information.

sentry-ruby/lib/sentry/transport/http_transport.rb
```diff
@@ -1,8 +1,13 @@
 require 'faraday'
+require 'zlib'
+require 'base64'
 
 module Sentry
   class HTTPTransport < Transport
-    CONTENT_TYPE = 'application/json'
+    GZIP_ENCODING = "gzip"
+    GZIP_THRESHOLD = 1024 * 30
```
jan-auer, line 8: What is the "usual" size and throughput of Ruby events that you've seen so far? A threshold of somewhere around 32kB seems reasonable intuitively, I'd just love to confirm.

jan-auer (changes requested):

sentry-ruby/lib/sentry/transport/http_transport.rb
st0012: In Rails apps, I think it's generally over 40kb because of the framework's stacktrace. But it could be quite small on non-Ruby apps, like 1x kb.

jan-auer (approved):
Looks good, thanks! Deferring final review to @HazAT.

The implementation sets `Content-Encoding: <empty>` if `should_compress?` returns false. I believe that this will work, but please make sure to test this against Sentry SaaS once.

st0012:
@jan-auer thanks for the review. I've tested it and it works as expected 🙂

HazAT (approved):

PR #1328, opened 2021-03-11
fix(tracing): Never send transactions if tracing is disabled

lobsterkatie:
**EDIT:** I've realized this contains a breaking change, so I'm going to close this in favor of #1326, which has a smaller scope, and apply the spec-fixing parts of this separately.

--------------------------------------------------------------

This fixes an edge case where an upstream SDK would send a `sentry-trace` header without a sampling decision, and that lack of a "no" decision would be read as a "yes," causing transactions to be sent from the downstream SDK, in spite of `traces_sample_rate` being either unset or set to `0`.

There are a number of changes included here, which together both fix the bug and bring the SDK better in line with the [performance spec](https://develop.sentry.dev/sdk/performance/):

- The criteria for tracing being enabled have been changed so that a) `0` is now a valid sample rate but b) no number outside of the interval `0,1]` is and nothing that isn't a number is.

- In every place where it's relevant, the SDK now bails immediately if tracing is disabled. (This used to happen in some but not all places.) New locations include multiple `start_transaction` functions/methods (which create a transaction and/or set  its sampling decision) and `transaction.finish`(which ultimately decides whether or not to create an event out of the transaction).

- Because this means transactions will sometimes not be created where they previously were, checks are put in place so transaction-y stuff is only done if the transaction actually exists.

- The logic for parsing the sampling portion of an incoming `sentry-trace` SDK header has been made more explicit, so that each of the three cases (incoming yes, incoming no, and incoming undecided) are handled explicitly. (This is, in and of itself, the primary fix for the aforementioned bug.)

- The logic handling converting an incoming parent transaction's sampling decision into a newly-created child's sampling decision has been moved to the method setting the transaction's sampling decision via `traces_sample_rate` or `traces_sampler`. This has two advantages: first, all the logic is in one place, and second, it allows us to use the proper precedence order for different ways of setting the sampling decision, rather than putting the parent's decision ahead of everything else.

- The `transaction.set_initial_sample_decision` method now doesn't set a sampling decision at all if tracing is disabled, to better differentiate, if we ever need to, between that situation and a tracing-enabled-but-transaction-not-sampled situation.

- The `transaction.finish` method no longer checks the parent's sampling decision, because it is always handled when the transaction is started.

- Finally, tests have been added and modified as necessary.

### DIFF
--- a/sentry-ruby/lib/sentry/rack/capture_exceptions.rb
+++ b/sentry-ruby/lib/sentry/rack/capture_exceptions.rb
@@ -17,23 +17,23 @@ module Sentry
           scope.set_rack_env(env)
 
           transaction = start_transaction(env, scope)
-          scope.set_span(transaction)
+          transaction ? scope.set_span(transaction) : nil
 
           begin
             response = @app.call(env)
           rescue Sentry::Error
-            finish_transaction(transaction, 500)
+            transaction ? finish_transaction(transaction, 500) : nil
             raise # Don't capture Sentry errors
           rescue Exception => e
             capture_exception(e)
-            finish_transaction(transaction, 500)
+            transaction ? finish_transaction(transaction, 500) : nil
             raise
           end
 
           exception = collect_exception(env)
           capture_exception(exception) if exception
 
-          finish_transaction(transaction, response[0])
+          transaction ? finish_transaction(transaction, response[0]) : nil
 
           response
         end
@@ -63,6 +63,8 @@ module Sentry
 
 
       def finish_transaction(transaction, status_code)
+        return unless transaction
+
         transaction.set_http_status(status_code)
         transaction.finish
       end
--- a/sentry-ruby/lib/sentry/rack/capture_exceptions.rb
+++ b/sentry-ruby/lib/sentry/rack/capture_exceptions.rb
@@ -54,6 +54,8 @@ module Sentry
       end
 
       def start_transaction(env, scope)
+        return unless Sentry.configuration.tracing_enabled?
+
         sentry_trace = env["HTTP_SENTRY_TRACE"]
         transaction = Sentry::Transaction.from_sentry_trace(sentry_trace, name: scope.transaction_name, op: transaction_op) if sentry_trace
         transaction || Sentry.start_transaction(name: scope.transaction_name, op: transaction_op)
--- a/sentry-ruby/lib/sentry/rack/capture_exceptions.rb
+++ b/sentry-ruby/lib/sentry/rack/capture_exceptions.rb
@@ -57,8 +57,11 @@ module Sentry
         return unless Sentry.configuration.tracing_enabled?
 
         sentry_trace = env["HTTP_SENTRY_TRACE"]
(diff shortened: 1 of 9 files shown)


sentry-ruby/lib/sentry/rack/capture_exceptions.rb
```diff
@@ -17,23 +17,23 @@ def call(env)
           scope.set_rack_env(env)
 
           transaction = start_transaction(env, scope)
-          scope.set_span(transaction)
+          transaction ? scope.set_span(transaction) : nil
```
st0012, line 20: since we don't need the return value from this line, it's not necessary to return the `nil`. this should do the work too

```ruby
scope.set_span(transaction) if transaction
```

sentry-ruby/lib/sentry/rack/capture_exceptions.rb
```diff
@@ -17,23 +17,23 @@ def call(env)
...
 
           begin
             response = @app.call(env)
           rescue Sentry::Error
-            finish_transaction(transaction, 500)
+            transaction ? finish_transaction(transaction, 500) : nil
```
st0012, line 25: this and the below change are unnecessary with the early return in `finish_transaction`

sentry-ruby/lib/sentry/rack/capture_exceptions.rb
```diff
@@ -54,13 +54,20 @@ def capture_exception(exception)
...
       def start_transaction(env, scope)
+        return unless Sentry.configuration.tracing_enabled?
+
         sentry_trace = env["HTTP_SENTRY_TRACE"]
-        transaction = Sentry::Transaction.from_sentry_trace(sentry_trace, name: scope.transaction_name, op: transaction_op) if sentry_trace
-        transaction || Sentry.start_transaction(name: scope.transaction_name, op: transaction_op)
+        options = {name: scope.transaction_name, op: transaction_op}
+
+        # if tracing is disabled, these will both return nil
+        transaction = Sentry::Transaction.from_sentry_trace(sentry_trace, **options) if sentry_trace
```
st0012, line 63: makes sense, will adopt it in my PR 👍

st0012 (commented):
thanks for the PR, I think it's a great work and I learned a lot from it 🙂 
I've picked some of the changes to my PR and will implement some of the riskier changes with you later 👍

sentry-ruby/lib/sentry/rack/capture_exceptions.rb
lobsterkatie: So is that formulation more idiomatic for Ruby? As someone coming from Python and JS/TS, having the condition after the action made my brain hurt a little, which is why I went with the ternary in the end. But if the `do x if y` pattern is the accepted way, that's good to know. 🙂

sentry-ruby/lib/sentry/rack/capture_exceptions.rb
lobsterkatie: Ah, right - good point.

sentry-ruby/lib/sentry/rack/capture_exceptions.rb
st0012: yeah this is a common practice in Ruby 😂 we're weirdos 😉

PR #1335, opened 2021-03-13
Fix sampling decision precedence

st0012:
1. When traces_sampler is set, the sentry-trace's sampling decision should be put into sampling context to let the user make explicit decision. Currently it forces the transaction to be sampled, which is wrong.
2. The value of traces_sample_rate should have a lower priority than the  traces_sampler and the value of parent_sampled.

### DIFF
--- a/sentry-ruby/CHANGELOG.md
+++ b/sentry-ruby/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 - Correct type attribute's usages [#1354](https://github.com/getsentry/sentry-ruby/pull/1354)
-
+- Fix sampling decision precedence [#1335](https://github.com/getsentry/sentry-ruby/pull/1335)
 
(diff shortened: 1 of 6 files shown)


lobsterkatie (commented):
Overall looks good.

There are a few tests I think we should add, to show that precedence is working:

- `"when traces_sampler is provided"` context:
  - `it "prioritizes traces_sampler over traces_sample_rate"`, where the sampler always says yes and the rate is 0, and/or the sampler always says no and the rate is 1
  - `it "prioritizes traces_sampler over inherited decision"` (similar idea)
- `"when traces_sampler is not set"` context:
  - `it "prioritizes inherited decision over traces_sample_rate"` (again, similar idea)

st0012:
@lobsterkatie I also added another commit to correct `sampling_context`'s creation.

PR #1336, opened 2021-03-13
Support category-based rate limiting

### LINKED ISSUE #1072: [4.0] Category-based Ratelimiting
SDKs that use the envelope endpoint should support Category-based Ratelimiting.

This is the code in JS
https://github.com/getsentry/sentry-javascript/pull/2962

It's all about reading the `x-sentry-rate-limits` response header and keeping an internal map in the transport from `type -> duration` to know what has been rate limited.

st0012:
Closes #1072


sentry-ruby/spec/sentry/transport/http_transport_rate_limiting_spec.rb
```diff
@@ -0,0 +1,252 @@
+require 'spec_helper'
+
+RSpec.describe "rate limiting" do
```
st0012, line 3: @jan-auer if you have time, can you go through this spec file to see if all the cases are described correctly? thanks

sentry-ruby/lib/sentry/transport.rb
```diff
@@ -7,20 +7,36 @@ class Transport
...
+      @rate_limits = {}
     end
 
     def send_data(data, options = {})
       raise NotImplementedError
     end
 
     def send_event(event)
+      event_hash = event.to_hash
+      event_type = event_hash[:type] || event_hash['type']
```
jan-auer, line 25: Is this the `type` coming from the event payload? I'm asking because `event` is not a valid event type.

For reference, the valid types are: `error`, `csp`, `hpkp`, `expectct`, `expectstaple`, `transaction`. I would propose that you check for `transaction`, and in all other cases fall back to `error`. 

Please ensure that `event.type: event` is not sent to the server.

sentry-ruby/lib/sentry/transport.rb
```diff
@@ -33,6 +49,23 @@ def send_event(event)
       event
     end
 
+    def is_rate_limited?(event_type)
+      # check category-specific limit
+      delay =
+        case event_type
+        when "event"
+          # confusing mapping, but it's decided by Sentry
```
jan-auer, line 57: To shed some light on the decision behind this naming convention: Sentry understands several event types, among others "error" events (for error tracking) and "transaction" events (for performance).

sentry-ruby/lib/sentry/transport.rb
```diff
@@ -33,6 +49,23 @@ def send_event(event)
...
+          @rate_limits["error"]
+        when "transaction"
+          @rate_limits["transaction"]
+        end
+
+      # check universal limit if not category limit
+      delay ||= @rate_limits[nil]
+
+      !!delay && delay > Time.now
```
jan-auer, line 66: Please check both the universal and the category-based limit. The stored category limit could be lower than the universal limit.

sentry-ruby/lib/sentry/transport/http_transport.rb
```diff
@@ -24,25 +28,87 @@ def send_data(data)
...
+    def has_rate_limited_header?(headers)
+      headers[RETRY_AFTER_HEADER] || headers[RATE_LIMIT_HEADER]
+    end
+
+    def handle_rate_limited_response(headers)
+      rate_limits =
+        if rate_limits = headers[RATE_LIMIT_HEADER]
+          parse_rate_limit_header(rate_limits)
+        elsif retry_after = headers[RETRY_AFTER_HEADER]
+          retry_after = retry_after.to_i
```
jan-auer, line 67: According to HTTP specification, the `retry-after` header can also contain a HTTP Date string. Sentry never responds with strings, however. It would be great to add a comment on that for the future.

sentry-ruby/lib/sentry/transport/http_transport.rb
```diff
@@ -24,25 +28,87 @@ def send_data(data)
...
+
+    def parse_rate_limit_header(rate_limit_header)
+      time = Time.now
+
+      result = {}
+
+      limits = rate_limit_header.split(",")
+      limits.each do |limit|
+        begin
+          retry_after, categories = limit.strip.split(":").first(2)
```
jan-auer, line 94: It's unlikely, but could be possible that `limit` is an empty string, potentially after splitting. Ideally, check for that and bail out.

sentry-ruby/spec/sentry/transport/http_transport_rate_limiting_spec.rb
```diff
@@ -0,0 +1,252 @@
...
+
+      it "returns false when passed limit" do
+        subject.rate_limits.merge!(nil => Time.now - 10)
+
+        expect(subject.is_rate_limited?("event")).to eq(false)
+      end
+    end
+
+    context "with both category-based and universal limits" do
+      it "prioritizes category limits" do
```
jan-auer, line 59: The assertion and behavior in this test are correct, but the assumption is not. This links together with the comment above: Instead of _prioritizing_ the category limit, `is_rate_limited` needs to check _both_.

Can you add another test case like this?

```ruby
      it "always checks universal limits" do
        subject.rate_limits.merge!(
          "error" => Time.now - 10,
          nil => Time.now + 60
        )

        expect(subject.is_rate_limited?("event")).to eq(true)
      end
```

sentry-ruby/spec/sentry/transport/http_transport_rate_limiting_spec.rb
```diff
@@ -0,0 +1,252 @@
...
+      configuration.transport.http_adapter = [:test, stubs]
+    end
+
+    shared_examples "rate limiting headers handling" do
+      context "with x-sentry-rate-limits header" do
+        now = Time.now
+
+        [
+          {
+            header: "", expected_limits: {}
```
jan-auer, line 81: It's unlikely to occur, but let's also add a test for this:

```
          {
            header: " ", expected_limits: {}
          }
```

jan-auer (changes requested):
Thank you! Please see the comments below. There is one request for change in the code checking for cached limits, and a few suggestions and clarifying questions.

sentry-ruby/spec/sentry/transport/http_transport_rate_limiting_spec.rb
jan-auer: Thank you, the specs all look good. Added two comments for additional specs below.

sentry-ruby/lib/sentry/transport.rb
st0012: wow this is the first time I learn this. where is this documented? the closest thing I find right now is in the [schema repo](https://github.com/getsentry/sentry-data-schemas/blob/main/relay/event.schema.json#L346-L347) (through [this document](https://develop.sentry.dev/sdk/event-payloads/types/)). but that doesn't exactly match watch you mentioned here.

jan-auer: https://github.com/getsentry/relay/pull/954

st0012: so to summarize:
- all transactions should have a field `event_type: "transaction"`
- events should not have `event_type` field
- when rate limiting, events should be mapped with `error` category regardless if the event actually contains an  error

jan-auer: The field should be `type`, otherwise that's correct 👍 

> when rate limiting, events should be mapped with error category regardless if the event actually contains an error

Yes. This is a simplification of the actual system, but it is correct enough and we'll also document it soon this way. Just ensure that `type: "transaction"` events are mapped with their own category.

st0012: thanks for the confirmation, I've implemented the type changes in a separate PR: https://github.com/getsentry/sentry-ruby/pull/1354

jan-auer (approved):

PR #1341, opened 2021-03-16
Fix handle the case where the logger messages is not of String type

fumitti:
## Description
Passing something other (like Integer) than a String type to the logger will fail because there is no "empty?" method.
Fixed based on old raven source
https://github.com/getsentry/sentry-ruby/blob/master/sentry-raven/lib/raven/breadcrumbs/sentry_logger.rb#L22


st0012 (approved):

st0012:
@fumitti thanks for catching this and making this PR 😄

justi:
@fumitti thanks for the fix. I've notice the same bug and temporary fix is replacing sentry_logger with active_support_logger ```
config.breadcrumbs_logger = %i[active_support_logger]
```

PR #1298, opened 2021-02-25
Improve SDK's error handling

st0012:
This implements the error handling rules described in #1290 

closes #1246, #1290 and #1289

PR #1343, opened 2021-03-16
Use debug log level for send event log statements

nhippenmeyer:
For applications that trigger large volumes of events, these log statements can quickly incur large costs from 3rd party logging services. We should have the option to disable them without having to set the log level application-wide to `warn` or higher.

The 4.3.0 release appears to be responsible for a large increase in the number of these log statements. Our application went from generating ~30k log statements per day on v4.2.2 to ~6M per day on v4.3.0.

### DIFF
--- a/sentry-ruby/lib/sentry/transport.rb
+++ b/sentry-ruby/lib/sentry/transport.rb
@@ -66,7 +66,7 @@ module Sentry
 
       event_id = event_hash[:event_id] || event_hash["event_id"]
       event_type = event_hash[:type] || event_hash["type"]
-      configuration.logger.info(LOGGER_PROGNAME) { "Sending #{event_type} #{event_id} to Sentry" }
+      configuration.logger.debug(LOGGER_PROGNAME) { "Sending #{event_type} #{event_id} to Sentry" }
       encode(event_hash)
     end
   end
--- a/sentry-ruby/spec/sentry/transport_spec.rb
+++ b/sentry-ruby/spec/sentry/transport_spec.rb
@@ -2,7 +2,7 @@ require 'spec_helper'
 
 RSpec.describe Sentry::Transport do
   let(:io) { StringIO.new }
-  let(:logger) { Logger.new(io) }
+  let(:logger) { Logger.new(io, :debug) }
   let(:configuration) do
     Sentry::Configuration.new.tap do |config|
       config.server = 'http://12345:67890@sentry.localdomain/sentry/42'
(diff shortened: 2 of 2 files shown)


st0012:
I made several changes in 4.3.0 to log more information when the SDK failed to send events. PR: https://github.com/getsentry/sentry-ruby/pull/1298
But the line you refer to is not one of them and doesn't look like the root cause. I suspect the increased volume is mainly caused by error messages (which could be swallowed before 4.3.0). Can you search for `sending failed:` in your log and let me know what you find?

nhippenmeyer:
Thanks for the feedback @st0012. The log statements we are seeing are of the form:

`sentry -- Sending transaction a793922ed2604f65aa0eed062c90bc23 to Sentry`

These are being generated from `sentry-rails` and `sentry-sidekiq`, however the large increase in log volume we started seeing in v4.3.0 is due to transactions recorded from `sentry-sidekiq`.

[This commit](https://github.com/getsentry/sentry-ruby/commit/58b15eb83209d845ecc0af34769bf7d63b42fa70#diff-35a3cf671853d934768e0205a63982512d8083dba25bd5343439d952fb1c3cc3) appears to be responsible for the change.

st0012:
@nhippenmeyer Thanks for the feedback. But based on Ruby logger's level definition:

> Generic (useful) information about system operation.

I think using `info` to notify outgoing error reporting requests is a reasonable use of logging level.

Instead, I think the issue could be on the newly added sidekiq tracing. If the sidekiq tracing increases the volume of transactions so much, would you instead considering stop/reduce that? Also, how valuable do you think this feature provides to your team?

nhippenmeyer:
I've disabled sidekiq tracing and confirmed the number of log events has returned to previous levels. I'm not sure how valuable the tracing feature would be to our team since we haven't utilized any of the features yet, but would consider enabling it again in the future if we can find away around this logging issue.

Thanks for your help here @st0012.

st0012:
How do you turn off sidekiq's performance tracing though? With `traces_sampler`? I'm just curious here 🙂 

Also, I'm not familiar with your application but it sounds like you have an enormous number of background jobs. If that's the case, I'd suggest you use `traces_sampler` instead of `traces_sample_rate` for better sampling. Something like this:

```ruby
config.traces_sampler = lambda do |sampling_context|
  transaction_context = sampling_context[:transaction_context]
  op = transaction_context[:op]
  if op.match? /request/
    0.1 # the rate for requests
  elsif op.match? /sidekiq/
    0.001 # so you (roughly) only get 1000 transactions per million jobs
  end
end
```

nhippenmeyer:
> How do you turn off sidekiq's performance tracing though? With `traces_sampler`? I'm just curious here 🙂

We were previously using a sample rate of 0.5 with: `config.traces_sample_rate = 0.5`.

I like your suggestion of configuring the sample rate independently for Sidekiq. We do run a very high number of background jobs (~12M/day) so sampling them at a lower rate makes sense. Thanks again for the help!

PR #1352, opened 2021-03-19
Minimize sentry-rails' dependency requirement

### LINKED ISSUE #1340: `sentry-rails` should depend on `railties` gem
**Describe the idea**
Right now if only `railties` and `active_record` are in the `Gemfile`, installing `sentry-rails` pulls all the `rails` with ActionText and whatever.

We shouldn't pull megabytes of unused dependencies.

**Possible implementation**
Change `rails` to `railties` in the gemspec & make integrations setup `if defined?` like it's already done for ActiveJob: `extend_active_job if defined?(ActiveJob)`.

st0012:
1. `sentry-rails` will only require `railties` and its dependencies as dependency. so users won't need to have `rails` installed anymore.
2. `sentry-rails` should now work with minimum Rails apps (e.g. without ActiveRecord).

closes #1340

### DIFF
--- a/sentry-rails/CHANGELOG.md
+++ b/sentry-rails/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 4.3.3
+
+- Minimize sentry-rails' dependency requirement [#1352](https://github.com/getsentry/sentry-ruby/pull/1352)
+
 ## 4.3.2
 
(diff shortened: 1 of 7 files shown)


lfittl:
@st0012 With today's rails + mimemagic problems, this becomes ever more important (if one does not use activestorage one can avoid to pull in rails, but sentry-rails right now makes that impossible) - thanks for working on this!

Is there a chance to get this merged and a new version released quickly?

PR #1354, opened 2021-03-21
Correct type attribute's usages

st0012:
Based on [this discussion](https://github.com/getsentry/sentry-ruby/pull/1336/files#r596851969), normal events shouldn't have the `type` attribute and only transaction events should have `type: "transaction"`.

So this PR:
- removes `type: "event"` from normal events
- corrects `Transport`'s envelope generation logic: envelope's type is not equivalent to event's type
- corrects the logging message for sending envelopes

### DIFF
--- a/sentry-ruby/lib/sentry/transport.rb
+++ b/sentry-ruby/lib/sentry/transport.rb
@@ -24,7 +24,7 @@ module Sentry
         return
       end
 
-      encoded_data = prepare_encoded_event(event)
+      encoded_data = encode(event)
 
       return nil unless encoded_data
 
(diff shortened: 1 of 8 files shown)


jan-auer:
Adding @rhcarvalho to double-check this.

sentry-ruby/lib/sentry/transport.rb
```diff
@@ -45,29 +45,22 @@ def generate_auth_header
...
 
-    def encode(event_hash)
-      event_id = event_hash[:event_id] || event_hash['event_id']
-      event_type = event_hash[:type] || event_hash['type']
+    def encode(event)
+      # Convert to hash
+      event_hash = event.to_hash
+
+      event_id = event_hash[:event_id] || event_hash["event_id"]
+      envelope_type = event_hash[:type] || event_hash["type"] || "event"
```
jan-auer, line 53: This is the type of the item in the envelope, not the type of the envelope itself. There could, for instance, be another item of type `"attachment"`. See [here](https://develop.sentry.dev/sdk/envelopes/#data-model).

```suggestion
      item_type = event_hash[:type] || event_hash["type"] || "event"
```

jan-auer (approved):
Looks good to me, one minor suggestion below to clarify on the item type.

sentry-ruby/lib/sentry/transport.rb
st0012: 👍 updated

rhcarvalho (approved):

PR #1355, opened 2021-03-21
Webpacker example

st0012:
This is specifically for debugging https://github.com/getsentry/sentry/issues/23937

To compile the javascript files and upload source maps: 

```
$ RAILS_ENV=production bundle exec rails assets:clobber webpacker:compile
```

### DIFF
--- a/sentry-rails/examples/rails-6.0/config/webpack/development.js
+++ b/sentry-rails/examples/rails-6.0/config/webpack/development.js
@@ -2,4 +2,19 @@ process.env.NODE_ENV = process.env.NODE_ENV || 'development'
 
 const environment = require('./environment')
 
+const SentryCliPlugin = require("@sentry/webpack-plugin")
+
+environment.config.merge({ devtool: "source-map" })
+environment.plugins.prepend(
+  "Sentry",
+  new SentryCliPlugin({
+    include: "./public/",
+    ignore: ["node_modules"],
+    authToken: process.env.SENTRY_AUTH_TOKEN,
+    org: "sentry-sdks",
+    project: "sentry-ruby",    
+    stripPrefix: ["public"],
+  })
+)
+
 module.exports = environment.toWebpackConfig()
--- a/sentry-rails/examples/rails-6.0/Gemfile
+++ b/sentry-rails/examples/rails-6.0/Gemfile
@@ -13,7 +13,7 @@ gem 'unicorn'
 # Use SCSS for stylesheets
 gem 'sass-rails', '~> 5'
 # Transpile app-like JavaScript. Read more: https://github.com/rails/webpacker
-gem 'webpacker', '~> 4.0'
+gem 'webpacker', "~> 5.2"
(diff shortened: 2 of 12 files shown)


sentry-rails/examples/rails-6.0/config/webpack/development.js
```diff
@@ -2,4 +2,19 @@ process.env.NODE_ENV = process.env.NODE_ENV || 'development'
...
+environment.config.merge({ devtool: "source-map" })
+environment.plugins.prepend(
+  "Sentry",
+  new SentryCliPlugin({
+    include: "./public/",
+    ignore: ["node_modules"],
+    authToken: process.env.SENTRY_AUTH_TOKEN,
+    org: "sentry-sdks",
+    project: "sentry-ruby",    
+    stripPrefix: ["public"],
```
cxspxr, line 16: so we basically need to both `include` pointing to `public` + a `stripPrefix` of `["public"]`? Because previously I thought that `include` would change a starting path to be based from `public`, thus not including it in the paths.

So what I tried before was either:
`include: ".", stripPrefix: ["public"]`
OR
`include: "./public/"` without `stripPrefix`

I will try doing both today

st0012: @cxspxr well this is not the practice we recommend, this PR is just for reproducing the issue you have. so I was trying to replicate the setting you posted in the issue report 🙂

cxspxr: oh, yeah, sorry, we have tried the very same config as well, had no luck though

cxspxr: yeah, thanks, just  noticed that I sent the sample config to you with both options set, so yes, we have tried this as well

st0012: have you tried using the same config but in `development.js` and run

```
$ bundle exec rails assets:clobber assets:precompile
```

it'd help to know if the issue is present in the development mode as well.

cxspxr: @st0012 it does upload correctly without `~/public` in the path with the same config in `development.js`
But it does upload it correctly for `production.js` as well, I am not sure how some artifacts have been uploaded to sentry with `public` in their paths, because it uploads artifacts without this prefix

cxspxr: Maybe guys discovered some earlier uploaded artifacts instead? Because we tried very different configs by now. And for review apps (heroku pipeline) Heroku fails to provide correct `SOURCE_VERSION` so we basically dumping the very same release ID every time for review apps. This is however correctly set for `staging` and `production` environments. 

My thought is that maybe those artifacts with `public` in their path were old? Just pointing to the newer releases as `SOURCE_VERSION` env variable on Heroku is never changed for review apps and we use the same for every single release there.

PR #1356, opened 2021-03-22
Change to reduced dependencies

mbj:
Change to reduced dependencies

  * Changes to depend exclusively on the rails subcomponents that are
    required to get specs passing.
  * Removes the need to pull in other rails dependencies that sentry-rails
    currently does NOT integrate with.
  * Improves compatibility with more ruby projects due reduced dependency
    footprint.

### DIFF
--- a/sentry-rails/Gemfile
+++ b/sentry-rails/Gemfile
@@ -15,7 +15,9 @@ else
   gem "sqlite3", platform: :ruby
 end
 
-gem "rails", "~> #{rails_version}"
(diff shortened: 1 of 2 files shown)


st0012:
thanks for the PR, but there's already one for this purpose 🙂 https://github.com/getsentry/sentry-ruby/pull/1352

mbj:
@st0012 Thanks! Sorry I did not spot the correlation between both when I browsed the open PRs. Cheers.

PR #1367, opened 2021-03-26
feat: allow setting fingerprint from Sentry module

fsateler:
This way `configure_scope` is not needed.

**Other Notes**

## Description
Describe your changes:

### DIFF
--- a/sentry-ruby/lib/sentry-ruby.rb
+++ b/sentry-ruby/lib/sentry-ruby.rb
@@ -58,7 +58,7 @@ module Sentry
     extend Forwardable
 
     def_delegators :get_current_client, :configuration, :send_event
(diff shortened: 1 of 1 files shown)


st0012:
@rhcarvalho I don't see a reason to reject this. do other SDKs support the top-level `set_fingerprint` as well?

rhcarvalho:
The SDK team has discussed this yesterday.

Other SDKs don't support this shortcut. Before we add more global methods there needs to be a broader discussion about where the API is heading, as we want to avoid making one-off decisions.

The `set_fingerprint` method, in particular, is meant to change the fingerprint of a single event, and thus having this global shortcut could bring unintended side-effects and usage error.

The decision was to not merge this now, collect the data point, and defer the broader discussion for later.

PR #1382, opened 2021-04-04
Fix Span/Transaction's nesting issue

### LINKED ISSUE #1372: Trouble nesting children successfully with start_child.
I'm trying to use the `Sentry.start_transaction`, `#start_child`, and `#finish` to create nested Transactions and spans.

```ruby
require 'sentry-ruby'
Sentry.init do |config|
  config.dsn = 'DSN'
  config.traces_sample_rate = 1
end


spans = []

op = 'testing depth'
description = 'test'
spans << Sentry.start_transaction(name: op, op: op, description: description)

op = "child at depth #{spans.count}"
spans << spans[-1].start_child(op: op, description: description)

op = "child at depth #{spans.count}"
spans << spans[-1].start_child(op: op, description: description)

op = "child at depth #{spans.count}"
spans << spans[-1].start_child(op: op, description: description)

spans.pop.finish until spans.empty?
```

I'd expect this to end up with a transaction 'testing_depth' with a child span 'child at depth 1', and that span to have a child 'child at depth 2' and a third child at 'child at depth 3', etc.

Instead, I just get the 'testing_depth' and 'child at depth 1' but that span shows as having no children.  Is there something I'm missing about how to add child spans?

Thanks!

st0012:
The current `#start_child` implementation on the `Span` and `Transaction` classes are both incorrect, which makes it only possible to create 2-level spans. This PR fixes the issue by making the behavior matching `sentry-python`'s.

Fixes #1372


sentry-ruby/lib/sentry/span.rb
```diff
@@ -19,7 +19,7 @@ class Span
 
 
     attr_reader :trace_id, :span_id, :parent_span_id, :sampled, :start_timestamp, :timestamp, :description, :op, :status, :tags, :data
-    attr_accessor :span_recorder
+    attr_accessor :span_recorder, :transaction
```
rhcarvalho, line 22: Ah nice, today I learned about `attr_accessor` after several "wtf" and bad suggestions to include a missing "@" in front of `transaction` and `span_recorder` down below 😄

sentry-ruby/spec/sentry/span_spec.rb
```diff
@@ -82,7 +82,7 @@
     it "initializes a new child Span" do
       # create subject span and wait for a sec for making time difference
       subject
-      sleep 1
+      sleep 0.3
```
rhcarvalho, line 85: Why do we need the `sleep` here?

sentry-ruby/lib/sentry/transaction.rb
```diff
@@ -51,12 +46,7 @@ def to_hash
 
     def start_child(**options)
```
rhcarvalho, line 47: Looking at [Python](https://github.com/getsentry/sentry-python/blob/19fa43fec5a20b3561a16970ce395c93ac1be57d/sentry_sdk/tracing.py#L218-L241) and [JS](https://github.com/getsentry/sentry-javascript/blob/438ed91bd457f897da89c7728fd074bbef0a0ce1/packages/tracing/src/span.ts#L167-L185), there is probably no reason to have `Transaction.start_child` different than `Span.start_child` -- inheritance should be enough.

rhcarvalho:
Thanks @st0012! I like that I can learn some Ruby from you :)

sentry-ruby/lib/sentry/scope.rb
```diff
@@ -146,8 +146,15 @@ def transaction_name
     end
 
     def get_transaction
-      # transaction will always be the first in the span_recorder
-      span.span_recorder.spans.first if span
+      return unless span
+
+      if span
```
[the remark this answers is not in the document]
st0012: yeah you're right 👍

sentry-ruby/lib/sentry/span.rb
st0012: yeah, `attr_*` helpers could be confusing at first 😂

sentry-ruby/lib/sentry/span.rb
```diff
@@ -129,5 +137,27 @@ def set_data(key, value)
     def set_tag(key, value)
       @tags[key] = value
     end
+
+    protected
+
+    def set_span_recorder(limit = 1000)
```
st0012, line 143: the `protected` statement above makes this method only callable within `Span` or `Transaction` objects. so `Span#dup` can still call it. but users won't be able to call it from their codebase anymore.

sentry-ruby/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 - Fix set_contexts [#1375](https://github.com/getsentry/sentry-ruby/pull/1375) 
 - Use thread variable instead of fiber variable to store the hub [#1380](https://github.com/getsentry/sentry-ruby/pull/1380)
   - Fixes [#1374](https://github.com/getsentry/sentry-ruby/issues/1374)
+- Fix Span/Transaction's nesting issue [#1382](https://github.com/getsentry/sentry-ruby/pull/1382) 
+  - Fixes [#1372](https://github.com/getsentry/sentry-ruby/issues/1372)
```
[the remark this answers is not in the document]
st0012: well I canceled the `4.3.2` release last friday so I can add this to it 😄

sentry-ruby/lib/sentry/span.rb
rhcarvalho: Is the discussion in https://www.rubyguides.com/2018/10/method-visibility/ relevant to us, in particular (see "Private vs Protected Methods") about preferring `private` over `protected` and the performance penalty of using `protected`?

sentry-ruby/lib/sentry/scope.rb
```diff
@@ -146,8 +146,13 @@ def transaction_name
...
     def get_transaction
-      # transaction will always be the first in the span_recorder
-      span.span_recorder.spans.first if span
+      return unless span
+
+      if span.is_a?(Sentry::Transaction)
+        span
+      else
+        span.transaction
+      end
```
rhcarvalho, line 155: Since `transactions` are `spans`, `Transaction.transaction` also points to the correct reference, right?

If so, we could simplify this to 

```suggestion
      span.transaction
```

Or even simplify the whole method to a single line:

```ruby
def get_transaction
  span.transaction if span
end
```

sentry-ruby/lib/sentry/transaction.rb
```diff
@@ -49,20 +45,9 @@ def to_hash
...
-        @span_recorder.add(child_span)
-      end
-
-      child_span
-    end
-
     def deep_dup
       copy = super
-      copy.set_span_recorder
+      copy.set_span_recorder(@span_recorder.max_length)
```
rhcarvalho, line 50: Hmmm this is interesting. What are the use cases for deep cloning a Span / Transaction, why do we implement this?

sentry-ruby/lib/sentry/span.rb
st0012: 1. it's a micro difference in most of the Ruby applications. I think the SDK has other places to improve if we start caring about the pure-Ruby computation performance. (I do watch memory consumption though from time to time)
2. the method is called once per transaction, which in most cases equals once per request. so I don't think the micro performance difference would add up to anything worth worrying.

sentry-ruby/lib/sentry/transaction.rb
st0012: I forgot the original motivation for it. but generally I want to avoid a mutation on copy being shared with its origin, like:

```ruby
      span = Sentry::Transaction.new(sampled: true)
      subject.set_span(span)
      copy = subject.dup

      span.start_child

      # the copy probably shouldn't sees the same recorder change, no?
      expect(copy.span.span_recorder.spans.count).to eq(1)
```

I've written several test cases to prevent sharing mutations between transaction/spans and their copies. so if we decide this is unnecessary and want to drop it, I prefer to do it in another PR.

sentry-ruby/lib/sentry/span.rb
rhcarvalho: Hmm I had to play a little bit with this to understand it.

<details>

```console
$ cat t.rb
class Span
  def do
    set_span_recorder
  end

  private

  def set_span_recorder
    puts "hello"
  end
end

class Transaction < Span
  def tdo
    set_span_recorder
  end
end

span = Span.new
span.do

transaction = Transaction.new
transaction.do
transaction.tdo

$ ruby t.rb
hello
hello
hello
```

</details>

So both `private` and `protected` would work here, and since we never need to call `Span.set_span_recorder` I was curious why you went with `protected`.

---

On trying it out and reading the code more closely, I realized that `set_span_recorder` is really instantiating a new `SpanRecorder` object, assigning the instance to an instance variable `@span_recorder` and adding a reference to the current span.

We call the equivalent method `init_span_recorder` in Python and `initSpanRecorder` in JS. In Python we define it on `Span`, but in JS we define it on `Transaction`, which seems to make more sense to me since we only ever start recording spans from a transaction. In both Python and JS it is the `start_transaction` method on `Hub` that initializes the recorder (I don't particularly like how the method leaks into the public API).

If we copy from Python/JS we'd end up making this method public again 😞 

The alternative would be to keep it as an internal implementation detail, but probably in `transaction.rb`, since we only ever need to create a new instance there. Spans need to get a reference to the private/internal span recorder from their parent.

sentry-ruby/lib/sentry/transaction.rb
```diff
@@ -136,6 +121,13 @@ def finish(hub: nil)
       hub.capture_event(event)
     end
 
+    protected
+
+    def init_span_recorder(limit = 1000)
```
st0012, line 126: @rhcarvalho I've undone the first commit, so it's back to `transaction.rb` now. I also renamed it to match other SDK's names.

sentry-ruby/lib/sentry/transaction.rb
rhcarvalho: Definitely something for another PR, sorry for the distraction. It caught my attention since it is the first time I'm looking in more detail in this SDK.

> the copy probably shouldn't sees the same recorder change, no?

I don't know what I would expect... both possibilities could be okay depending on what it means to duplicate a span.

I understand and noticed how it is used in tests. In production code, nobody should probably be calling `Transaction.dup` or `Span.dup`?! ~~I am genuinely curious for the use cases of duplicating spans~~ I found one.

I see this is used when cloning a `Scope`. Unfortunately I think the behavior here is not clearly specified in the SDK specs. Python and JS do not perform a deep copy, only shallow, the recorder is shared (whether that is useful or not).

rhcarvalho (approved):
Thanks @st0012 !

st0012:
@rhcarvalho thanks for the thorough review! 👍

PR #1384, opened 2021-04-05
Let Transaction constructor take an optional hub argument

st0012:
1. This makes the SDK matches sentry-python's implementation.
2. Having a hub inside a Span/Transaction may be necessary for future  features.
3. This eliminates the optional configuration arguments in different  places, which makes it easier to understand and refactor.

### DIFF
--- a/sentry-ruby/lib/sentry/transaction.rb
+++ b/sentry-ruby/lib/sentry/transaction.rb
@@ -10,19 +10,20 @@ module Sentry
     UNLABELD_NAME = "<unlabeled transaction>".freeze
     MESSAGE_PREFIX = "[Tracing]"
 
-    attr_reader :name, :parent_sampled
+    attr_reader :name, :parent_sampled, :hub
 
-    def initialize(name: nil, parent_sampled: nil, **options)
+    def initialize(name: nil, parent_sampled: nil, hub: Sentry.get_current_hub, **options)
       super(**options)
 
       @name = name
       @parent_sampled = parent_sampled
       @transaction = self
+      @hub = hub
       init_span_recorder
     end
 
-    def self.from_sentry_trace(sentry_trace, configuration: Sentry.configuration, **options)
-      return unless configuration.tracing_enabled?
+    def self.from_sentry_trace(sentry_trace, hub: Sentry.get_current_hub, **options)
+      return unless hub.configuration.tracing_enabled?
       return unless sentry_trace
 
       match = SENTRY_TRACE_REGEXP.match(sentry_trace)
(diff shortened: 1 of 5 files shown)


sentry-ruby/lib/sentry/transaction.rb
```diff
@@ -123,15 +126,16 @@ def set_initial_sample_decision(sampling_context:, configuration: Sentry.configu
     end
 
     def finish(hub: nil)
-      super() # Span#finish doesn't take arguments
+      hub ||= @hub || Sentry.get_current_hub
```
rhcarvalho, line 129: Now we probably don't need `finish` to take a `hub` as argument.

Ah, okay, I see that [in Python the `finish` method takes a `hub`](https://github.com/getsentry/sentry-python/blob/19fa43fec5a20b3561a16970ce395c93ac1be57d/sentry_sdk/tracing.py#L408-L417). In JavaScript it doesn't, and the [actual hub associated with the transaction is defined at initialization time](https://github.com/getsentry/sentry-javascript/blob/438ed91bd457f897da89c7728fd074bbef0a0ce1/packages/tracing/src/transaction.ts#L22-L38).

The important things are:

1. Transactions know about the hub, Spans don't need to/should not know about the hub.
2. The hub association typically should not change during the lifetime of the transaction, so:
  a. we shouldn't need to check `@hub || Sentry.get_current_hub` in multiple places.
  b. methods on span/transaction (not the constructor) should not take a hub as argument.

rhcarvalho (changes requested):
Good change 👍 
I have one suggestion to simplify the relationship between transactions and hub (and spans should not know about hub).

sentry-ruby/lib/sentry/transaction.rb
st0012: in `sentry-python`, even Span holds a hub for adding breadcrumbs. I don't think `sentry-ruby` needs that right now. but I don't want to remove the hub from span and needs to add it back for a feature laster.

https://github.com/getsentry/sentry-python/blob/19fa43fec5a20b3561a16970ce395c93ac1be57d/sentry_sdk/tracing.py#L409-L426

also, I generally use `sentry-python` as a reference, cause I think it should be the most advanced one. but from what I see from your recent comments, `sentry-javascript` seems to be a better sample?

rhcarvalho: They've unfortunately diverged in certain ways, and while Python was first in some parts of the API, JavaScript was first in other parts.

st0012: I see. in that case, I'll keep aligning with `sentry-python` as the language characteristics and applications are more similar.

sentry-ruby/lib/sentry/transaction.rb
```diff
@@ -108,6 +111,8 @@ def set_initial_sample_decision(sampling_context:, configuration: Sentry.configu
     end
 
     def finish(hub: nil)
```
rhcarvalho, line 113: I think callers should never pass a different hub here than the one used to start the transaction, but I'm okay if it is not removed from the API (I'm aware Python has it for historical reasons, committed 2 years ago...)

rhcarvalho (approved):

PR #1395, opened 2021-04-11
Use context for delayed_job's job info

st0012:
This should improve job information's readability.

**Before:**

job info in extra

**After:**

job info in context

### DIFF
--- a/sentry-delayed_job/CHANGELOG.md
+++ b/sentry-delayed_job/CHANGELOG.md
@@ -2,7 +2,10 @@
 
 ## Unreleased
 
+### Features
+
 - Add the `report_after_job_retries` configuration option to only report an exception to Sentry if this is the last job's retry after multiple exceptions. [#1364](https://github.com/getsentry/sentry-ruby/pull/1364)
+- Use context for delayed_job's job info [#1395](https://github.com/getsentry/sentry-ruby/pull/1395)
 
 ## 4.3.1
 
--- a/sentry-delayed_job/lib/sentry/delayed_job/plugin.rb
+++ b/sentry-delayed_job/lib/sentry/delayed_job/plugin.rb
@@ -4,12 +4,16 @@ require "delayed_job"
 module Sentry
   module DelayedJob
     class Plugin < ::Delayed::Plugin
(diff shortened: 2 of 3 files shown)


st0012:
@rhcarvalho is there a check requirement set recently? not all PR would run `sentry-ruby Test` workflow as changes on specific integrations (like this one) don't touch the core SDK and thus doesn't need to run them. but now this PR looks un-mergeable.

rhcarvalho:
>  is there a check requirement set recently? not all PR would run `sentry-ruby Test` workflow as changes on specific integrations (like this one) don't touch the core SDK and thus doesn't need to run them. but now this PR looks un-mergeable.

I'm not aware of any changes. @getsentry/open-source has there been settings changes recently?

BYK:
@st0012 - yup, it was me (w/ consulting to @mitsuhiko). Rolling back some of these to unblock.

BYK:
@st0012 it is good now that said it would be nice to have a static list of baseline checks to block on (I had to remove all).

Also, wouldn't it be safer to run all integration/framework tests for all patches to ensure they don't break anything?

st0012:
> Also, wouldn't it be safer to run all integration/framework tests for all patches to ensure they don't break anything?

It's not necessary in some cases (like this one, because no other integrations relies on or interacts with `sentry-delayed_job`) but I'm not against it. I guess this will make applying checks easier though. Will open a PR later.

I'm aware that the team is trying to make SDKs and projects more consistent for better quality. And I totally support this. But maybe the next time you can open an issue or PM me on discord for a discussion so I'm aware of the change? 🙂

BYK:
>I'm aware that the team is trying to make SDKs and projects more consistent for better quality. And I totally support this. But maybe the next time you can open an issue or PM me on discord for a discussion so I'm aware of the change? 🙂

Yup, totally fair and sorry. I copied the list from a recent PR you merged and the latest build from master and genuinely thought this would essentially be a no-op for you. Will do a better job getting your feedback next time.

st0012:
@BYK no worries 😉 
I've enabled all checks in https://github.com/getsentry/sentry-ruby/pull/1407. I'll merge this one first then you can re-enabled all checks (except `sentry-raven`'s) 😄

Issue #1383, opened 2021-04-05
Multiple Unreported Events 


abhinavmsra:
**Describe the bug**
While most errors are still reported to Sentry, some bugs are not tracked. For untracked errors, following is the stacktrace

```
FATAL -- 
NoMethodError (undefined method `shopname' for nil:NilClass):  
app/models/transaction.rb:132:in `assign_shopname'
app/controllers/api/v1/transactions_controller.rb:69:in `create'
ERROR -- sentry: Event sending failed: undefined method `new' for NilClass:Class
ERROR -- sentry: Unreported Event: NoMethodError: undefined method `shopname' for nil:NilClass

---- 

NoMethodError (undefined method `classify' for nil:NilClass):
app/policies/transaction_policy.rb:20:in `resolve'
app/controllers/api/v1/transactions_controller.rb:30:in `block (2 levels) in index'
app/controllers/api/v1/transactions_controller.rb:17:in `index'
ERROR -- sentry: Event sending failed: undefined method `new' for NilClass:Class
ERROR -- sentry: Unreported Event: NoMethodError: undefined method `classify' for nil:NilClass
```

**To Reproduce**
I could not figure out how to reproduce specific untracked errors.

**Expected behavior**
ALL errors to be reported 

**Actual behavior**
Some errors are missed

**Environment**
 - Ruby Version: 3
 - SDK Version:
 ```
sentry-rails (4.3.3)
      railties (>= 5.0)
      sentry-ruby-core (~> 4.3.0)
    sentry-ruby (4.3.1)
      concurrent-ruby (~> 1.0, >= 1.0.2)
      faraday (>= 1.0)
      sentry-ruby-core (= 4.3.1)
    sentry-ruby-core (4.3.1)
      concurrent-ruby
      faraday
    sentry-sidekiq (4.3.0)
      sentry-ruby-core (~> 4.3.0)
 ```
 - Integration Versions (if any):
  - Rails 6.1.3.1, Sidekiq 6.2.0

```ruby
Sentry.init do |config|
    filter = ActiveSupport::ParameterFilter.new(Rails.application.config.filter_parameters)

    config.dsn = Rails.application.credentials.sentry_dsn
    config.breadcrumbs_logger = [:active_support_logger]
    config.traces_sample_rate = 0.2

    config.before_send = ->(event, _hint) do
      request = event.request
      if request
        request.data = filter.filter(request.data)
      end
      event
    end
  end
```

st0012:
[...]

PR #1400, opened 2021-04-16
Support config.debug configuration option

### LINKED ISSUE #1297: Support debug option
The debug option should make debugging SDK problems more easily. So when it is enabled, the SDK should output detail information that normally would be noise to users, like:
1. Full backtrace of the errors it encounters.
2. Full payload of failed events.

It also ***should not** change how the SDK works, so it doesn't complicate the problem.

st0012:
This PR adds the new `debug` config option:

```ruby
config.debug = true # default is false
```

As of now, SDK related errors are only logged with their error messages. 

For example, assume we have this `before_send` callback:

```ruby
    foo
  end
```

The SDK only logs this

```
Event sending failed: undefined local variable or method `foo' for main:Object
```

This is inconvenient for our users to debug their customization on the SDK. It also makes determining/reporting SDK issues harder (see https://github.com/getsentry/sentry-ruby/issues/1383#issuecomment-813143453 for example).

So with when the new `config.debug` option set with `true`, it will log SDK errors with backtrace, just like we normally treat errors in our applications.

```
Event sending failed: undefined local variable or method `foo' for main:Object
/Users/st0012/projects/sentry-ruby/sentry-rails/examples/rails-6.0/config/initializers/sentry.rb:10:in `block (2 levels) in <main>'
/Users/st0012/projects/sentry-ruby/sentry-ruby/lib/sentry/client.rb:78:in `send_event'
/Users/st0012/projects/sentry-ruby/sentry-ruby/lib/sentry/client.rb:102:in `block in dispatch_background_event'
```

Closes #1297

### DIFF
--- a/sentry-ruby/CHANGELOG.md
+++ b/sentry-ruby/CHANGELOG.md
@@ -18,6 +18,10 @@ The SDK now records a new `net.http` breadcrumb whenever the user makes a reques
 
 <img width="869" alt="net http breadcrumb" src="https://user-images.githubusercontent.com/5079556/114298326-5f7c3d80-9ae8-11eb-9108-222384a7f1a2.png">
 
+- Support config.debug configuration option [#1400](https://github.com/getsentry/sentry-ruby/pull/1400)
+
+It'll determine whether the SDK should run in the debugging mode. Default is `false`. When set to true, SDK errors will be logged with backtrace.
+
 ### Refactorings
 
 - Let Transaction constructor take an optional hub argument [#1384](https://github.com/getsentry/sentry-ruby/pull/1384)
--- a/sentry-ruby/lib/sentry/client.rb
+++ b/sentry-ruby/lib/sentry/client.rb
@@ -38,7 +38,7 @@ module Sentry
 
       event
     rescue => e
-      log_error("Event capturing failed: #{e.message}")
+      log_error("Event capturing failed", e)
       nil
     end
 
@@ -88,7 +88,7 @@ module Sentry
       event
     rescue => e
       loggable_event_type = (event_type || "event").capitalize
-      log_error("#{loggable_event_type} sending failed: #{e.message}")
+      log_error("#{loggable_event_type} sending failed", e)
       log_info("Unreported #{loggable_event_type}: #{Event.get_log_message(event.to_hash)}")
       raise
     end
@@ -114,7 +114,7 @@ module Sentry
       end
     rescue => e
       loggable_event_type = event_hash["type"] || "event"
-      log_error("Async #{loggable_event_type} sending failed: #{e.message}")
+      log_error("Async #{loggable_event_type} sending failed", e)
       send_event(event, hint)
     end
   end
--- a/sentry-ruby/lib/sentry/client.rb
+++ b/sentry-ruby/lib/sentry/client.rb
@@ -38,7 +38,7 @@ module Sentry
 
       event
     rescue => e
-      log_error("Event capturing failed", e)
+      log_error("Event capturing failed", e, debug: configuration.debug)
       nil
     end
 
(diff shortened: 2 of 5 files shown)


sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -72,6 +72,10 @@ class Configuration
     # RACK_ENV by default.
     attr_reader :environment
 
+    # Whether the SDK should run in the debugging mode. Default is false.
+    # If set to true, the SDK will log every error with full backtrace
```
rhcarvalho, line 76: Perhaps here we don't need to promise what debug mode does in such detail ("SDK will log every error with full backtrace"), leaving some possibility to change it later. In any case, we are probably not fully specifying all the consequences of debug mode.

sentry-ruby/spec/sentry/configuration_spec.rb
```diff
@@ -238,6 +238,19 @@
           end
         end
       end
+
+      context "when having an error detecting the release" do
```
rhcarvalho, line 242: Is this related to the new debug mode?

sentry-ruby/CHANGELOG.md
```diff
@@ -18,6 +18,10 @@ The SDK now records a new `net.http` breadcrumb whenever the user makes a reques
 
 <img width="869" alt="net http breadcrumb" src="https://user-images.githubusercontent.com/5079556/114298326-5f7c3d80-9ae8-11eb-9108-222384a7f1a2.png">
 
+- Support config.debug configuration option [#1400](https://github.com/getsentry/sentry-ruby/pull/1400)
+
+It'll determine whether the SDK should run in the debugging mode. Default is `false`. When set to `true`, the SDK will log every error with full backtrace.
```
rhcarvalho, line 23: Perhaps here we don't need to promise what debug mode does in such detail (see below)

rhcarvalho (commented):
Maybe I'm misreading this, but it seems that regardless of `debug = true/false` we're always printing full event payloads to _STDOUT_?

sentry-ruby/spec/sentry/configuration_spec.rb
st0012: @rhcarvalho no, it's just an uncovered scenario I found during the earlier commits 😅

sentry-ruby/lib/sentry/client.rb
```diff
@@ -88,8 +88,10 @@ def send_event(event, hint = nil)
       event
     rescue => e
       loggable_event_type = (event_type || "event").capitalize
-      log_error("#{loggable_event_type} sending failed: #{e.message}")
-      log_error("Unreported #{loggable_event_type}: #{Event.get_log_message(event.to_hash)}")
+      log_error("#{loggable_event_type} sending failed", e, debug: configuration.debug)
+
+      event_info = Event.get_log_message(event.to_hash)
```
st0012, line 93: @rhcarvalho this is not full event payload. you can see the old code does the exact same thing and I just moved it to a new line.
what it does is extracting the event's message attribute **or** the exception's name and no more. here's an example

```
Unreported Event: ZeroDivisionError: divided by 0
```

rhcarvalho: Aha, sorry, I was mislead by the name 😢

sentry-ruby/lib/sentry/utils/logging_helper.rb
```diff
@@ -1,7 +1,12 @@
 module Sentry
   module LoggingHelper
-    def log_error(message)
-      logger.error(LOGGER_PROGNAME) { message }
+    def log_error(message, exception, debug: false)
```
rhcarvalho, line 3: Was this considered public API? If so, the new required argument would probably be a breaking change.

st0012: this is not meant to be a public API and actually doesn't exist in previous versions. it's introduced in a previous refacotring.

rhcarvalho (approved):
Thanks @st0012 !

PR #1401, opened 2021-04-16
Make hub a required argument for Transaction constructor

### LINKED ISSUE #1393: Make `hub` a required argument in Transaction's constructor
This issue is more about the general design principle on all Sentry SDKs instead of just `sentry-ruby`.  I want to discuss the necessity of falling back to `Sentry.get_current_hub` when the `hub` is not provided.

In `sentry-python`, it's done in `Transaction.finish`:

https://github.com/getsentry/sentry-python/blob/19fa43fec5a20b3561a16970ce395c93ac1be57d/sentry_sdk/tracing.py#L409-L417

In `sentry-javascript`, it looks to be done in the constructor?

https://github.com/getsentry/sentry-javascript/blob/fbdc8753ab5e54fb8c83101de78a805ec7335522/packages/tracing/src/transaction.ts#L22-L46

And currently, `sentry-ruby` does it like this

```ruby
def initialize(name: nil, parent_sampled: nil, hub: Sentry.get_current_hub, **options)
```

Shouldn't we make the `hub` argument a hard requirement when initializing a `Transaction` object? Which means, we should never need to fallback to the current active hub. 

I think we'd agree that a `Transaction` object should never be initialized manually by the user, which is also stated in `sentry-javascript`'s comment:

```javascript
   * This constructor should never be called manually. Those instrumenting tracing should use
   * `Sentry.startTransaction()`, and internal methods should use `hub.startTransaction()`.
```

If that's the case, `hub.startTransaction()` should always be able to provide a `hub` to the transaction, and thus never need to fallback to `getCurrentHub()`.

So what I'm proposing are:

1. Stop making the `hub` argument optional for the `Transaction` constructor.
2. Remove all the fallback logic from the constructor or `Transaction.finish`. Because it must present and must be a `Hub` instance.

Wdyt?

st0012:
This implements the change proposed in #1393.

Closes #1393

### DIFF
--- a/sentry-ruby/CHANGELOG.md
+++ b/sentry-ruby/CHANGELOG.md
@@ -23,6 +23,7 @@ The SDK now records a new `net.http` breadcrumb whenever the user makes a reques
 - Let Transaction constructor take an optional hub argument [#1384](https://github.com/getsentry/sentry-ruby/pull/1384)
 - Introduce LoggingHelper [#1385](https://github.com/getsentry/sentry-ruby/pull/1385)
 - Raise exception if a Transaction is initialized without a hub [#1391](https://github.com/getsentry/sentry-ruby/pull/1391)
+- Make hub a required argument for Transaction constructor [#1401](https://github.com/getsentry/sentry-ruby/pull/1401) 
 
 ## 4.3.2
 
--- a/sentry-ruby/lib/sentry/transaction.rb
+++ b/sentry-ruby/lib/sentry/transaction.rb
@@ -14,16 +14,13 @@ module Sentry
 
     attr_reader :name, :parent_sampled, :hub, :configuration, :logger
 
-    def initialize(name: nil, parent_sampled: nil, hub: Sentry.get_current_hub, **options)
+    def initialize(name: nil, parent_sampled: nil, hub:, **options)
       super(**options)
 
       @name = name
       @parent_sampled = parent_sampled
       @transaction = self
       @hub = hub
-
-      raise Sentry::Error.new("please initialize the SDK with Sentry.init before initializing a Transaction") unless @hub
-
       @configuration = hub.configuration
       @logger = configuration.logger
       init_span_recorder
--- a/sentry-ruby/lib/sentry/transaction.rb
+++ b/sentry-ruby/lib/sentry/transaction.rb
@@ -112,6 +112,15 @@ module Sentry
     end
 
     def finish(hub: nil)
+      if hub
+        log_warn(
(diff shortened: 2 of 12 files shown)


sentry-ruby/CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ The SDK now records a new `net.http` breadcrumb whenever the user makes a reques
 - Let Transaction constructor take an optional hub argument [#1384](https://github.com/getsentry/sentry-ruby/pull/1384)
 - Introduce LoggingHelper [#1385](https://github.com/getsentry/sentry-ruby/pull/1385)
 - Raise exception if a Transaction is initialized without a hub [#1391](https://github.com/getsentry/sentry-ruby/pull/1391)
+- Make hub a required argument for Transaction construbtor [#1401](https://github.com/getsentry/sentry-ruby/pull/1401)
```
rhcarvalho, line 26:
```suggestion
- Make hub a required argument for Transaction constructor [#1401](https://github.com/getsentry/sentry-ruby/pull/1401) 
```

st0012: 🤦‍♂️

sentry-ruby/lib/sentry/transaction.rb
```diff
@@ -14,16 +14,13 @@ class Transaction < Span
 
     attr_reader :name, :parent_sampled, :hub, :configuration, :logger
 
-    def initialize(name: nil, parent_sampled: nil, hub: Sentry.get_current_hub, **options)
+    def initialize(name: nil, parent_sampled: nil, hub:, **options)
```
rhcarvalho, line 17: @st0012 I'm learning Ruby from you, thanks!

I just tried this out on `irb`. So if `hub` is not set, there's an error:

```
ArgumentError (missing keyword: hub)
```

Nice!

Not in the scope of this PR, but wanted to share that this could as well be the behavior for `name`. Transactions must have a name. Deciding if it is required or not in the constructor is debatable though, sometimes you don't know the name or need to update it later before sending the transaction. In some languages we write a debug message if the name is not set right before sending the transaction to Sentry.

If a transaction is sent without a name it gets a default value, but that's not very useful. The transaction name is what is used to group the many different traces.

rhcarvalho (approved):

sentry-ruby/lib/sentry/transaction.rb
st0012: thanks for pointing out the potential issue. I'll keep that in mind 🙂

PR #1402, opened 2021-04-16
Add the third tracing state

st0012:
- `rate == 0` - Tracing enabled. Rejects all locally created transactions but  respects sentry-trace.
- `1 > rate > 0` - Tracing enabled. Samples locally created transactions  with the rate and respects sentry-trace.
- `rate < 0` or `rate > 1` - Tracine disabled.

### DIFF
--- a/sentry-ruby/CHANGELOG.md
+++ b/sentry-ruby/CHANGELOG.md
@@ -45,6 +45,11 @@ The SDK now records a new `net.http` breadcrumb whenever the user makes a reques
 
 It'll determine whether the SDK should run in the debugging mode. Default is `false`. When set to true, SDK errors will be logged with backtrace.
 
+- Add the third tracing state [#1402](https://github.com/getsentry/sentry-ruby/pull/1402)
+  - `rate == 0` - Tracing enabled. Rejects all locally created transactions but  respects sentry-trace.
+  - `1 > rate > 0` - Tracing enabled. Samples locally created transactions  with the rate and respects sentry-trace.
+  - `rate < 0` or `rate > 1` - Tracing disabled.
+
 ### Refactorings
 
 - Let Transaction constructor take an optional hub argument [#1384](https://github.com/getsentry/sentry-ruby/pull/1384)
--- a/sentry-ruby/spec/sentry/transaction_spec.rb
+++ b/sentry-ruby/spec/sentry/transaction_spec.rb
@@ -24,7 +24,7 @@ RSpec.describe Sentry::Transaction do
       Sentry.configuration
     end
 
-    context "when tracing is enabled" do
+    context "when tracing is enabled (> 0)" do
       before do
         configuration.traces_sample_rate = 1.0
       end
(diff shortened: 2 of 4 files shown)


sentry-ruby/CHANGELOG.md
```diff
@@ -18,6 +18,11 @@ The SDK now records a new `net.http` breadcrumb whenever the user makes a reques
 
 <img width="869" alt="net http breadcrumb" src="https://user-images.githubusercontent.com/5079556/114298326-5f7c3d80-9ae8-11eb-9108-222384a7f1a2.png">
 
+- Add the third tracing state [#1402](https://github.com/getsentry/sentry-ruby/pull/1402)
+  - `rate == 0` - Tracing enabled. Rejects all locally created transactions but  respects sentry-trace.
+  - `1 > rate > 0` - Tracing enabled. Samples locally created transactions  with the rate and respects sentry-trace.
+  - `rate < 0` or `rate > 1` - Tracine disabled.
```
rhcarvalho, line 24: This PR doesn't reference anything, so I don't know what was the motivation behind the change.

I can point to https://github.com/getsentry/develop/issues/265

What we've been converging towards when clearing up inconsistencies is that `traces_sample_rate` > 0 or `traces_sampler != nil` enables tracing.

`traces_sample_rate` not explicitly set, `nil`, means no transactions are sent. The `sentry-trace` header is only looked into if tracing is enabled as per definition above.

https://develop.sentry.dev/sdk/performance/#tracessamplerate `traces_sample_rate` out of the [0, 1] range is undefined behavior.

st0012: sorry I should've added some references here, but I thought it's already a consensus. here are the 2 drivers:
- I saw `sentry-python` defined it as [invalid](https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/tracing.py#L698-L704) 
- @lobsterkatie also wrote [a test case](https://github.com/getsentry/sentry-ruby/pull/1328/files#diff-ca2b59d3ec608f0ce8907b919472b48f2b603b28343985143e9f1ef364b42c20R36-R43) to declare that `rate > 1` is invalid and should disable tracing.

rhcarvalho:
```suggestion
  - `rate < 0` or `rate > 1` - Tracing disabled.
```

sentry-ruby/spec/sentry/transaction_spec.rb
```diff
@@ -47,11 +47,28 @@
...
+        expect(child_transaction.parent_sampled).to eq(true)
+        # doesn't set the sampled value
+        expect(child_transaction.sampled).to eq(nil)
+        expect(child_transaction.op).to eq("child")
+      end
+    end
+
+    context "when tracing is disabled" do
+      before do
+        configuration.traces_sample_rate = false
```
rhcarvalho, line 69: This value should typically be numeric or `nil`. I suspect the default value is `nil`? If so I'd use that here.

rhcarvalho (approved):
Few suggestions, generally ltgm

PR #1404, opened 2021-04-18
Un-sample asset/static file requests in a less-hacky way

st0012:
**Note: This change only affects applications that enables file server.**

Right now, `sentry-rails` is smart enough to un-sample transactions of static file/assets requests.

But users may still see tracing logs in those requests like:

```
[Tracing] Starting <rails.request> transaction </favicon.ico>
```

This is because the current implementation un-samples those transactions after the initial sampling decision has been made. 
The approach's downsides are:

1. It requires hacking the `FileHandler` class.
2. It generates annoying and sometimes confusing tracing logs, like the above one, for requests that are always un-sampled.

The reason behind it is the order of middlewares:

```
request 
  -> Sentry::CaptureExceptions (starts the transaction) 
  -> ActionDispatch::Static (checks if the request is for static files) 
      -> A patch from the SDK then manually un-sample the transaction if the request is for static files
```

As you can see, when starting the transaction, the SDK can't possibly know if the request is for static files unless `Sentry::CaptureExceptions` also knows how to check static file requests. 

Well, this is what we do in this PR. The upsides are:

1. It doesn't require hacking any Rails components.
2. It uses the SDK's sampling mechanism to properly un-sample the transaction. So no additional logging would be made.

It still has some downsides though:

1. To check if the request is for static files, we need to replicate some private behaviors of the `ActionDispatch::FileHandler` class.
2. Because the replication, we're doing file checking twice in a request.
3. The file checking logic is prone to change, so we need to maintain different implementations for different versions of Rails.


st0012:
Closing this because I think #1405 is a better solution to the same problem.

PR #1407, opened 2021-04-19
Always run builds for all integrations

st0012:
Although this may not help catching issues, it will make it easier to set merge blockers.


BYK:
Set all checks that ran on this PR as required. LMK if this causes any issues please.

PR #1415, opened 2021-04-23
Check Scope#set_context's value argument

st0012:
A context's value should always be an object when sent to Sentry. So the SDK should make sure users pass the right type of value (A Ruby Hash).


rhcarvalho (approved):
👍

PR #1418, opened 2021-04-27
Bump integration dependency

st0012:


st0012:
@BYK can you help me remove the `Compare memory allocation`  jobs from the list? they're not really used and will be removed soon.

BYK:
@st0012 I think somebody beat me to it 😁

st0012:
@BYK ah right! sorry I should've updated that in the comment 😓

PR #1421, opened 2021-04-28
Disable tracing if events are not allowed to be sent

st0012:
Users can enable/disable the SDK's event sending with `enabled_environments` config option. And when the current environment doesn't match the `enabled_environments`, the SDK will not send any events. But currently, the SDK would still start transactions under such condition.

Even though those transactions would not be sent later (stopped by `Client#send_event`), it'd still create sampling logs, which are noisy and confusing to users.

So this commit fixes the issue by making tracing activation aware of the event sending state.

### DIFF
--- a/sentry-ruby/CHANGELOG.md
+++ b/sentry-ruby/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Bug Fixes
+
+- Disable tracing if events are not allowed to be sent [#1421](https://github.com/getsentry/sentry-ruby/pull/1421)
+
 ## 4.4.0-beta.0
 
 ### Features
--- a/sentry-ruby/lib/sentry/configuration.rb
+++ b/sentry-ruby/lib/sentry/configuration.rb
@@ -293,7 +293,7 @@ module Sentry
     end
 
     def tracing_enabled?
-      !!((@traces_sample_rate && @traces_sample_rate >= 0.0 && @traces_sample_rate <= 1.0) || @traces_sampler)
+      !!((@traces_sample_rate && @traces_sample_rate >= 0.0 && @traces_sample_rate <= 1.0) || @traces_sampler) && sending_allowed?
     end
 
     def stacktrace_builder
--- a/sentry-ruby/spec/sentry/configuration_spec.rb
+++ b/sentry-ruby/spec/sentry/configuration_spec.rb
@@ -2,39 +2,66 @@ require 'spec_helper'
 
 RSpec.describe Sentry::Configuration do
   describe "#tracing_enabled?" do
-    it "returns false by default" do
-      expect(subject.tracing_enabled?).to eq(false)
+    context "when sending not allowed" do
+      before do
+        allow(subject).to receive(:sending_allowed?).and_return(false)
+      end
+
+      context "when traces_sample_rate > 0" do
+        it "returns false" do
+          subject.traces_sample_rate = 0.1
+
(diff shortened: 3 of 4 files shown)


sentry-ruby/spec/sentry/transaction_spec.rb
```diff
@@ -173,6 +173,21 @@
...
 
+      context "when event reporting is not enabled" do
+        before do
+          Sentry.configuration.enabled_environments = ["production"]
+        end
+
+        it "sets @sampled to false and return" do
+          transaction = described_class.new(sampled: true, hub: Sentry.get_current_hub)
+          transaction.set_initial_sample_decision(sampling_context: {})
+          expect(transaction.sampled).to eq(false)
```
rhcarvalho, line 184: Thought about this for a while. At first I thought neither `true` nor `false` fit here, because both of those mean a sampling decision was actually made, in which case `nil` would be more appropriate.

However, the [sampling rule](https://develop.sentry.dev/sdk/performance/#sampling) is that explicit sample decision always has precedence:

> 1. If a sampling decision is passed to startTransaction (startTransaction({name: "my transaction", sampled: true})), that decision will be used, regardlesss of anything else

So in this particular example, since the sampling decision was passed to `.new`, it should not have changed.

In any case, tracing is effectively disabled, `startTransaction` should not instantiate transactions.

sentry-ruby/lib/sentry/transaction.rb
```diff
@@ -64,7 +64,7 @@ def deep_dup
     end
 
     def set_initial_sample_decision(sampling_context:)
-      unless configuration.tracing_enabled?
+      unless configuration.tracing_enabled? && configuration.sending_allowed?
         @sampled = false
         return
       end
```
rhcarvalho, line 70: Tracing being enabled or not should probably not interfere with the sampling decision.

We don't have an equivalent of `enabled_environments` and `sending_allowed` in other SDKs, so I'm thinking here if it makes sense to have `tracing_enabled == true` and `sending_allowed == false`, or if `sending_allowed == false` should imply `tracing_enabled = false`.

rhcarvalho (changes requested):
I think I generally understood the problem this is trying to solve and the current solution addresses the problem. 👍 

However there's one violation of the sampling rules, and we need to sync on how the config of the SDK works with regards to tracing.

sentry-ruby/lib/sentry/transaction.rb
st0012: I didn't expect the later option to work without breaking tons of test cases, but I was wrong. So I've rewritten the commits and the description. Thanks for the suggestion 👍

rhcarvalho (approved):

sentry-ruby/spec/sentry/transaction_spec.rb
st0012: I've addressed the decision making issue in https://github.com/getsentry/sentry-ruby/pull/1423

rhcarvalho: Thanks @st0012. I've chatted with the team and looked at code in other SDKs, I retract my comment for now.
Python and JS do exactly as Ruby does now, despite not adhering 100% to the spec as written.
So instead of going back and forth with this, let's just leave it as it is for now, and revisit it later as we rethink tracing as a whole.

st0012: Yeah, I implemented this behavior based on what I saw in `sentry-python`. I thought the spec you posted was to correct that.

PR #1423, opened 2021-04-29
set_initial_sample_decision should always respect the original decision

st0012:
This is to address the issue found in this comment https://github.com/getsentry/sentry-ruby/pull/1421#discussion_r622870135


rhcarvalho:
See https://github.com/getsentry/sentry-ruby/pull/1421#discussion_r623411787

Let's not change this right now.

PR #1426, opened 2021-05-03
Fix original user context mutated in block

larsklevan:
using `.merge!` in `user_context` resulted in the original context being mutated. The "sets user context only in the block" spec appeared to pass but the line `expect(subject.context.user).to eq previous_user_context` was misleading because `previous_user_context` had been mutated. So it looks like it's asserting `expect(subject.context.user).to eq(id: 9999)` but it was actually asserting `expect(subject.context.user).to eq(id: 1)` which was incorrect.

I added a second spec for nested calls but that wouldn't be necessary to demonstrate the fix.


larsklevan:
@st0012 added a CHANGELOG entry, let me know if that format looks ok!

PR #1390, opened 2021-04-09
ci(release): Fix changelog entry paths in craft config

BYK:
Since we do per-directory releases in this repo, we need to tell Craft where the changelog for each project is (otherwise it just looks at the repo root). This is why our GitHub releases were lacking release notes. This PR also moves the GitHub target to the last to ensure all main targets are published before creating a GitHub release tag.

PR #1435, opened 2021-05-07
fix: Fix changelog config in craft

BYK:
This is a fix up to https://github.com/getsentry/sentry-ruby/pull/1390#issuecomment-834143608

### DIFF
--- a/sentry-delayed_job/.craft.yml
+++ b/sentry-delayed_job/.craft.yml
@@ -3,6 +3,7 @@ github:
     owner: getsentry
     repo: sentry-ruby
 changelogPolicy: simple
+changelog: sentry-delayed_job/CHANGELOG
 preReleaseCommand: ruby ../.scripts/bump-version.rb
 releaseBranchPrefix: release-sentry-delayed_job
 statusProvider:
@@ -17,4 +18,3 @@ targets:
           canonical: 'gem:sentry-delayed_job'
     - name: github
       tagPrefix: sentry-delayed_job-v
-      changelog: sentry-delayed_job/CHANGELOG
--- a/sentry-rails/.craft.yml
+++ b/sentry-rails/.craft.yml
@@ -3,6 +3,7 @@ github:
     owner: getsentry
     repo: sentry-ruby
 changelogPolicy: simple
+changelog: sentry-rails/CHANGELOG.md
 preReleaseCommand: ruby ../.scripts/bump-version.rb
 releaseBranchPrefix: release-sentry-rails
 statusProvider:
@@ -17,4 +18,3 @@ targets:
           canonical: 'gem:sentry-rails'
     - name: github
       tagPrefix: sentry-rails-v
-      changelog: sentry-rails/CHANGELOG.md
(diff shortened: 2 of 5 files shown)


st0012 (approved):
Thanks! I'll make a release today so can test this soon 👍

st0012:
@BYK the prepare failed: https://github.com/getsentry/sentry-ruby/runs/2526077118?check_suite_focus=true

```
[9:34:10 AM] Error: Changelog does not exist: "sentry-ruby/CHANGELOG.md"
```

I think the release workflow enters the sub-folder before running the command, so the `sentry-*` folder name is not actually needed? But if that's the case, there's no need to specify the `changelog` attribute after all? However, not having the `changelog` didn't work either 😕

BYK:
Ehh, I see the issue now. Prepare operates on the `.craft.yml` directory whereas publish fetches the changelog directly from the repo, meaning it looks at the root dir. That said I'm surprised that prepare complains about the missing changelog. Let me check the code. You can revert this patch in the meantime to get the release going.

BYK:
Okay, I see the reason now. I'll write a patch for Craft so when it is released, this should be good to go.

You may still need to define the changelog files as some have the `md` extension and some doesn't (or rather, fix that inconsistency? :)).

st0012:
@BYK I think the inconsistency only exists in `sentry-delayed_job/.craft.yml`. But before I make any change again, how will the changelog pulling work now though? Like this?
- prepare - read what's specified in the `changelog` attribute from **project root**.
- publish - read what's specified in the `changelog` attribute from **project root**.

BYK:
@st0012 nah, I'll make them relative to `.craft.yml` file as that's more sensible. During `publish`, we'll have to calculate the path from `.git` root which should be easy.

st0012:
@BYK why can't `prepare` pass the absolute changelog path (no matter how it's computed) to the issue or wherever the intermedium state is stored, so this chagnelog path calculation only needs to be done once?
I think this will make debugging easier. Because if they're done separately, it'll always require the SDK developer and Sentry's OSS team to work together to debug the issue from end to end.

BYK:
@st0012 craft is built around the assumption that it runs in the same directory as `.craft.yml` which also is the project root, which also is the `.git` root.

We need to break that assumption to support monorepos like this. That's why we are keeping the `.craft.yml` being the "project root" but don't assume it is the `.git` root, hence calculate the path for the changelog when we fetch it from the git repo on the fly. From a project maintainer, this is transparent: all they do is define everything relative to the project root and hence `.craft.yml` file.

st0012:
@BYK I see. Is `sentry-ruby` the only monorepo that releases packages separately from each sub-folder?
I should mention that we're planning to:
1. unify all gem's release and have only 1 change log at the repo root
2. but for `sentry-raven`, it'll still be released on its own and have its changelog under `sentry-raven/`

Will this change cause any problem with the craft update you're making?

st0012:
@BYK I see, thanks 👍 I'll wait for the craft update and make the release next week then.

PR #1444, opened 2021-05-14
Add sentry_trace connection example

st0012:
- Parent: sentry-rails' example app
- Child: sentry-ruby's sinatra example

Usage:

1. `$ cd sentry-rails/examples/rails-6.0`
2. `$ bundle exec rails s`
3. `$ cd sentry-ruby/examples/sinatra`
4. `$ bundle exec ruby app.rb`
5. Visit `http://localhost:3000/connect_trace` in browser

connect sentry trace

### DIFF
--- a/sentry-rails/examples/rails-6.0/app/controllers/welcome_controller.rb
+++ b/sentry-rails/examples/rails-6.0/app/controllers/welcome_controller.rb
@@ -5,6 +5,18 @@ class WelcomeController < ApplicationController
     1 / 0
   end
 
+  def connect_trace
+    transaction = Sentry.get_current_scope.get_transaction
+    # see the sinatra example under the `sentry-ruby` folder
+    uri = URI("http://localhost:4567/connect_trace")
+    http = Net::HTTP.new(uri.host, uri.port)
+    request = Net::HTTP::Get.new(uri.request_uri)
+    request["SENTRY_TRACE"] = transaction.to_sentry_trace
+    response = http.request(request)
+
+    render plain: response.code
+  end
+
   def view_error
   end
 
--- a/sentry-rails/examples/rails-6.0/config/routes.rb
+++ b/sentry-rails/examples/rails-6.0/config/routes.rb
@@ -3,6 +3,7 @@ Rails.application.routes.draw do
   get '500', :to => 'welcome#report_demo'
   root to: "welcome#index"
 
+  get 'connect_trace', to: 'welcome#connect_trace'
   get 'view_error', to: 'welcome#view_error'
   get 'worker_error', to: 'welcome#worker_error'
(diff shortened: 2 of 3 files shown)


sentry-rails/examples/rails-6.0/app/controllers/welcome_controller.rb
```diff
@@ -5,6 +5,18 @@ def index
     1 / 0
   end
 
+  def connect_trace
+    transaction = Sentry.get_current_scope.get_transaction
+    # see the sinatra example under the `sentry-ruby` folder
+    uri = URI("http://localhost:4567/connect_trace")
+    http = Net::HTTP.new(uri.host, uri.port)
+    request = Net::HTTP::Get.new(uri.request_uri)
+    request["SENTRY_TRACE"] = transaction.to_sentry_trace
```
rhcarvalho, line 14: @st0012 if this line is required, then it explains the current difficulty in https://github.com/getsentry/sentry-ruby/issues/1411 as this would be a required manual step for trace propagation.

We try to make HTTP clients automatically propagate traces with auto instrumentation, perhaps that's missing in Ruby?

rhcarvalho (commented):
Thanks for the example!

sentry-rails/examples/rails-6.0/app/controllers/welcome_controller.rb
st0012: I'm not aware of this requirement in the backend SDK 🤔  can you point me the implementation in `sentry-python` and `sentry-go`?

rhcarvalho: Python: https://github.com/getsentry/sentry-python/blob/51987c57157102bbd32e1e7b084c26f4dc475d86/sentry_sdk/integrations/stdlib.py#L57-L112

JS Node: https://github.com/getsentry/sentry-javascript/blob/f7fc73313cb1f602dfbde856e116bb74f39dde81/packages/node/src/integrations/http.ts#L109-L123

JS Browser: https://github.com/getsentry/sentry-javascript/blob/13eea94a4da4334dd79f829a2657c03dbb50f780/packages/tracing/src/browser/request.ts#L87-L264

Go doesn't have this yet (know limitation), and when we do it won't be like Python and Ruby because we can't monkey-patch libraries.

But the idea is that we automatically create spans for outgoing requests and propagate the trace header `sentry-trace` with information on the current span (the one that was just created for the request).

st0012: @rhcarvalho I see, thanks. I'll merge this PR now and then implemented the propagation. after that's done I'll update the example again.

Issue #1449, opened 2021-05-21
Release gems in batch


st0012:
## Current Workflow & Problems

Currently, all gems are all prepared and released individually. It's great for releasing patch version individually, but it also makes releasing minor versions complicated. Here's the general workflow:

1. Prepare `sentry-ruby`'s release
2. Wait for `sentry-ruby` to be published
3. Update integration gems' (like `sentry-rails`) `sentry-ruby-core` dependency version.
  - We can't do this before the release branch of `sentry-ruby` is merged to master.
4. Prepare integration gems' release individually
5. Wait for them to be published individually

Also, we can't have a shared changelog under the current releasing structure due to potential version collision.

## New Workflow

Definition of the new workflow:

- All gems (except `sentry-raven`) will always be released under the same version.
  - For example, to release `sentry-sidekiq` to `4.4.4`, other `sentry-*` gems will also be bumped and released with version `4.4.4`, regardless they have change or not.
- `sentry-raven` will still be released solely.
- New gems will firstly be released individually before joining the batch releasing process.

And the new release process will be:

1. Prepare batch release
2. Publish all of them at once 🙂

st0012:
cc @HazAT @BYK

HazAT:
Cool stuff, also we should have one changelog in release with different sections like we have in JS monorepo https://github.com/getsentry/sentry-javascript/releases/tag/6.4.1

So our users always can look at the release and changelog to understand what changed.
This will bring us a large step closer to making things easier for our users to version.

(The only downside is potential no-op releases for packages which is something we def. can live with)

st0012:
@HazAT yeah that's part of the main drive behind this decision 😄

st0012:
This is now completed with the help from @BYK 😉

PR #1450, opened 2021-05-21
Batch release prepare

st0012:
First part of #1449

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -1,0 +1,33 @@
+minVersion: '0.13.2'
+github:
+    owner: getsentry
+    repo: sentry-ruby
+changelogPolicy: simple
+preReleaseCommand: ruby .scripts/batch_release.rb
+releaseBranchPrefix: batch-release-sentry-ruby
+statusProvider:
+  name: github
+artifactProvider:
+  name: github
+targets:
+    - name: gem
+    - name: registry
+      type: sdk
+      config:
+          canonical: 'gem:sentry-ruby'
+    - name: registry
+      type: sdk
+      config:
+          canonical: 'gem:sentry-ruby-core'
+    - name: registry
+      type: sdk
+      config:
+          canonical: 'gem:sentry-rails'
+    - name: registry
+      type: sdk
+      config:
+          canonical: 'gem:sentry-sidekiq'
+    - name: registry
+      type: sdk
+      config:
+          canonical: 'gem:sentry-delayed_job'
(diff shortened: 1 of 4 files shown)


.craft.yml
```diff
@@ -0,0 +1,33 @@
...
+github:
+    owner: getsentry
+    repo: sentry-ruby
+changelogPolicy: simple
+preReleaseCommand: ruby .scripts/batch_release.rb
+releaseBranchPrefix: batch-release-sentry-ruby
+statusProvider:
+  name: github
+artifactProvider:
+  name: github
```
BYK, line 11: These are now the defaults when we set `0.22.2` as the min version. (well, anything >=0.21.0 actually https://github.com/getsentry/craft/releases/tag/0.21.0)

.craft.yml
```diff
@@ -0,0 +1,33 @@
...
+targets:
```
BYK, line 12: No `github` target?

.scripts/batch_release.rb
```diff
@@ -0,0 +1,44 @@
...
+
+def update_gemspec_dependency(gem_name, version)
+  file_name = "#{gem_name}/#{gem_name}.gemspec"
+  text = File.read(file_name)
+  new_contents = text.gsub(/spec.add_dependency "sentry-ruby-core", ".+"/, "spec.add_dependency \"sentry-ruby-core\", \"~> #{version}\"")
+  File.open(file_name, "w") {|file| file.puts new_contents }
+end
+
+# when craft runs scripts it inserts version to the 2nd argument
+version = ARGV[1]
```
BYK, line 30: It also sets environment variables if that's easier:

https://github.com/getsentry/craft/blob/5030c50f8c37e81f39ce51c4a419ec787cc8f774/src/commands/prepare.ts#L249-L250

We may wanna document this.

BYK:
Did not approve due to lack of `github` target. Once we clarify that, I'll approve as the rest looks good.

.scripts/batch_release.rb
st0012: @BYK thanks for the tip. I'll change this later with the other version bumping script 👍

.craft.yml
```diff
@@ -0,0 +1,29 @@
+minVersion: '0.22.2'
+github:
+    owner: getsentry
+    repo: sentry-ruby
```
BYK, line 4: You can remove these too, they are now inferred automatically.
```suggestion
```

BYK (approved):

PR #1451, opened 2021-05-23
Build gem source in batch prepare workflow

st0012:
Instead of writing another "batch build" workflow, I find it easier to just prepare + build all gems in the same workflow and then upload the artifacts right away.

### DIFF
--- a/.github/workflows/prepare_batch_release.yml
+++ b/.github/workflows/prepare_batch_release.yml
@@ -24,3 +24,8 @@ jobs:
         with:
           version: ${{ github.event.inputs.version }}
           force: ${{ github.event.inputs.force }}
+      - name: Archive Artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ github.sha }}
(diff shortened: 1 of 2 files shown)


st0012:
@BYK The script I wrote should update the version number + build the gem source. It worked when I used `craft prepare` locally but when I gave it a test run [here](https://github.com/getsentry/sentry-ruby/runs/2649581769?check_suite_focus=true) it didn't build the artifacts. Can you give the log a log and see what did I do wrong? Thanks.

BYK (changes requested):
>Instead of writing another "batch build" workflow, I find it easier to just prepare + build all gems in the same workflow and then upload the artifacts right away.

Prepare step should not build anything at all. It should just get the repo ready to be built and push that to the release branch. This is tho ensure we have a separate, isolated branch that is built with the same system with all CI and also easily addressable.

st0012:
@BYK I see, thanks! I will open another PR for the batch build workflow.

PR #1452, opened 2021-05-25
Rack-like lambda integration (#1)

david-christensen:
## TODOS:

**Other Notes**

## Description
Describe your changes:

[./sentry-ruby]

* Adds Sentry::Lambda::CaptureExceptions
* Sentry::Event .timestamp should include milliseconds

Creates a Rack-like integration which can be used in a lambda
function like so:

```
def lambda_handler(event:, context:)
  Sentry::Lambda.wrap_handler(event: event, context: context) do
    # my biz logic here....
  end
end
```

A warning message can be enabled with `capture_timeout_warning: true` like so:
```
def lambda_handler(event:, context:)
  Sentry::Lambda.wrap_handler(event: event, context: context, capture_timeout_warning: true) do
    # my biz logic here....
  end
end
```

---------

These changes provide the basic sentry support that our company needs to start using sentry within AWS Lambda functions, but we also wanted to create an opportunity for feedback and next steps.

This implementation was largely based off of the [Rack integration](https://github.com/paymentspring/sentry-ruby/blob/master/sentry-ruby/lib/sentry/rack/capture_exceptions.rb). From there I attempted to capture sentry info similar to how the [python integration](https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/integrations/aws_lambda.py) does.

The python example raises a [Timeout Warning](https://docs.sentry.io/platforms/python/guides/aws-lambda/#timeout-warning), whereas I decided to capture a sentry message and allow the function to continue.

At this time we do NOT handle request data from Lambda event hashes the way the other examples do. I noticed that `Sentry::RequestInterface` is primarily written for Sentry::Rack as it stands. I would imagine that `RequestInterface` would be abstracted and there would be a Sentry::RackRequestInterface, AwsAPIGWInterface, AwsALBInterface, etc. 

I thought this might be a good time to "hit pause" and see if the folks at sentry would like to help steer the implementation.

Let us know what you think!

Thanks!


st0012:
@david-christensen hey thanks for the PR 👍 I think it's a great idea to provide lambda support, but given that it's not what most Ruby developers would need, can you build an integration gem instead? 

to build one, you can follow [this guideline](https://github.com/getsentry/sentry-ruby/blob/master/EXTENSION.md) and use `sentry-rails` or `sentry-sidekiq` as examples. also, you can choose to publish the gem yourself and maintain it at your pace.

please feel free to open issues for the APIs you need or any questions for building the gem 🙂

st0012:
@david-christensen wow that's quick. great job 👍 

some advice:
- it doesn't need to be a fork of this repo. and other gem folders should be useless to you so you can remove them.
- you can remove the `.craft.yml` file. it's only needed for the release flow we have at Sentry.
- this may not be the case for lambda applications, but recording important incidents during the request as breadcrumbs could be helpful for debugging. it can be done with

```ruby
crumb = Sentry::Breadcrumb.new(
  data: data,
  category: category,
  timestamp: started.to_i
)
Sentry.add_breadcrumb(crumb)
```

- I'd also recommend users to use 2 breadcrumb loggers: `:sentry_logger` and `:http_logger` with this gem in the readme:

```ruby
```

david-christensen:
@st0012 
Nice! 
Yeah - I noticed that was kinda messy... At least it's working for now!
Thanks for the advice - that will help me clean things up when I have time!

Issue #1378, opened 2021-04-01
How to Prevent Sending Exceptions for Rails console or db:migrate


westonganger:
I do not want the exceptions from my Rails console sessions or db:migrate to send to sentry. Is there some way to disable this.

helmiItsavirus:
@westonganger did you activate sentry on the development environment?

westonganger:
No it is disabled in development. This is not a development environment issue.

westonganger:
I am also recieving errors on server when running tasks that dont exist. IE. Typos:

```ruby
bundle exec rails consolee
```

RuntimeError: Don't know how to build task 'consolee'

Again I dont want to send error notifications while I am performing server tasks

st0012:
the SDK integrates with rake and intercepts exceptions raised during a rake task. I can add an option to disable the integration all together, but it won't be practical to introduce something like "don't capture this exception only during rake tasks". do you think that'll solve your problem?

westonganger:
Yes please add option to disable the rake integration. Would this also remove the notifications during `rails console` sessions?

st0012:
I don't think the SDK captures errors from the console. At least not in an intentional way and I can't reproduce this behavior with my apps. Can you check the event's stacktrace and point out what captures the error?

westonganger:
Hmm it appears you are correct about the `rails:console` I guess I was incorrect about that one.

st0012:
@westonganger it's been shipped in `v4.5.0` 😄

MaicolBen:
What about `capture_message` ? How can we skip it? In the past it was possible to do `Raven.configuration.environments = %w[ignore_on_console] `

PR #1453, opened 2021-05-25
Add configuration option skip_rake_integration

westonganger:
Add configuration option `skip_rake_integration`. As mentioned in https://github.com/getsentry/sentry-ruby/issues/1378#issuecomment-815415357

### DIFF
--- a/sentry-ruby/spec/sentry/configuration_spec.rb
+++ b/sentry-ruby/spec/sentry/configuration_spec.rb
@@ -429,4 +429,12 @@ RSpec.describe Sentry::Configuration do
       expect(instance.var2). to eq 2
     end
   end
+
+  describe "#skip_rake_integration" do
+    it "returns true" do
+      subject.skip_rake_integration = true
+      expect(subject.skip_rake_integration).to eq(true)
+    end
+  end
+
 end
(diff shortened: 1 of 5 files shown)


sentry-ruby/spec/sentry/configuration_spec.rb
```diff
@@ -429,4 +429,12 @@ class SentryConfigurationSample < Sentry::Configuration
       expect(instance.var2). to eq 2
     end
   end
+
+  describe "#skip_rake_integration" do
+    it "returns true" do
```
st0012, line 434: can you add another one to confirm that its default is `false`

sentry-ruby/spec/support/Rakefile.rb
```diff
@@ -9,3 +9,10 @@
 task :raise_exception do
   1/0
 end
+
+task :raise_exception_without_rake_integration do
+  Sentry.init do |config|
```
st0012, line 14: since the SDK is already initialized, this can be `Sentry.configuration.skip_rake_integration = true`

st0012 (changes requested):
sorry for the delay and thanks for the PR 🙏 
just 2 minor changes then I think it'll be good to merge.
also, can you add a changelog entry for this?

st0012:
@westonganger actually, I'll cut a release today so I'll change them manually and make it in 4.5.0 as well 🙂

PR #1455, opened 2021-05-27
Add "Batch Build Release" workflow

st0012:


.github/workflows/batch_build_release.yml
```diff
@@ -0,0 +1,22 @@
...
+  push:
+    branches:
+      - release/**
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v1
```
BYK, line 13: Why do we need this step, can you add a comment?

BYK (approved):
LGTM

.github/workflows/batch_build_release.yml
st0012: sorry that's redundant. I've removed it 😅

PR #1461, opened 2021-06-02
ref: Clean up the release clutter

BYK:
Follow up to #1450.


.vscode/numbered-bookmarks.json
```diff
@@ -0,0 +1,3 @@
+{
```
st0012, line 1: @BYK can you remove this? thx

.github/workflows/prepare_raven_release.yml
```diff
@@ -1,10 +1,7 @@
-name: Prepare Release
```
st0012, line 1: I'd prefer keeping it for now for releasing new integrations in the future.

BYK: We can add this back easily. I'd prefer if we nudge people to use the batch release. What do you think?

.vscode/numbered-bookmarks.json
BYK: Yup, sorry 😅

.github/workflows/prepare_raven_release.yml
st0012: ok let's drop it for now 👍

st0012 (approved):

st0012:
@BYK Thanks!

PR #1462, opened 2021-06-02
ref: Use the new batch config for Craft registry target

BYK:


BYK:
This is a draft since Craft `0.23.0` is not released yet.

st0012 (approved):

Issue #1249, opened 2021-01-29
sentry + eager_load breaks ActiveJob rescue_from


CroneKorkN:
**Describe the bug**

Sentry eats errors in Jobs. They return nil instead.

**To Reproduce**

- make a new rails app
- add sentry-rails to Gemfile: https://github.com/CroneKorkN/sentry_broken_with_eager_load/blob/master/Gemfile#L6
- initialize sentry: https://github.com/CroneKorkN/sentry_broken_with_eager_load/blob/master/config/initializers/sentry.rb
- add an rescue_from to you application_job.rb: https://github.com/CroneKorkN/sentry_broken_with_eager_load/blob/master/app/jobs/application_job.rb
- create a job that fails with an around_perform: https://github.com/CroneKorkN/sentry_broken_with_eager_load/blob/master/app/jobs/error_job.rb
- enable eagerload in development config: https://github.com/CroneKorkN/sentry_broken_with_eager_load/blob/0734f3648f9687c27f3f8ea505b1b26f5975cd97/config/environments/development.rb#L12
- `rails c`
- `ErrorJob.perform_later`

**Expected behavior**

This line should never be executed because the line before it throws an error: https://github.com/CroneKorkN/sentry_broken_with_eager_load/blob/master/app/jobs/error_job.rb#L4

You should not see `>>> I SHOULD NEVER BE EXECUTED! (NilClass)`

**Actual behavior**

The line is executed because `block.call` returns `nil`.

You see `>>> I SHOULD NEVER BE EXECUTED! (NilClass)`

**Environment**
 - Ruby Version: 2.7.1
 - Sentry-Rails: 4.1.6
 - Rails: 6.1.0 

**Raven Config**

https://github.com/CroneKorkN/sentry_broken_with_eager_load/blob/master/config/initializers/sentry.rb

**Test Repo**

On this commit, the error occurs: https://github.com/CroneKorkN/sentry_broken_with_eager_load/tree/0734f3648f9687c27f3f8ea505b1b26f5975cd97
This commit fixes it by disabling eager_load: https://github.com/CroneKorkN/sentry_broken_with_eager_load/tree/ddf2e41ebee1b31afb5c533601048ae740bb00f6

Removing the sentry-rails gem obviously also fixes it.

CroneKorkN:
This return should not happen, from my limited understanding: https://github.com/getsentry/sentry-ruby/blob/master/sentry-rails/lib/sentry/rails/active_job.rb#L30

[...]

PR #1464, opened 2021-06-03
Correct the timing of loading ActiveJobExtensions

st0012:
Fixes https://github.com/getsentry/sentry-ruby/issues/1249


sentry-rails/lib/sentry/rails/railtie.rb
```diff
@@ -12,13 +12,19 @@ class Railtie < ::Rails::Railtie
       app.config.middleware.use(Sentry::Rails::RescuedExceptionInterceptor)
     end
 
+    # because the extension works by registering the around_perform callcack, it should always be ran
```
kml, line 15: "callcack" => "callback"

Issue #1468, opened 2021-06-10
Automatically set user via sidekiq middleware


shouichi:
**Describe the idea**

It would be nice if a sidekiq middleware transparently set `user` to the current scope so that we can track the number of affected users even for background jobs.

**Why do you think it's beneficial to most of the users**

**Possible implementation**

st0012:
I know it'd be a convenient feature to many users, but I don't think it's something the SDK should do. 

Because it requires the SDK to "guess" what models & attributes the application uses for users and to provide to Sentry, this could be error-prone and hard to debug. 

Also, to fit different applications' use cases, the SDK will need to expose several new config options for it. For example, in my company's application the primary user model is called `Customer`, so the SDK must provide a way to configure the model name like `config.user_model_name`. Similar things are also required for attributes and scoping...etc.

Given the error-prone nature of such features + the maintenance effort, I don't plan to implement this feature yet. 

However, I think it'd be a great feature for a SDK extension though 🙂

shouichi:
What I meant by `user` was `Sentry.get_current_scope.user` which is just a hash and application-agnostic I believe.

st0012:
so how should the SDK get the data to set the user?

shouichi:
In the rails app I work for we set user id to sentry in `before_action`.

```
class ApplicationController
	before_action :set_sentry_user

	def set_sentry_user
		Sentry.set_user(id: current_user.id)
	end
end
```

SDK can pick up that.

```
class ClientMiddleware
	def call(worker_class, job, queue, redis_pool)
		scope = Snetry.get_current_scope
		job["sentry_user_id"] = scope.user[:id]
		yield
	end
end

class ServerMiddleware
	def call(worker, job, queue)
		Sentry.set_user(id: job["sentry_user_id"])
		yield
	end
end
```

st0012:
I see. I think this approach may work but I'll need to do some testing for it first 🙂

shouichi:
I'm currently implementing this feature and would like to send a PR 😃

st0012:
I'm looking forward to see your PR then 👍 thanks

PR #1469, opened 2021-06-11
Set user to the current scope via sidekiq middleware

shouichi:
Close https://github.com/getsentry/sentry-ruby/issues/1468.

### DIFF
--- a/sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
+++ b/sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
@@ -2,7 +2,7 @@ require 'sentry/sidekiq/context_filter'
 
 module Sentry
   module Sidekiq
-    class SentryContextMiddleware
+    class SentryContextServerMiddleware
       def call(_worker, job, queue)
         return yield unless Sentry.initialized?
 
@@ -10,6 +10,9 @@ module Sentry
 
         Sentry.clone_hub_to_current_thread
         scope = Sentry.get_current_scope
+        if (user = job["sentry_user"])
+          scope.set_user(user)
+        end
         scope.set_tags(queue: queue, jid: job["jid"])
         scope.set_contexts(sidekiq: job.merge("queue" => queue))
         scope.set_transaction_name(context_filter.transaction_name)
@@ -34,6 +37,16 @@ module Sentry
 
         transaction.set_http_status(status)
         transaction.finish
+      end
+    end
+
+    class SentryContextClientMiddleware
+      def call(_worker_class, job, _queue, _redis_pool)
+        return yield unless Sentry.initialized?
+
+        user = Sentry.get_current_scope.user
+        job["sentry_user"] = user unless user.empty?
+        yield
       end
     end
   end
--- a/sentry-sidekiq/spec/sentry/sidekiq/sentry_context_middleware_spec.rb
+++ b/sentry-sidekiq/spec/sentry/sidekiq/sentry_context_middleware_spec.rb
@@ -1,0 +1,109 @@
+require "spec_helper"
+
+RSpec.shared_context "sidekiq", shared_context: :metadata do
+  let(:user) { { "id" => rand(10_000) } }
+
+  let(:client) do
+    Sidekiq::Client.new.tap do |client|
+      client.middleware do |chain|
+        chain.add Sentry::Sidekiq::SentryContextClientMiddleware
+      end
+    end
+  end
+
+  let(:random_empty_queue) do
+    Sidekiq::Queue.new(rand(10_000)).tap do |queue|
+      queue.clear
+    end
+  end
+end
+
+RSpec.describe Sentry::Sidekiq::SentryContextServerMiddleware do
+  include_context "sidekiq"
+
+  after { Sidekiq::RetrySet.new.clear }
+
+  it "sets user from the job to events" do
+    perform_basic_setup { |config| config.traces_sample_rate = 1.0 }
+    Sentry.set_user(user)
+
+    queue = random_empty_queue
+    options = { queues: [queue.name] }
+    processor = Sidekiq::Manager.new(options).workers.first
+
+    client.push('queue' => queue.name, 'class' => HappyWorker, 'args' => [])
+
+    expect { processor.send(:process_one) }.
+      to change { Sentry.get_current_client.transport.events.size }.by(1)
+
+    event =  Sentry.get_current_client.transport.events.first
+    expect(event).not_to be_nil
+    expect(event.user).to eq(user)
+  end
+
+  it "sets user from the job to sidekiq event if worker raises an exception" do
+    perform_basic_setup { |config| config.traces_sample_rate = 0 }
+    Sentry.set_user(user)
+
+    queue = random_empty_queue
+    options = { queues: [queue.name] }
+    processor = Sidekiq::Manager.new(options).workers.first
+
+    client.push('queue' => queue.name, 'class' => SadWorker, 'args' => [])
+
+    expect do
+      begin; processor.send(:process_one); rescue RuntimeError; end
+    end.
+      to change { Sentry.get_current_client.transport.events.size }.by(1)
+
+    Sentry.get_current_client.transport.events.each do |event|
+      expect(event.user).to eq(user)
+    end
+  end
+
+  it "sets user from the job to sidekiq and error events if worker raises an exception when trace enabled" do
+    perform_basic_setup { |config| config.traces_sample_rate = 1.0 }
+    Sentry.set_user(user)
+
+    queue = random_empty_queue
+    options = { queues: [queue.name] }
+    processor = Sidekiq::Manager.new(options).workers.first
+
+    client.push('queue' => queue.name, 'class' => SadWorker, 'args' => [])
+
+    # XXX: In ruby 2.4, two events are pushed. In other versions, only one
+    # event is pushed. Use by_at_least.
+    expect do
+      begin; processor.send(:process_one); rescue RuntimeError; end
+    end.
+      to change { Sentry.get_current_client.transport.events.size }.by_at_least(1)
+
+    Sentry.get_current_client.transport.events.each do |event|
+      expect(event.user).to eq(user)
+    end
+  end
+end
+
+RSpec.describe Sentry::Sidekiq::SentryContextClientMiddleware do
+  include_context "sidekiq"
+
+  before { perform_basic_setup }
+
+  it "does not user to the job if user is absence in the current scope" do
+    queue = random_empty_queue
+    client.push('queue' => queue.name, 'class' => HappyWorker, 'args' => [])
+
+    expect(queue.size).to be(1)
+    expect(queue.first["sentry_user"]).to be_nil
+  end
+
+  it "sets user of the current scope to the job if present" do
+    queue = random_empty_queue
+    Sentry.set_user(user)
+
+    client.push('queue' => queue.name, 'class' => HappyWorker, 'args' => [])
+
+    expect(queue.size).to be(1)
+    expect(queue.first["sentry_user"]).to eq(user)
+  end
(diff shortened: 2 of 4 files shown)


sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
```diff
@@ -36,5 +37,13 @@ def finish_transaction(transaction, status)
         transaction.finish
       end
     end
+
+    class SentryContextClientMiddleware
+      def call(_worker_class, job, _queue, _redis_pool)
```
shouichi, line 42: Should we call the following methods? I'm not sure.

- `return yield unless Sentry.initialized?`
- Sentry.clone_hub_to_current_thread

sentry-sidekiq/spec/sentry/sidekiq/sentry_context_middleware_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  before { perform_basic_setup }
+
+  it "sets user to the current scope from the job" do
+    user = { id: "u1" }
+    job = instance_double("job")
+    allow(job).to receive(:merge)
+    allow(job).to receive(:[])
+    allow(job).to receive(:[]).with("sentry_user").and_return(user)
+
+    expect(Sentry.get_current_scope).to receive(:set_user).with(user).exactly(1).time
```
shouichi, line 13: This spec does not work.

shouichi (commented):
@st0012 Could you help to finish this PR?

sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
```diff
@@ -2,14 +2,15 @@
...
-    class SentryContextMiddleware
+    class SentryContextServerMiddleware
       def call(_worker, job, queue)
         return yield unless Sentry.initialized?
 
         context_filter = Sentry::Sidekiq::ContextFilter.new(job)
 
         Sentry.clone_hub_to_current_thread
         scope = Sentry.get_current_scope
+        scope.set_user(scope.user)
```
st0012, line 13: shouldn't the server middleware get the user from job hash instead of its scope (which would just be `nil`)?

sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
st0012: `return yield unless Sentry.initialized?` is needed but `Sentry.clone_hub_to_current_thread` is not.

sentry-sidekiq/spec/sentry/sidekiq/sentry_context_middleware_spec.rb
st0012: can you test this feature end-to-end instead?

sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
shouichi: I'll fix, thanks :+1:

sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
shouichi: That's so true, thanks for catching it!

sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
```diff
@@ -2,20 +2,29 @@
 
 module Sentry
   module Sidekiq
-    class SentryContextMiddleware
+    class SentryContextServerMiddleware
+      def initialize(testing_only_callback: nil)
+        @testing_only_callback = testing_only_callback
```
shouichi, line 7: The only way I can imagine was to pass a callback. Is there a better way?

sentry-sidekiq/spec/spec_helper.rb
```diff
@@ -1,6 +1,10 @@
 require "bundler/setup"
 require "pry"
 
+# This allows us to call sidekiq private methods for testing.
+# Make sure to set this before requiring sidekiq.
+$TESTING = true
```
shouichi, line 6: This global variable is used in the sidekiq codebase and allows us to call private methods. Using this can potentially improve the current code that uses `send` to call private methods.

shouichi (commented):
@st0012 Hi, I rewrote the test in an E2E way. Could you take a look?

sentry-sidekiq/spec/sentry/sidekiq/sentry_context_middleware_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+
+    queue = random_empty_queue
+    options = { fetch: Sidekiq::BasicFetch.new(queues: [queue.name]) }
+    processor = Sidekiq::Processor.new(nil, options)
+
+    client.push('queue' => queue.name, 'class' => TestWorker, 'args' => [])
+
+    expect(queue.size).to be(1)
+    processor.process_one
+    expect(queue.size).to be(0)
```
st0012, line 57: it's a bit tricky but you can test user setting with transaction events:

1. Enable tracing

```ruby
perform_basic_setup do |config|
end
```

2. process the job (`processor.process_one`)
3. use `Sentry.get_current_client.transport.events.first` to get the event.
4. check if the event contains the correct user data.

besides using tracing events, I think making `TestWorker` raise exception should also generate an event that contains the user info. personally I'd test both scenarios to make sure the user data is properly set under any condition.

sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
st0012: I've left a comment for a better approach below. I think this hack is unnecessary.

sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
```diff
@@ -36,5 +45,15 @@ def finish_transaction(transaction, status)
         transaction.finish
       end
     end
+
+    class SentryContextClientMiddleware
+      def call(_worker_class, job, _queue, _redis_pool)
+        return yield unless Sentry.initialized?
+
+        scope = Sentry.get_current_scope
+        job["sentry_user"] = scope.user
```
st0012, line 54: can this check if `scope.user` is presented? adding a `nil` or an empty hash is not necessary.

sentry-sidekiq/spec/spec_helper.rb
st0012: I think calling `Processor#process` with `send` in a few occasions is acceptable. so I prefer not to add this now.

thanks for sharing this information though, I never knew we can do that 🙂

sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
shouichi: Sure.

sentry-sidekiq/spec/spec_helper.rb
shouichi: Do you think it's better to delete this to be consistent?

st0012: yeah that'd be better, thanks 👍

sentry-sidekiq/spec/sentry/sidekiq/sentry_context_middleware_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    queue = random_empty_queue
+    options = { fetch: Sidekiq::BasicFetch.new(queues: [queue.name]) }
+    processor = Sidekiq::Processor.new(nil, options)
+
+    client.push('queue' => queue.name, 'class' => SadWorker, 'args' => [])
+
+    expect do
+      begin; processor.send(:process_one); rescue RuntimeError; end
+    end.
+      to change { Sentry.get_current_client.transport.events.size }.by(1)
```
shouichi, line 59: This spec fails because two events are pushed. Honestly, I don't understand why. @st0012 Could you please help?

shouichi (commented):
I believe it's almost done but there's one outstanding issue. The notable changes made from the last review are

- Use `traces_sample_rate = 1.0` "hack".
- Use HappyWroker/SadWoker from spec helper.

sentry-sidekiq/spec/sentry/sidekiq/sentry_context_middleware_spec.rb
st0012: if you enable tracing, it'll create a tracing event. and because you also use the `SadWorker`, which  raises an exception, it'll create another error event.

I think we should have 3 scenarios:

- Use `traces_sample_rate = 1` 
  - with `HappyWorker` to test the no exception scenario.
  - with `SadWorker` to test the exception with tracing scenario.
- Use `SadWorker` to test the exception scenario.

shouichi: Added those scenarios.

sentry-sidekiq/spec/sentry/sidekiq/sentry_context_middleware_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+
+  after { Sidekiq::RetrySet.new.clear }
+
+  it "sets user from the job to events" do
+    perform_basic_setup { |config| config.traces_sample_rate = 1.0 }
+    Sentry.set_user(user)
+
+    queue = random_empty_queue
+    options = { fetch: Sidekiq::BasicFetch.new(queues: [queue.name]) }
+    processor = Sidekiq::Processor.new(nil, options)
```
st0012, line 32: I think the way you initialize sidekiq processor is not compatible with sidekiq v5.2 (which runs in Ruby 2.4 test suite).

instead of initializing the processor directly, can we initialize a manager and get the processor from it? like

https://github.com/getsentry/sentry-ruby/blob/6d2fc60fe21ea3def9da891a470d1a1c6a83760c/sentry-sidekiq/spec/sentry/sidekiq_spec.rb#L15-L19

I will merge this PR once the test is fixed 🙂

shouichi: Fixed!

st0012: actually it still fails with Sidekiq 5

shouichi: Run tests with sidekiq 5 and confirmed it works.

sentry-sidekiq/spec/sentry/sidekiq/sentry_context_middleware_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+    perform_basic_setup { |config| config.traces_sample_rate = 1.0 }
+    Sentry.set_user(user)
+
+    queue = random_empty_queue
+    options = { queues: [queue.name] }
+    processor = Sidekiq::Manager.new(options).workers.first
+
+    client.push('queue' => queue.name, 'class' => SadWorker, 'args' => [])
+
+    # XXX: In ruby 2.4, two events are pushed. In other versions, only one
```
shouichi, line 74: @st0012 this behavior is strange...

st0012: I'll check this in a follow up PR.

st0012 (approved):
thanks for the contribution 👍 
I'll add another PR to refactor the test cases and add changelog entry for this.

shouichi:
Thanks a lot for the help :+1:

PR #1482, opened 2021-06-24
Add tracing support to sentry-delayed_job

### LINKED ISSUE #1312: Support performance monitoring in sentry-delayed_job

fsateler:
Fixes #1312

### DIFF
--- a/sentry-delayed_job/lib/sentry/delayed_job/plugin.rb
+++ b/sentry-delayed_job/lib/sentry/delayed_job/plugin.rb
@@ -13,14 +13,21 @@ module Sentry
           next block.call(job, *args) unless Sentry.initialized?
 
           Sentry.with_scope do |scope|
-            scope.set_contexts(**generate_contexts(job))
+            contexts = generate_contexts(job)
+            scope.set_transaction_name(contexts.dig(ACTIVE_JOB_CONTEXT_KEY, :job_class) || contexts.dig(DELAYED_JOB_CONTEXT_KEY, :job_class))
+            scope.set_contexts(**contexts)
             scope.set_tags("delayed_job.queue" => job.queue, "delayed_job.id" => job.id.to_s)
+
+            transaction = Sentry.start_transaction(name: scope.transaction_name, op: "delayed_job")
+            scope.set_span(transaction) if transaction
 
             begin
               block.call(job, *args)
+
+              finish_transaction(transaction, 200)
             rescue Exception => e
               capture_exception(e, job)
-
+              finish_transaction(transaction, 500)
               raise
             end
           end
@@ -40,7 +47,8 @@ module Sentry
           queue: job.queue,
           created_at: job.created_at,
           last_error: job.last_error&.byteslice(0..1000),
-          handler: job.handler&.byteslice(0..1000)
+          handler: job.handler&.byteslice(0..1000),
+          job_class: compute_job_class(job.payload_object),
         }
 
         if job.payload_object.respond_to?(:job_data)
@@ -54,6 +62,15 @@ module Sentry
         context
       end
 
+      def self.compute_job_class(payload_object)
+        if payload_object.is_a? Delayed::PerformableMethod
+          klass = payload_object.object.is_a?(Class) ? payload_object.object.name : payload_object.object.class.name
+          "#{klass}##{payload_object.method_name}"
+        else
+          payload_object.class.name
+        end
+      end
+
       def self.capture_exception(exception, job)
         Sentry::DelayedJob.capture_exception(exception, hint: { background: false }) if report?(job)
       end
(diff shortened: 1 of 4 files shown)


sentry-delayed_job/spec/sentry/delayed_job_spec.rb
```diff
@@ -234,6 +237,93 @@ def perform
...
+      it "records transaction with exception" do
+        FailedJob.perform_later
+        enqueued_job = Delayed::Backend::ActiveRecord::Job.last
+        begin
+          enqueued_job.invoke_job
+        rescue ZeroDivisionError
+          nil
+        end
+  
+        expect(transport.events.count).to eq(4)
```
fsateler, line 272: I have no idea why the event count is 2 larger than without activejob. Is this a problem? If so, hints as to why are appreciated.

sentry-delayed_job/spec/sentry/delayed_job_spec.rb
```diff
@@ -234,6 +237,93 @@ def perform
         expect(event[:contexts][:"Active-Job"][:job_class]).to eq("FailedJob")
       end
     end
+
+    context "when tracing is enabled" do
+      before do
+        perform_basic_setup do |config|
+          config.traces_sample_rate = 1.0
```
st0012, line 244: you also need this line here:

```ruby
          config.rails.skippable_job_adapters << "ActiveJob::QueueAdapters::DelayedJobAdapter"
```

the 2 extra events are from the `ActiveJob` integration. 

in the earlier tests it's disabled with

https://github.com/getsentry/sentry-ruby/blob/53f0bd9a0beb5c742b7979135f70f5ff98d82021/sentry-delayed_job/spec/sentry/delayed_job_spec.rb#L182-L184

but because this block is evaluated after that, you need to register the adapter again.

fsateler: Aha! thanks, fixed.

sentry-delayed_job/lib/sentry/delayed_job/plugin.rb
```diff
@@ -40,8 +49,15 @@ def self.generate_contexts(job)
           queue: job.queue,
           created_at: job.created_at,
           last_error: job.last_error&.byteslice(0..1000),
-          handler: job.handler&.byteslice(0..1000)
+          handler: job.handler&.byteslice(0..1000),
+          handler_class: job.payload_object.class,
         }
+        if job.payload_object.is_a? Delayed::PerformableMethod
```
st0012, line 55: I prefer having these computation in a `compute_job_class(payload_object)` method and at here we just do:
```ruby
  handler_class: job.payload_object.class,
  job_class: compute_job_class(job.payload_object)
}
```

also, can you write test cases for these changes too? thx

sentry-delayed_job/lib/sentry/delayed_job/plugin.rb
```diff
@@ -40,8 +49,15 @@ def self.generate_contexts(job)
           queue: job.queue,
           created_at: job.created_at,
           last_error: job.last_error&.byteslice(0..1000),
-          handler: job.handler&.byteslice(0..1000)
+          handler: job.handler&.byteslice(0..1000),
+          handler_class: job.payload_object.class,
```
st0012, line 53: is it necessary to have both `handler_class` and `job_class`? I think the latter should include the former?

sentry-delayed_job/lib/sentry/delayed_job/plugin.rb
```diff
@@ -12,15 +12,24 @@ class Plugin < ::Delayed::Plugin
...
+
             begin
               block.call(job, *args)
+
+              finish_transaction(transaction, 200)
             rescue Exception => e
               capture_exception(e, job)
-
+              finish_transaction(transaction, 500)
+              scope.clear
```
st0012, line 32: this is unnecessary. we have this in `sentry-sidekiq` because the middleware can't be wrapped inside a local scope (`Sentry.with_scope`).

sentry-delayed_job/lib/sentry/delayed_job/plugin.rb
```diff
@@ -12,15 +12,24 @@ class Plugin < ::Delayed::Plugin
         lifecycle.around(:invoke_job) do |job, *args, &block|
           next block.call(job, *args) unless Sentry.initialized?
 
+
```
st0012, line 15: nitpick: can you remove this newline? thx

st0012 (commented):
hey thanks a lot for adding this 👍 
just left some comments for polishing. I think you also need to rebase the branch and resolve the conflict in order to start the build.

sentry-delayed_job/lib/sentry/delayed_job/plugin.rb
fsateler: No, it is not. handler_class was the first version and I forgot to delete.

st0012 (approved):
Thanks for implementing this 😄

PR #1493, opened 2021-06-30
Fix broken URL.

shmokmt:
## Description
Describe your changes:

https://github.com/getsentry/raven-ruby/blob/master/lib/raven/configuration.rb is a broken link.
so I replaced the new one with it.


st0012 (approved):
thanks 👍

PR #1494, opened 2021-07-02
Don't force ActiveJob to load during initialization

jdelStrother:
Since the 4.5.1 release, setting `Rails.application.config.active_job.foo = bar` in an initializers file (eg config/initializers/foo.rb) would be ignored.

As soon as ActiveJob::Base is referenced, it applies Rails.application.config.active_job, and any further changes to that config are ignored. By loading ActiveJob::Base before eager_load, it prevents ActiveJob being configured in, eg, config/initializers/new_framework_defaults.rb

How about using `ActiveSupport.on_load(:active_job)` instead?

This tweaks the behaviour in https://github.com/getsentry/sentry-ruby/pull/1464 which was introduced to fix https://github.com/getsentry/sentry-ruby/issues/1249, would be good to confirm that the original issue poster isn't affected by this new fix.

I tried to add a test but found it kind of tricky. ActiveJob::Base is already loaded by the time we call `make_basic_app` in specs, so it seems impossible to hook in early enough to test the failure case.

cc @CroneKorkN, @st0012

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+- Correct the timing of loading ActiveJobExtensions, take 2 [#1494](https://github.com/getsentry/sentry-ruby/pull/1494)
+
 ## 4.6.0
 
 ### Features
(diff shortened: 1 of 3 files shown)


st0012:
@jdelStrother thanks for spotting the issue and opening this PR. I can confirm that it fixes the original issue by manually testing it 👍 
but given that there are 2 scenarios to cover, I want to come up with a test setup for testing patches in isolation. give me a few days and I'll come back to this.

st0012:
@jdelStrother I've added an isolated test for ActiveJob patching in https://github.com/getsentry/sentry-ruby/pull/1497. can you modify it and make it fail without your fix?

jdelStrother:
Sure, I've pushed an update with a test.

st0012 (approved):
@jdelStrother thanks for the fix 😄

st0012:
@jdelStrother btw I also updated the PR description and changelog entry to make the change more clear to users.

PR #1498, opened 2021-07-07
Fix `Sentry::Utils::RealIP` not filtering trusted proxies when part of IP subnet passed as `IPAddr` to `trusted_proxies`.

vojtad:
Filtering trusted proxies based on list of IP subnets passed as instances of `IPAddr` got broken in #1288.

When calling `IPAddr#to_s` resulting string doesn't contain the IP mask. Therefore when creating `IPAddr` again from this string IP mask is lost and filtering trusted proxies is broken.

I changed `Sentry::Utils::RealIP` to only call `.to_s` when trusted proxy entry is not an instance of `IPAddr`. This keeps complete information about defined IP subnet as trusted proxy and makes filtering work as expected.

I've also added a test for this case.

This PR also fixes `Sentry::Utils::RealIP` with proxies loaded from Rails' `config.action_dispatch.trusted_proxies` because they are stored as instances of `IPAddr`.

### DIFF
--- a/sentry-ruby/lib/sentry/utils/real_ip.rb
+++ b/sentry-ruby/lib/sentry/utils/real_ip.rb
@@ -29,7 +29,13 @@ module Sentry
         @client_ip = client_ip
         @real_ip = real_ip
         @forwarded_for = forwarded_for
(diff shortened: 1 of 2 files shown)


st0012 (approved):
thank you for the fix 👍

Issue #1499, opened 2021-07-07
Content Security Policy


tmaier:
It would be great, if you could show how to (dynamically) set the content security policy reporting URI.

For example in the demo app at https://github.com/getsentry/sentry-ruby/blob/94c28fe9a7fd1c15a10eeeb0e1c4fcac099d4166/sentry-rails/examples/rails-6.0/config/initializers/content_security_policy.rb

st0012:
@tmaier do you mean parsing and generating the `Security Header Endpoint` used for CSP from the given DSN?

tmaier:
Yes.

I believe I got it working with the [secure_headers](https://github.com/github/secure_headers) gem

My `secure_headers.rb`:

```ruby
# Require sentry initializer so that Sentry is ready to be called within CSP
require_relative 'sentry'

SecureHeaders::Configuration.default do |config|
  # ...snip...
  if Sentry.configuration.dsn
    config.csp_report_only[:report_uri] = [
      "https://#{Sentry.configuration.dsn.host}/api/#{Sentry.configuration.dsn.project_id}/security/?sentry_key=#{Sentry.configuration.dsn.public_key}&sentry_environment=#{Sentry.configuration.environment}&sentry_release=#{Sentry.configuration.release}"
    ]
  end
end
```

tmaier:
Having this officially documented would be cool, having here a public API, which generates this endpoint would be even cooler

st0012:
@tmaier given that the SDK does hold all the information for the url, I think it's a great idea to have a helper for generating that for users. thanks for the suggestion 👍 I'll implement that soon.

st0012:
close via #1507

PR #1507, opened 2021-07-17
Generate Security Header Endpoint with Sentry.csp_report_uri from DSN

st0012:
Sentry requires a separated `Security Header Endpoint` for [Security Policy Reporting](https://docs.sentry.io/product/security-policy-reporting/). But the endpoint's major components like project id, public key or host are all identical with the SDK's DSN. Also, the additional environment or release information are also obtained by the SDK.

So this PR adds a helper `Sentry.csp_report_uri` to generate the `Security Header Endpoint` for users who report CSP events to the same project as DSN's (as suggested in #1499). It'll reduce the number of values they need to manage.

Take Rails' csp configuration for example:

**Without helper**

```ruby
Rails.application.config.content_security_policy do |policy|
  policy.report_uri ENV["SENTRY_CSP_URI"]
end
```

**With the helper**

```ruby
Rails.application.config.content_security_policy do |policy|
  policy.report_uri Sentry.csp_report_uri
end
```

### DIFF
--- a/sentry-ruby/lib/sentry/configuration.rb
+++ b/sentry-ruby/lib/sentry/configuration.rb
@@ -324,6 +324,15 @@ module Sentry
       log_error("Error detecting release", e, debug: debug)
     end
 
+    def csp_report_uri
+      if dsn && dsn.valid?
+        uri = dsn.csp_report_uri
+        uri += "&sentry_release=#{CGI.escape(release)}" if release && !release.empty?
+        uri += "&sentry_environment=#{CGI.escape(environment)}" if environment && !environment.empty?
+        uri
+      end
+    end
+
     private
 
     def excluded_exception?(incoming_exception)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 ## Unreleased
 
+### Features
+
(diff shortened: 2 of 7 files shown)


sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -324,6 +324,15 @@ def detect_release
       log_error("Error detecting release", e, debug: debug)
     end
 
+    def csp_report_uri
+      if dsn && dsn.valid?
+        uri = dsn.csp_report_uri
+        uri += "&sentry_release=#{release}" if release && !release.empty?
+        uri += "&sentry_environment=#{environment}" if environment && !environment.empty?
```
rhcarvalho, line 331: I'd expect we URL-encode `release` and `environment` to avoid creating malformed URI (and some form of URL-injection).

rhcarvalho (commented):
Generally I find it okay to add this. Would consider adding some escaping code instead of plain string concatenation.

sentry-ruby/lib/sentry/configuration.rb
st0012: Updated, thanks 👍

PR #1509, opened 2021-07-22
Disable background worker when executing rake tasks

### LINKED ISSUE #1508: `Sentry.capture_message` can't send a message from Rake
**Describe the bug**

`Sentry.capture_message` can't send a message from Rake.

This is a similar to #1324. I think message doesn't send because of process exit before send a message. A exception can be sent because `sentry-ruby` override `display_error_message` and specify `background: false` inside it.   
https://github.com/getsentry/sentry-ruby/blob/bae8c5be5a8d34ed9cade27c7d18c13bac86c521/sentry-ruby/lib/sentry/rake.rb#L7-L12

So I think this only happens in case of Sentry.capture_message` used.

**To Reproduce**

Add a following rake task and run it.

```rake 
namespace :dummy do
  desc 'Send a message Sentry'
  task test_sentry: :environment do
    Sentry.capture_message("Raise an exception for Sentry from Rake")
  end
end
```

```bash
$ bin/rails dummy:test_sentry
```

**Expected behavior**

Message send our sentry project.

**Actual behavior**

No message is being logged.

**Environment**
 - Ruby Version: 2.7.2
 - SDK Version: 4.6.1
 - Rails 6.0.3.7

st0012:
This fixes #1508 

(I know disabling background worker by setting `background_worker_threads = 0` looks hacky, but I don't want to introduce another config option for background toggling atm.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 - Declare `delayed_job` and `sidekiq` as integration gem's dependency [#1506](https://github.com/getsentry/sentry-ruby/pull/1506)
 - `DSN#server` shouldn't include path [#1505](https://github.com/getsentry/sentry-ruby/pull/1505)
 - Fix `sentry-rails`' `backtrace_cleanup_callback` injection [#1510](https://github.com/getsentry/sentry-ruby/pull/1510)
+- Disable background worker when executing rake tasks [#1509](https://github.com/getsentry/sentry-ruby/pull/1509)
+  - Fixes [#1508](https://github.com/getsentry/sentry-ruby/issues/1508)
 
 ## 4.6.1
 
--- a/sentry-ruby/examples/rake/Rakefile.rb
+++ b/sentry-ruby/examples/rake/Rakefile.rb
@@ -7,3 +7,7 @@ end
 task :raise_exception do
   1/0
 end
+
+task :send_message do
+  Sentry.capture_message("message from rake")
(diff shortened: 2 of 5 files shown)


y-yagi:
@st0012 Thanks for your quick fix! But, unfortunately, this can't fix my issue. 

`Rake::Application#top_level` calls before the initialization process of a Rails application. So if users configure `Sentry` inside an initializer(that mentioned in [doc](https://docs.sentry.io/platforms/ruby/guides/rails/)), `Sentry.initialized?` inside `Rake::Application#top_level` returns `false` always and not using `Sentry.get_current_hub.with_background_worker_disabled`.

st0012:
@y-yagi I see. I now see that the Rails rake task's initialization order is like this:

```
Rails rake_tasks block
top_level
Sentry SDK initializer
inside the task
```

I'll dig into this more.

st0012:
@y-yagi I've updated the fix. It should work for both vanilla rake and Rails rake tasks.

y-yagi:
@st0012 Thanks! I confirmed this fixes my issue.

st0012:
@y-yagi sorry for the trouble and thanks for helping me verify the fix 🙂

y-yagi:
@st0012 No problem! Thanks for your quick fix :heart:

st0012:
@y-yagi fyi it's been released in version `4.6.2`

PR #1514, opened 2021-07-26
Correctly pass arguments to a rake task

y-yagi:
## Description

Due to effect of #1509, arguments of Rake task are lost now. This patch pass arguments to Rake as before.

I'm sorry to miss this!

### DIFF
--- a/sentry-ruby/lib/sentry/rake.rb
+++ b/sentry-ruby/lib/sentry/rake.rb
@@ -20,10 +20,10 @@ module Rake
     alias orig_execute execute
 
     def execute(args=nil)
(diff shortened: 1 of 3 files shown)


sentry-ruby/lib/sentry/rake.rb
```diff
@@ -23,7 +23,7 @@ def execute(args=nil)
       return orig_execute unless Sentry.initialized? && Sentry.get_current_hub
```
st0012, line 23: `orig_execute` should receive the argument too

y-yagi: Thanks, I fixed!

st0012 (approved):
thanks for catching this and the PR 👍

Issue #1427, opened 2021-05-04
Stack level to deep error after upgrading to 4.4.0


st0012:
In the version `4.4.0`, we added support for capturing `net/http` requests as span/breadcrumbs. This feature requires [patching the class](https://github.com/getsentry/sentry-ruby/blob/master/sentry-ruby/lib/sentry/net/http.rb), which can be done in 2 different but conflicting ways: method aliasing and module prepending. Here's a [great article from bearer](https://blog.bearer.sh/module-overrides-ruby-prepend-alias/) explaining how these 2 approaches conflict with each other.

Fortunately, the community has been adopting the `prepend` approach for a while now + `sentry-ruby` requires Ruby 2.4+. So the chance of having the issue should be low.

### How we're going to help you upgrade

With that being said, if you're reading this issue, chances are you're having the conflict inside your app. So here's our plan to help you upgrade:

1. We'll see if it's possible to align your other dependencies toward the `prepend` approach. Most of the gems that patch this library should have provided an option to switch the mode to `prepend`. For example, [`rack-mini-profiler`](https://github.com/MiniProfiler/rack-mini-profiler) and its [`prepend_net_http_patch` option](https://github.com/MiniProfiler/rack-mini-profiler#nethttp-stack-level-too-deep-errors).
2. If the above option is not viable to you, we'll decide whether which approach to take based on your feedback:
    - Provide a way to `opt-out` this patch.
    - Provide an option to switch modes like some other gems do.

### Update

The version `4.4.1` should solve the issue.

st0012:
I just checked `scout_apm`'s implementation and it looks like users can only solve the problem by disabling their `net/http` instrumentation. I also checked their instrumentation implementations, which mostly (if not all) use the alias approach.

Consider `scout_apm` is one of the most popular apm services in the community, I will add an option for users to switch between `prepend` and `alias` mode. Something like: `prepend_net_instrumentation: boolean`.

st0012:
I think changing the timing of prepend could also solve the problem without introducing another patching method. I've implemented it in https://github.com/getsentry/sentry-ruby/pull/1432 and will cut a release soon.

st0012:
`sentry-ruby 4.4.1` is out now. I hope that solve the issue.

mjobin-mdsol:
I was hit by similar problem with redis instrumentation between opentelemetry and scout. they have migrated to using prepend in redis, but this issue tells me they should do it everywhere...

[...]

PR #1517, opened 2021-07-28
Extend Rake with a more elegant and reliable way

knapo:
Extend Rake with a more elegant and reliable way

Closes https://github.com/getsentry/sentry-ruby/issues/1520

## Description

Per https://github.com/getsentry/sentry-ruby/commit/4dc603b4f45eb2f34ed2b283fa3084360b067f32#r54079829

Overwriting methods with aliasing is is rather a bad practice and, since it may cause conflicts, circular dependencies and infinite loops. `Module#prepend` is more elegant and reliable.

### DIFF
--- a/sentry-ruby/lib/sentry/rake.rb
+++ b/sentry-ruby/lib/sentry/rake.rb
@@ -4,7 +4,7 @@ require "rake/task"
 module Sentry
   module Rake
     module Application
-      def display_error_message(*)
+      def display_error_message(ex)
         Sentry.capture_exception(ex, hint: { background: false }) do |scope|
           task_name = top_level_tasks.join(' ')
           scope.set_transaction_name(task_name)
@@ -16,7 +16,7 @@ module Sentry
     end
 
     module Task
-      def execute(*)
+      def execute(args=nil)
         return super unless Sentry.initialized? && Sentry.get_current_hub
 
         Sentry.get_current_hub.with_background_worker_disabled do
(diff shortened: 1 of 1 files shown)


st0012:
As I mentioned in https://github.com/getsentry/sentry-ruby/commit/4dc603b4f45eb2f34ed2b283fa3084360b067f32#r54080150, using `prepend` doesn't guarantee conflict-free.

> in fact, I used prepend over alias when patching Net::HTTP. but the conflicts aren't that few, you can see #1427 for more detail.

> I choose alias for the rake integration because display_error_message was patched this way since the sentry-raven era and it seemed to work fine. so I didn't want to change it and potentially introduce another set of conflicts.

Can you tell me what caused the conflicts in your application?

knapo:
@st0012 

> Can you tell me what caused the conflicts in your application?

My other gem extends `Rake::Task` with prepend, what doesn't cooperates with aliasing and getting a `stack level too deep (SystemStackError)`.  It's basically same case as in https://gist.github.com/knapo/d21070a00e870626087516530dd4cfda.

sentry-ruby/lib/sentry/rake.rb
```diff
@@ -1,30 +1,31 @@
 require "rake"
 require "rake/task"
 
-module Rake
-  class Application
+module Sentry
+  module Rake
+    module Application
+      def display_error_message(*)
```
st0012, line 7: shouldn't this method take `ex` as the argument?

knapo: yes, we need to do that to make it working with ruby 2.4, let me fix that

st0012 (approved):
Thanks for the fix 👍

PR #1523, opened 2021-08-04
SDK should drop the event when any event processor returns nil

st0012:
This is something we should've supported but was missed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+- SDK should drop the event when any event processor returns nil [#1523](https://github.com/getsentry/sentry-ruby/pull/1523)
+
 ## 4.6.4
 
 - Extend Rake with a more elegant and reliable way [#1517](https://github.com/getsentry/sentry-ruby/pull/1517)
--- a/sentry-ruby/lib/sentry/client.rb
+++ b/sentry-ruby/lib/sentry/client.rb
@@ -26,7 +26,12 @@ module Sentry
     def capture_event(event, scope, hint = {})
       return unless configuration.sending_allowed?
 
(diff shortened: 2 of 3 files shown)


st0012:
@rhcarvalho can you give this a quick look today? thanks 😄

rhcarvalho (commented):
Late LGTM 👍 

By the way, is the SDK implementing the same behavior for global event processors and `beforeSend`?

st0012:
What're global event processors? I thought they're always attached to a scope? 

It's implemented with `before_send` though.

According to the documentation, there are 5 steps in the event pipeline:

1. If the SDK is disabled, Sentry discards the event right away.
2. The client samples events as defined by the configured sample rate. Events may be discarded randomly, according to the sample rate.
3. The scope is applied, using apply_to_event. The scope’s event processors are invoked in order.
4. Sentry invokes the before-send hook.
5. Sentry passes the event to the configured transport. The transport can discard the event if it does not have a valid DSN; its internal queue is full; or due to rate limiting, as requested by the server.


4 Is already implemented and this PR fixes 3.

rhcarvalho:
Some SDKs have, in addition to event processors added to Scope, event processors on Client and global. Those event processors are the same type of functions as beforeSend, the difference being that there can be multiple of them.

Most integrations use event processors to modify or drop events.

If the Ruby SDK doesn't have other event processors, that's okay and I suggest keeping things simple and not adding anything more.

My question was in the sense to make sure we fixed all similar issues related to the behavior of dropping events, it seems we did. Thanks! 🙏

st0012:
I see. Thanks for explaining 👍

PR #1531, opened 2021-08-13
Add monotonic_active_support_logger

morissetcl:
Hi, 

This is a proposal to support `monotonic_time`.

 The regular `subscriber` use `real_time` and can jump forwards and backwards as the system time-of-day clock is changed. It can be confusing, especially when computing elapsed time.

~~Few years ago~~ Two years ago Rails add a `monotonic_subscriber` for instrumentation, I think it will be great to add the ability to set it.

There is other places in the code where it could be propagated. (especially for tracing)
Just would like have your feelings about this option and if it could make sense for you.

BTW thanks for maintaining this gem it's really useful.


sentry-rails/spec/sentry/rails/active_support_breadcrumbs_spec.rb
```diff
@@ -88,7 +92,7 @@
         ActiveSupport::Notifications.publish "foo", Object.new
       end.not_to raise_error
 
-      expect(transport.events).to be_empty
+      expect(breadcrumb_buffer.count).to be_zero
```
morissetcl, line 95: While testing I figure out that this test is not really reliable:
If you run 

```
    it "ignores events that doesn't have a started timestamp" do
      expect do
        ActiveSupport::Notifications.publish "foo", Time.now
      end.not_to raise_error

      expect(transport.events).to be_empty
    end
```

It will pass.
So I use `breadcrumb_buffer` to check if we record or not an event. Maybe we can do better, let me know.

st0012:
@morissetcl thanks for the proposal. I didn't know about the `monotonic` feature until now!

but it looks like this feature is only added after Rails 6.1 (commit: https://github.com/rails/rails/commit/93b652affbed41bce568a0fac4ca7b56aa3b691f). and having a global config only for a certain Rails version is confusing.

so I want to propose another idea: you can add a new breadcrumb logger: `monotonic_active_support_logger` for this support (which will print warnings + not being activated if the client's Rails version < 6.1). and the user can just use the `config.breadcrumb_loggers` option for this feature too.

morissetcl:
@st0012 hmm excellent point. I'm going to make these changes.

What about https://github.com/getsentry/sentry-ruby/pull/1531#discussion_r688756659 ? Any thought ? If changes are necessary I propose to do them in a separate PR.

st0012:
> What about #1531 (comment) ? Any thought ? If changes are necessary I propose to do them in a separate PR.

Ah that's a good catch. Thank you 👍

sentry-rails/spec/sentry/rails/breadcrumbs/active_support_breadcrumbs_spec.rb
```diff
@@ -0,0 +1,91 @@
+require "spec_helper"
```
st0012, line 1: is this a copy of the old test file? if it is, shouldn't the old one get deleted?

morissetcl: Oups 🤦🏻  
Yes it's a copy, thanks it's done!

st0012:
@morissetcl oops, looks like #1535 caused an conflict on this. can you fix it?

st0012 (approved):
Thank you 👍

PR #1535, opened 2021-08-15
Make ignoring events test more reliable

morissetcl:
While testing this https://github.com/getsentry/sentry-ruby/pull/1531 I figured out that a test was not really reliable.

If you run:

```
    it "ignores events that doesn't have a started timestamp" do
      expect do
        ActiveSupport::Notifications.publish "foo", Time.now
      end.not_to raise_error

      expect(transport.events).to be_empty
    end
```

It will pass.
So I use `breadcrumb_buffer` to check if we record or not an event.

You can have additional information here https://github.com/getsentry/sentry-ruby/pull/1531#discussion_r688756659

Thanks.

### DIFF
--- a/sentry-rails/spec/sentry/rails/active_support_breadcrumbs_spec.rb
+++ b/sentry-rails/spec/sentry/rails/active_support_breadcrumbs_spec.rb
@@ -22,6 +22,10 @@ RSpec.describe "Sentry::Breadcrumbs::ActiveSupportLogger", type: :request do
     Sentry.get_current_client.transport
   end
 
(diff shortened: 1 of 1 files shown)


st0012:
@morissetcl thanks for the fix 👍 
but we don't include non-SDK change into the changelog. can you remove the commit?

morissetcl:
> but we don't include non-SDK change into the changelog. can you remove the commit?

Make sense, sorry first time contribution :) 
I am going to remove it.

st0012 (approved):
Thank you for the fix!

PR #1542, opened 2021-08-19
fix typo in migration guide

yukimurasawa:
## Description

Hi, 

I just found typo in migration guide.
I think `Raven.tags_context` is right.


st0012 (approved):
good catch. thanks 👍

PR #1555, opened 2021-09-01
Add root license

st0012:
GitHub can't capture the repo's license from a link file. So I think we need to give it a copy of the actual license.


rhcarvalho (approved):
Thanks @st0012 !

Most other SDKs call this file LICENSE (without extension), if it makes no difference otherwise I'd rename the file to match the other SDKs.

st0012:
@rhcarvalho 👍 updated. merging now.

Issue #1556, opened 2021-09-01
report_after_job_retries does not report errors for jobs with 0 retries


TAGraves:
**Describe the bug**
I was excited to see the `report_after_job_retries` feature get released in sentry-sidekiq this morning. Unfortunately we have many jobs configured with `sidekiq_options retry: 0` and failures in these jobs do not appear to be sent to Sentry. Thankfully the fix looks pretty simple so I will open a PR with an attempt at fixing it.

**To Reproduce**
Set `config.sidekiq.report_after_job_retries = true` and then configure a job that will fail with `sidekiq_options retry: 0`. Enqueue that job and observe that it ends up in the Sidekiq dead queue and does not report the error to Sentry.

**Expected behavior**
The error should be reported to Sentry.

**Actual behavior**
The error is not reported to Sentry.

**Environment**
 - Ruby Version: 3.0.2
 - SDK Version: 4.7
 - Integration Versions (if any): Sidekiq 6.2.2

**Sentry Config**

```ruby
Sentry.init do |config|
  config.sidekiq.report_after_job_retries = true
end
```

st0012:
@TAGraves your fix has been released now. thanks again 🙂

PR #1557, opened 2021-09-01
Fixes report_after_job_retries for retry: 0

TAGraves:
## Description
This fixes https://github.com/getsentry/sentry-ruby/issues/1556 by checking to see if retry_option is positive.

The issue is that `retry_count` is nil when `retry: 0` is set. Since retries will only occur if `retry_option` is positive we can skip this altogether in the case where it is not positive.


st0012:
@TAGraves thanks for reporting the issue and making the fix!
can you change the `Unreleased` title to `4.7.1`? I'll make a release right after this is merged 🙂

TAGraves:
@st0012  done, thanks!

st0012 (approved):
👍

PR #1567, opened 2021-09-08
Only report one test coverage per workflow

st0012:
Currently every CI job sends a coverage report to codecov, which generates about 40 reports per CI run. This makes CI's status highly depend on codecov's because if it can't accept a report request (due to server issue...etc.), it fails the CI job.


st0012:
@BYK I want to rename workflow names so they can provide more information about the job. Can you help me cancel the PR check for a while so I can merge this PR?

BYK:
Will do tomorrow :)

PR #1578, opened 2021-09-20
Allow tag values to be callables and allow default tags

ojab:
Right now it's difficult to add default tags to events and sometimes this tags should be evaluated at runtime, for example if different APM is used and foreign trace id is desired in all the events.

Alternative could be using specific middlewares or rails's `before_action`, but it quickly becomes an issue because it should be done for every single piece of the app (`ActionController`, `ActiveJob`, … rails part and I'm not even talking about using sidekiq directly, for example) and it should be carefully placed just after trace id allocation to prevent missing trace ids.

Rails tagged logging uses the same mechanism for `config.log_tags`, but passes `request` there and allows to use symbols as a `request` method name. Here we're not passing anything and just calling the value proc.


st0012:
I assume you added global default tag support with the intention to add a callable tag? If not I'm not seeing any reason for adding a static global tag.

And regarding callable tag support, if what you want is to have an event's tag lazily evaluated right before it's serialized with some code, I recommend you to use the SDK's event processors feature. It works like this:

```ruby
Sentry.get_current_scope.add_event_processor do |event, hint|
  event.tags << SomeCode.generate_tags
  event # remember to return the event
end
```

This can be put anywhere after the SDK is initialized and the processor will be applied to all events. So you only need to write it once.

ojab:
Looks like it will do the trick, but it's not documented and not sure if it's not cleared anywhere & passed to the inner scopes. I'll check later today and probably close this PR.

st0012:
@ojab I hope you already solved your problem with event processors. given that we don't plan to introduce callable tags, I'm closing this now.

kml:
Hi, @st0012. If I want to create a gem registering even processor for *all* scopes, how could I do it using the mentioned solution?

What I'm afraid of is that 'current scope` may change and my processor won't be registered for all scopes (https://github.com/getsentry/sentry-ruby/blob/3ddb146ed2feaf2d2c078c979629ab567a2701d8/sentry-ruby/lib/sentry/hub.rb#L66)
For a simplified use case, I'd like to add a single initializer file to the application that would register event processor for all scopes (not only the current one). 

Do you plan feature like `Sentry.get_current_hub.add_event_processor`? As hub is an entity shared between all scopes, right?
Or maybe you have some other idea of solving this issue? Like registering to new scope creation so then processor could be added dynamically by the subscriber.

Some hacky workaround could be:

```ruby
Sentry::Scope.prepend(Module.new do
  def initialize(*)
    super
    
    add_event_processor do |event, hint|
      event.tags << SomeCode.generate_tags
      event # remember to return the event
    end
  end
end)
```

kml:
To give more context - I'd like to log https://www.rubydoc.info/gems/newrelic_rpm/NewRelic%2FAgent%2FTracer%2Ecurrent_trace_id as tag on event.
As it's hold in fiber-local variable I'd like it to be applied to current scope as early as possible to don't loose information.

st0012:
@kml I just checked how other SDKs handle this type of cases. And `sentry-javascript` seems to have "global event processors", which will be stored directly under the `Sentry` namespace. So I'm thinking about APIs like:

```rb
Sentry.add_global_event_processor { ... }
Sentry.global_event_processors # this will be an array, so you can also do
Sentry.global_event_processors << my_proc
```

Wdyt?

kml:
@st0012 Looks great! For my case it only depends when exactly it will be evaluated (if data I need would be still available).
I'll try to do some testing using existing solution and get back with feedback.

Simplified version, of what I try to do now looks like:

```ruby
app.config.middleware.insert_after(
  Sentry::Rails::CaptureExceptions,
  MyMiddleware
)

class MyMiddleware
  def initialize(app)
    @app = app
  end

  def call(env)
   Sentry.set_tags(trace_id: NewRelic::Agent::Tracer.current_trace_id) if Sentry.initialized?

   @app.call(env)
  end
end
```

While it needs similar solution for any kind of process (like worker processes).

BTW: I talk about same code base as @ojab, so it's not yet escalation in the area of dynamic tags/global handler :wink:

st0012:
I checked your comments and the example again, and I'm not sure if I understand your needs.

> As it's hold in fiber-local variable I'd like it to be applied to current scope as early as possible to don't loose information.

So it looks like the middleware is a perfect solution to this issue. Why do you want to use event processor in this case?

> While it needs similar solution for any kind of process (like worker processes).

If the other place that needs this is a worker process, you can just add a similar code to that using callbacks or middlewares (if you use Sidekiq).

I can understand that you may want to have a global callback (event processors) to save you all these work. But because the event processors will only be called before an event is sent, which could vary a lot depend on your usages, it probably won't save you much work and could cause more trouble.

kml:
> So it looks like the middleware is a perfect solution to this issue.

OK. Great to hear that :)

> Why do you want to use event processor in this case?

It's about understanding if that's possible. For example, I also use [SemanticLogger](https://github.com/reidmorrison/semantic_logger/blob/master/lib/semantic_logger/logger.rb#L8) and just refactored the solution partially from middleware to a single registered subscriber (analog to processor):

```ruby
SemanticLogger::Logger.subscribe do |log|
  log.payload[:new_value] = ThisGenerate.value
end
```

This is somehow future-proof - will work for web, sidekiq, or any other new integration.
But... it's for logs, not exceptions,
so they (logs) are built and processed close to the moment when they are created (so evaluated in the same context).
The exception is propagated up and reported "far" from the place where it was raised.

> can just add a similar code to that using callbacks or middlewares

:+1: Right, I solved it exactly the same way.

> But because the event processors will only be called before an event is sent, which could vary a lot depend on your usages, it probably won't save you much work and could cause more trouble.

Exactly what I'm afraid of. With dynamic tag adding `Time.now` or assigning random UUID benefit of having global handler would be more clear.

@st0012 :bow: for your opinion on that. It's much more clear for me now, and I'll stay with middlewares then.

st0012:
@kml no problem. glad to help 😄

PR #1599, opened 2021-10-28
sentry-sidekiq pass extras from client to server

pacoguzman:
## Description

We have the need to pass through a set of extras one of them is the `x-request-id` which generated the sidekiq job, so we find useful to be able to pass this through.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
   This feature should only introduce negligible performance overhead in most Ruby applications. But if you notice obvious performance regression, please file an issue and we'll investigate it.
 
 - Support `ActiveStorage` spans in tracing events [#1588](https://github.com/getsentry/sentry-ruby/pull/1588)
+- Support sentry extras in sidekiq jobs [#1599](https://github.com/getsentry/sentry-ruby/pull/1599](https://github.com/getsentry/sentry-ruby/pull/1599)
 
 ### Miscellaneous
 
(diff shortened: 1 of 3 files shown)


st0012:
> just middleware that are setting global information that current user, current request-id, current transaction, current opentelemetry-trace, etcetera.

I think you mixed two types of data together here. Request id, transaction, trace id, and sometimes user data, can be used as identifier to connect different events. So having them across events isn't going to cause confusion. I'm totally fine with passing this type of data around.

But this PR will propagate "all" extra data to the downstream job, which can contain very contextual data like the `batch_count` mentioned in the example. This will be a new behavior to all the SDK users and if it bothers them, I don't think I can just tell them to get used to it.

What I proposed was, when upstream data like `batch_count`, is added to the downstream job, its key should indicate that. For example, it'll become `upstream.batch_count` so that users can know it doesn't come from the current context. 

Also, I know some clients that have a dozen entries of extra data. To them, having the amount of data added to the downstream job can also be annoying. So I think this should be an opt-in feature. It should only work when the client set something like `config.sidekiq.propagate_contextual_data = true`.

pacoguzman:
OK, I understood that now, just could be confusing. I'm not sure if we can set up at sidekiq level some kind of context_propagator which the user can tweak, so it can decide which keys must be propagated as `extra` and which ones filter out when propagating. So defining that implies the user want to propagate something, this is kind of extension to your idea about `config.sidekiq.propagate_contextual_data = true`.

By the way, if this gets a little bit more complex I would say users will end defining their own sidekiq middlewares that they can tweak as much as they can to propagate their context, which could include setup sentry.extra too. Probably we're going to give it a try to this approach while we discuss the sentry support.

st0012:
> By the way, if this gets a little bit more complex I would say users will end defining their own sidekiq middlewares

Yeah, so I think we only need to let users decide whether to propagate data or not. If they want to manually select which data to propagate, they'll need to do it on their own. (Or if the demand is high we can consider that support later). 

So my opinion is that we can add this feature to the SDK if there's a way user can opt-out from it.

PR #1601, opened 2021-10-31
Avoid duplicated sampling on Transaction events

st0012:
As @sl0thentr0py pointed out on discord, having `sample_allowed?` inside `sending_allowed?` causes double sampling on transaction events and should be avoided. So this PR fixes the issue and removes an unnecessary sending check from the `Transport` class.

### DIFF
--- a/sentry-ruby/lib/sentry/client.rb
+++ b/sentry-ruby/lib/sentry/client.rb
@@ -25,7 +25,7 @@ module Sentry
 
     def capture_event(event, scope, hint = {})
       return unless configuration.sending_allowed?
-      return unless configuration.sample_allowed? || event.is_a?(TransactionEvent)
+      return unless event.is_a?(TransactionEvent) || configuration.sample_allowed?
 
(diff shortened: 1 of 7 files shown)


sentry-ruby/lib/sentry/client.rb
```diff
@@ -25,6 +25,7 @@ def initialize(configuration)
 
     def capture_event(event, scope, hint = {})
       return unless configuration.sending_allowed?
+      return unless configuration.sample_allowed? || event.is_a?(TransactionEvent)
```
sl0thentr0py, line 28: maybe the other way round so we short circuit the rng call completely if it's a transaction?
```suggestion
      return unless event.is_a?(TransactionEvent) || configuration.sample_allowed?
```

sl0thentr0py (approved):
thx for the quick fix @st0012!
Looks good to my newbie eyes, just one minor optimization suggestion but can also ignore

PR #1604, opened 2021-11-05
feat: Add client reports

sl0thentr0py:
This PR adds support for [Client Reports](https://develop.sentry.dev/sdk/client-reports/).

* new config `send_client_reports`, true by default
* record lost events in `Transport` 
* attach them to an existing envelope while sending, max 1 per 30 seconds
  * PS: this differs from python which needs to do extra scheduling when it sends events via the `store` endpoint

Some notes:
* I used minitest before, so if I did something incorrect while writing the specs, please let me know :)
* I still want to test this a bit more manually since it's my first PR, so making draft for now


sentry-ruby/lib/sentry/transport.rb
```diff
@@ -5,16 +5,24 @@ module Sentry
...
     USER_AGENT = "sentry-ruby/#{Sentry::VERSION}"
+    CLIENT_REPORT_INTERVAL = 60
 
     include LoggingHelper
 
+    attr_accessor :configuration
     attr_reader :logger, :rate_limits
 
     def initialize(configuration)
+      @configuration = configuration
```
st0012, line 16: can we only store the `send_client_reports` attribute instead of the entire configuration object?

sentry-ruby/spec/sentry/client/event_sending_spec.rb
```diff
@@ -34,6 +34,7 @@
 
         it "doesn't send the event when it's not sampled" do
           allow(Random).to receive(:rand).and_return(0.51)
+          expect(subject.transport).to receive(:record_lost_event).with(:sample_rate, 'event')
```
st0012, line 37: can we test this (and below scenarios) end-to-end instead of using mocking? you can even write a custom helper like `have_recorded_lost_event`. so it'll be like:

```rb
# this will check the value in @discarded_events
expect(transport).to have_recorded_lost_event(:sample_rate, 'event') 
```

sentry-ruby/spec/sentry/client/event_sending_spec.rb
```diff
@@ -156,6 +157,17 @@
           expect(transport.events.count).to eq(1)
         end
       end
+
+      it "records queue overflow" do
+        allow(Sentry.background_worker).to receive(:perform).and_return(false)
+        expect(subject.transport).to receive(:record_lost_event).with(:queue_overflow, anything)
```
st0012, line 163: same as the above

sentry-ruby/spec/sentry/client/event_sending_spec.rb
```diff
@@ -267,6 +279,8 @@
         end
 
         it "discards the event and logs a info" do
+          expect(subject.transport).to receive(:record_lost_event).with(:event_processor, anything)
```
st0012, line 282: same as the above

sentry-ruby/spec/sentry/client/event_sending_spec.rb
```diff
@@ -307,6 +321,7 @@
         end
 
         it "swallows and logs Sentry::ExternalError (caused by transport's networking error)" do
+          expect(subject.transport).to receive(:record_lost_event).with(:network_error, anything)
```
st0012, line 324: same as the above

sentry-ruby/spec/sentry/client/event_sending_spec.rb
```diff
@@ -330,6 +345,7 @@
...
```
st0012, line 348: same as the above

sentry-ruby/spec/sentry/client/event_sending_spec.rb
```diff
@@ -418,6 +434,20 @@ class ConnectionError < StandardError; end
...
+
+      context "before_send returns nil" do
+        before do
+          configuration.before_send = lambda do |_event, _hint|
+            nil
+          end
+        end
+
+        it "records lost event" do
+          expect(subject.transport).to receive(:record_lost_event).with(:before_send, anything)
```
st0012, line 446: same as the above

sentry-ruby/spec/sentry/transaction_spec.rb
```diff
@@ -371,6 +371,11 @@
 
         expect(events.count).to eq(0)
       end
+
+      it "records lost event" do
+        expect(subject.hub.current_client.transport).to receive(:record_lost_event).with(:sample_rate, 'transaction')
```
st0012, line 376: same as the above

sentry-ruby/spec/sentry/transport_spec.rb
```diff
@@ -126,6 +156,17 @@
...
       end
     end
+
+    context "when rate limited" do
+      before do
+        allow(subject).to receive(:is_rate_limited?).and_return(true)
+      end
+
+      it "records lost event" do
+        expect(subject).to receive(:record_lost_event).with(:ratelimit_backoff, 'event')
```
st0012, line 166: same as the above

sentry-ruby/lib/sentry/transport.rb
```diff
@@ -83,27 +92,85 @@ def generate_auth_header
...
       # Convert to hash
-      event_hash = event.to_hash
+      event_payload = event.to_hash
 
-      event_id = event_hash[:event_id] || event_hash["event_id"]
-      item_type = get_item_type(event_hash)
+      event_id = event_payload[:event_id] || event_payload["event_id"]
+      item_type = get_item_type(event_payload)
+
+      envelope_header = {
```
st0012, line 100: nice refactor 👍

sentry-ruby/lib/sentry/transport.rb
```diff
@@ -5,16 +5,24 @@ module Sentry
...
 
     def initialize(configuration)
+      @configuration = configuration
       @logger = configuration.logger
       @transport_configuration = configuration.transport
       @dsn = configuration.dsn
       @rate_limits = {}
+
+      if configuration.send_client_reports
+        @discarded_events = Hash.new(0)
```
st0012, line 23: is this different than direct hash declaration `{}`?

st0012 (commented):
Generally looks good. Just some suggestions on styling and testing 👍

sentry-ruby/lib/sentry/transport.rb
sl0thentr0py: it sets a default value of 0 so you don't have to do `@hash[key] ||= 0` every time.

sentry-ruby/lib/sentry/transport.rb
sl0thentr0py: I thought of this as a good intermediate intermediate step to having `Envelope` and `Item` classes later.

sentry-ruby/lib/sentry/transport.rb
st0012: ah so it's like passing the default proc. thanks!

sl0thentr0py:
hmm I can finish making the changes and testing by tomorrow I think, so unless I find some problems, I'd say yes?

st0012:
@sl0thentr0py awesome. I'm thinking about releasing version `4.8.0` with this feature this week (align with RubyConf).

sl0thentr0py:
@iker-barriocanal 
short answer: no it doesn't send it's own envelope currently, it is always attached to an existing envelope.

---
long answer: 
If the actual requirement is to always ensure 1 client report per minute, that's not so trivial because you would need some coordinating parent process that ensures this scheduling logic. AFAICT, none of the sdks implement this in that sense.

What the other 2 sdks do:
* js only flushes them on browser tab state changing [(`visibilitychange`)](https://github.com/getsentry/sentry-javascript/blob/6b278f1ee5534736503c8d5c5eea293796dcd962/packages/browser/src/transports/base.ts#L57-L60)
* python adds them to an [existing envelope once every 30 seconds](https://github.com/getsentry/sentry-python/blob/ed4ba68cad42ebfbab162b37bf7edad25ebeae55/sentry_sdk/transport.py#L371) but also schedules one every [60 seconds](https://github.com/getsentry/sentry-python/blob/ed4ba68cad42ebfbab162b37bf7edad25ebeae55/sentry_sdk/transport.py#L300)

However, this 60 second scheduling is itself done only in [`capture_event`](https://github.com/getsentry/sentry-python/blob/ed4ba68cad42ebfbab162b37bf7edad25ebeae55/sentry_sdk/transport.py#L454) and [`capture_envelope`](https://github.com/getsentry/sentry-python/blob/ed4ba68cad42ebfbab162b37bf7edad25ebeae55/sentry_sdk/transport.py#L471) which implies that you eventually need some event/transaction to be fired to schedule this 'client report only' envelope. 

Another difference: python has both `capture_event` which goes to the `store` endpoint and `capture_envelope` which goes to the `envelope` endpoint, hence the need for the separate envelope scheduling. Ruby sends everything to `envelope` anyway so I'm not sure we need to schedule the other one at all.

For now, considering my current understanding of what we need client reports for, I think this implementation for ruby is fine. Let me know if you have questions.

sentry-ruby/lib/sentry/transport.rb
```diff
@@ -83,27 +92,85 @@ def generate_auth_header
...
     end
 
+    # valid reasons are
+    #   :ratelimit_backoff
+    #   :queue_overflow
+    #   :cache_overflow - NA
+    #   :network_error
+    #   :sample_rate
+    #   :before_send
+    #   :event_processor
```
AbhiPrasad, line 128: Should we link to the develop docs?

Also what do you think about enforcing an allow list here for valid reasons?

sentry-ruby/lib/sentry/transport.rb
```diff
@@ -83,27 +92,85 @@ def generate_auth_header
...
+      event_header = { type: item_type, content_type: 'application/json' }
 
       envelope = <<~ENVELOPE
-        {"event_id":"#{event_id}","dsn":"#{@dsn.to_s}","sdk":#{Sentry.sdk_meta.to_json},"sent_at":"#{Sentry.utc_now.iso8601}"}
-        {"type":"#{item_type}","content_type":"application/json"}
-        #{JSON.generate(event_hash)}
+        #{JSON.generate(envelope_header)}
+        #{JSON.generate(event_header)}
+        #{JSON.generate(event_payload)}
+        #{fetch_pending_client_report}
```
AbhiPrasad, line 113: Should we open a GH issue about implementing an envelope/envelope item class?

Even if we decide not to pursue it, we can leave discussion and reasoning in there.

iker-barriocanal:
@sl0thentr0py sure, I was just asking the question, LGTM as it is. The only thing I'd add to this conversation is to explicitly mention it somewhere, and IMO the short answer in the PR description should do it (or be more explicit on the third point).

sentry-ruby/lib/sentry/transport.rb
sl0thentr0py: https://github.com/getsentry/sentry-ruby/issues/1611

sl0thentr0py:
Tested the sample rate case end-to-end, looks good to me
Screenshot 2021-11-09 at 13 22 22

sentry-ruby/spec/spec_helper.rb
```diff
@@ -44,6 +44,12 @@
   config.before(:each, rack: true) do
     skip("skip rack related tests") unless defined?(Rack)
   end
+
+  RSpec::Matchers.define :have_recorded_lost_event do |reason, type|
```
sl0thentr0py, line 48: @st0012 not an rspec pro so not sure if this is the right way to do this? let me know :)

sl0thentr0py (commented):
@st0012 changed the specs and made some minor changes based on various comments above, can you take it for another spin?

st0012 (approved):
Thanks for the amazing feature 👍

sentry-ruby/spec/spec_helper.rb
st0012: I think it looks good and I've also tested it locally 👍

PR #1616, opened 2021-11-12
Extract envelope construction logic from Transport

### LINKED ISSUE #1611: Implement Envelope/Item classes
It would be nice to move the [envelope creation](https://github.com/getsentry/sentry-ruby/blob/06e6d785c293659913608037edf57e77e026caa3/sentry-ruby/lib/sentry/transport.rb#L91-L95) logic into a separate class so we can add more items in the future to the same envelope.

st0012:
Closes #1611

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 4.8.1
+
+### Refactoring
+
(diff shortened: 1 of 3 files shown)


st0012:
@sl0thentr0py does this look like what you'd expect?

sl0thentr0py (approved):
@st0012 yes that looks great!

PR #1617, opened 2021-11-13
Gracefully shutdown background worker before the process exits

### LINKED ISSUE #1612: Synchronously drain async error reporting queue
If you are using a background queue like Resque, threads are killed after a job completes. In this scenario, I believe errors would fail to send to sentry.

Is there a way to synchronously drain the async error reporting queue? This could be added to a before exit hook on the resque side.

st0012:
Currently, when a process exits, SDK's background worker will discard all WIP sending tasks and cause lost of events. This is why we need to [disable background worker in rake integration](https://github.com/getsentry/sentry-ruby/blob/master/sentry-ruby/lib/sentry/rake.rb#L22-L24). And from the discussion in #1612, it seems to cause issues in the resque integration too.

Even though I want to avoid using the `at_exit` block, I don't think the responsibility of shutting down the background worker should fall on users. It should be handled by the SDK whenever possible. So in addition to providing a shutdown API in this PR, the SDK would also start the shutdown procedure when exiting the process.

Closes #1612

### DIFF
--- a/sentry-ruby/lib/sentry/rake.rb
+++ b/sentry-ruby/lib/sentry/rake.rb
@@ -5,7 +5,7 @@ module Sentry
   module Rake
     module Application
       def display_error_message(ex)
-        Sentry.capture_exception(ex, hint: { background: false }) do |scope|
+        Sentry.capture_exception(ex) do |scope|
           task_name = top_level_tasks.join(' ')
           scope.set_transaction_name(task_name)
           scope.set_tag("rake_task", task_name)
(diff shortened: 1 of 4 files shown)


sl0thentr0py (approved):
very cool! I was just explaining this to support the other day why his test code wasn't working because the parent process died without flushing.
Thanks for adding!

st0012:
@sl0thentr0py np 👍

sentry-ruby/lib/sentry/rake.rb
```diff
@@ -19,9 +19,7 @@ module Task
       def execute(args=nil)
         return super unless Sentry.initialized? && Sentry.get_current_hub
 
-        Sentry.get_current_hub.with_background_worker_disabled do
```
natikgadzhi, line 22: Looking at `execute` here — isn't it just not doing anything in this form? Looks like it falls back to `super` in every branch, no? /cc @st0012

PR #1624, opened 2021-11-20
Log errors happened in BackgroundWorker#perform

### LINKED ISSUE #1618: Background worker send fails if ActiveRecord defined but not correctly configured
### Issue Description

Our rails app doesn't yet use `ActiveRecord::Base` for anything, but has some code that relies on the ActiveRecord library. In dev/test we have basically the default `database.yml` file with sqlite for dev/test, but production had a misconfiguration (e.g. `database: ''`). 

Since we never call ActiveRecord::Base explicitly in production it never caused exceptions. Sentry messages were silently failing to send in production, but worked in dev. If I set `config.background_worker_threads = 0`, it would work in production.

The issue appears to be [this code ](https://github.com/getsentry/sentry-ruby/blob/7aab67cc4af6115488568982c1a8c5250d206d19/sentry-rails/lib/sentry/rails/background_worker.rb)from sentry-rails:
```
module Sentry
  class BackgroundWorker
    def perform(&block)
      @executor.post do
        # make sure the background worker returns AR connection if it accidentally acquire one during serialization
        ActiveRecord::Base.connection_pool.with_connection do
          block.call
        end
      end
    end
  end
end
```

When I correctly configured the dummy sqlite connection, it worked in production.

### Reproduction Steps

Default sentry-rails install with a misconfigured `database.yml`:
```
# sentry.rb
Sentry.init do |config|
  config.dsn = ENV["SENTRY_DSN"]
end
```

```
# database.yml
development:
  adapter: sqlite3
  # database: db/development.sqlite3
  # Bad configuration
  database: ''
  pool: 5
  timeout: 5000
```

Call `Sentry.capture_message("test")`

### Expected Behavior

"Test" message captured OR log or inline message explaining where it failed.

### Actual Behavior

No message at sentry.io. With sentry debug turned on, there are no indications of failure.

### Ruby Version

2.6.8

### SDK Version

4.8.0

Rails

### Sentry Config

```
Sentry.init do |config|
  config.dsn = "https://abcd@o71791.ingest.sentry.io/1234"
end
```

st0012:
I assumed the block passed to `BackgroundWorker#perform` should already handle exceptions (e.g. `Client#send_event`). But #1618 showed it's not entirely true. So this PR adds exception logging to background worker and provide a safer approach to monkey patch its method.

Closes #1618

### DIFF
--- a/sentry-rails/lib/sentry/rails/background_worker.rb
+++ b/sentry-rails/lib/sentry/rails/background_worker.rb
@@ -1,11 +1,9 @@
 module Sentry
   class BackgroundWorker
-    def perform(&block)
-      @executor.post do
-        # make sure the background worker returns AR connection if it accidentally acquire one during serialization
-        ActiveRecord::Base.connection_pool.with_connection do
-          block.call
-        end
+    def _perform(&block)
+      # make sure the background worker returns AR connection if it accidentally acquire one during serialization
+      ActiveRecord::Base.connection_pool.with_connection do
+        block.call
       end
     end
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 - Merge context with the same key instead of replacing the old value. [#1621](https://github.com/getsentry/sentry-ruby/pull/1621)
   - Fixes [#1619](https://github.com/getsentry/sentry-ruby/issues/1619)
 - Fix `HTTPTransport`'s `ssl` configuration [#1626](https://github.com/getsentry/sentry-ruby/pull/1626)
+- Log errors happened in `BackgroundWorker#perform` [#1624](https://github.com/getsentry/sentry-ruby/pull/1624)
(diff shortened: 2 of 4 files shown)


sentry-rails/lib/sentry/rails/background_worker.rb
```diff
@@ -1,6 +1,6 @@
 module Sentry
   class BackgroundWorker
-    def perform(&block)
+    def _perform(&block)
       @executor.post do
```
sl0thentr0py, line 4: since this is already inside the `@executor.post` in `perform`, I think we don't need it here anymore?

sl0thentr0py (changes requested):

sentry-rails/lib/sentry/rails/background_worker.rb
st0012: @sl0thentr0py great catch, thanks 👍

sl0thentr0py (approved):

PR #1626, opened 2021-11-21
Fix HTTPTransport's ssl configuration

st0012:
Transport's configuration allows users to configure ssl options with a Ruby hash:

```rb
configuration.transport.ssl = { key: value }
```

or with specific options:

```rb
configuration.transport.ssl_verification = false
configuration.transport.ssl_ca_file = file
```

But there's a bug that will always override the `ssl` (Hash) option's `verify` and `ca_file` values with specific options' values (usually `nil` in this case).

This bug seems to exist for years (from `sentry-raven` era), which makes me wonder if anyone really uses the `ssl` option 🤔 

The PR fixes the issue and adds tests for all transport configuration options.

### DIFF
--- a/sentry-ruby/lib/sentry/transport/http_transport.rb
+++ b/sentry-ruby/lib/sentry/transport/http_transport.rb
@@ -139,10 +139,10 @@ module Sentry
     end
 
     def ssl_configuration
-      (@transport_configuration.ssl || {}).merge(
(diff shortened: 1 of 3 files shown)


sl0thentr0py (approved):
> which makes me wonder if anyone really uses the ssl option
^^

lgtm!

Issue #1629, opened 2021-11-23
Exception data is lost in active support instrumentation events


Linuus:
### Issue Description

I'm having some issues using Sentry, ActiveJob and LogSubscribers.
We have an ActiveJob log subscriber which logs stuff in production. When an exception occurs in a job, we log it and the instrumentation event payload includes the Exception that was raised.
However, now that we are trying to move over to Sentry, the exception is no longer present in the instrumentation event payload. The reason for this is that here:
https://github.com/getsentry/sentry-ruby/blob/master/sentry-rails/lib/sentry/rails/active_job.rb#L36-L39 

All exceptions are rescued. The exception is only re-raised if there are no handlers for it. In our case, we have a retry_on handler so the error won't be re-raised. The issue with this though is that ActiveSupport::Notifications::Instrumenter is also rescuing any exceptions that happen and adds the exception information to the event payload, here: https://github.com/rails/rails/blob/8337645e852c7a2f4f112ea03db6fa4e00081c5e/activesupport/lib/active_support/notifications/instrumenter.rb#L25-L27
So, now that Sentry is rescuing the exception, AS instrumenter doesn't get any exception and we lose that information.

This is what the order of "rescues" are:

- [ActiveJob rescue](https://github.com/rails/rails/blob/main/activejob/lib/active_job/execution.rb#L47-L48)
  - [Instrumentation rescue](https://github.com/rails/rails/blob/8337645e852c7a2f4f112ea03db6fa4e00081c5e/activesupport/lib/active_support/notifications/instrumenter.rb#L25-L27)
    - [Sentry rescue](https://github.com/getsentry/sentry-ruby/blob/master/sentry-rails/lib/sentry/rails/active_job.rb#L36-L39 )

Is this expected behavior?

### Reproduction Steps

Add an active job log subscriber and log `event.payload[:exception]`
Add a job that has a `retry_on` handler.
Raise an error in the job that is handled by the `retry_on` handler.

### Expected Behavior

The exception is logged in the log subscriber.

### Actual Behavior

The exception is not logged in the log subscriber.

### Ruby Version

2.7.3

### SDK Version

4.8.0

### Integration and Its Version

Sentry Rails 4.8.0

st0012:
@Linuus thanks for reporting the issue. although it's not originally made for this issue, but I think #1631 may fix this as well. can you give it a try? thx

Linuus:
[...]

PR #1631, opened 2021-11-27
Use prepended method instead of around_perform for ActiveJob integration

### LINKED ISSUE #956: Calling the error handler twice when raising an exception in a rescue_from block
When using an ActiveJob Adapter which is not listed in

https://github.com/getsentry/raven-ruby/blob/ff59e93057ac8ac98603f8745f5024df7ed47dc2/lib/raven/integrations/rails/active_job.rb#L4

capture_and_reraise_with_sentry calls `rescue_with_handler`
https://github.com/getsentry/raven-ruby/blob/ff59e93057ac8ac98603f8745f5024df7ed47dc2/lib/raven/integrations/rails/active_job.rb#L21-L24

However, if the error handler (defined by a rescue_from) raises an error, the same error will be reported to the error handler itself again (double calling the error handler).

Curious what would be a proper workaround? If no workaround seems reasonable, can we provide a way to skip capture_and_reraise_with_sentry (or simply disable the `ActiveJobExtensions`)?

st0012:
Since the sdk registers its around_perform as the "first" callback, it needs to call all error handlers manually (which naturally should happen later) to see if it should report the exception.

Usually, this approach works well because if an exception will be handled later, handling it ealier doesn't make much of a difference.

However, as described in #956, if there's another exception happened inside one of the error handlers, it'll re-enter the error handler again in here: https://github.com/rails/rails/blob/38cb5610b1/activejob/lib/active_job/execution.rb#l50-l51

Of course, it may be possible to handle such case by adding more rescue block inside the `capture_and_reraise_with_sentry` method. But it'll force the entire exception handling flow to be performed inside the first `around_perform` block. And that's something we should avoid.

This also fixes a `rescue_from` callback issue in `Sentry::SendEventJob`.

Closes #956 and #1629

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Unreleased
+
+### Bug Fixes
+
+- Use prepended method instead of `around_perform` for `ActiveJob` integration [#1631](https://github.com/getsentry/sentry-ruby/pull/1631)
+  - Fixes [#956](https://github.com/getsentry/sentry-ruby/issues/956) and [#1629](https://github.com/getsentry/sentry-ruby/issues/1629)
+
 ## 4.8.1
 
 ### Bug Fixes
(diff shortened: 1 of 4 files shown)


sl0thentr0py (approved):
cool fix!

PR #1637, opened 2021-12-08
Fixup `Net::HTTP` breadcrumb url & span creation when `Net::HTTP.start` is used

ojab:

Right now we're not recording URL and creating single span for all the requests.

## Description
Not sure if body addition is a welcome change and if 100 is a good length, but  it would be useful for us.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Use prepended method instead of `around_perform` for `ActiveJob` integration [#1631](https://github.com/getsentry/sentry-ruby/pull/1631)
   - Fixes [#956](https://github.com/getsentry/sentry-ruby/issues/956) and [#1629](https://github.com/getsentry/sentry-ruby/issues/1629)
 - Fix `Net::HTTP` breadcrump url when using `Net::HTTP.new` [#1637](https://github.com/getsentry/sentry-ruby/pull/1637)
+- Fix trace span creation when using `Net::HTTP.start` [#1637](https://github.com/getsentry/sentry-ruby/pull/1637)
 
 ## 4.8.1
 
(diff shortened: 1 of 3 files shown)


ojab:
CI failure is unrelated and I will update changelog with PR number anyway, so sending to review.

ojab:
OTOH breadcrumbs are already limiting max message size since https://github.com/getsentry/sentry-ruby/pull/1465/files, so limiting body size could be dropped

st0012:
@ojab thanks for the PR. I think they're both good additions but we can't add those information directly as it may contain pii data. we should only add them when `Sentry.configuration.send_default_pii` is `true`, like

https://github.com/getsentry/sentry-ruby/blob/efcf170b5f6dd65c3b047825bddd8fde87fc6b7b/sentry-ruby/lib/sentry/interfaces/request.rb#L40-L44

ojab:
Yeah, I saw that and guessed that only incoming stuff is counted as PII.
I'll make a change to save query & body only with `send_default_pii` probably tomorrow then.

ojab:
Updated, CI is green.

ojab:
Also pushed commit to fix tracing to multiple requests in a single connection.

st0012:
@ojab it looks like you've made a few other changes to address different issues? since you're essentially rewriting the `Net::HTTP` patch, can you separate the changes for different issues/features? it'll help me review this more carefully, as it's an important feature of the SDK. thanks.

st0012:
also, the feature and fixes will likely go out separately (feature in `4.9.0` while bug fixes in `4.8.x`, depending on the impact).

ojab:
Left only bugfixes in this PR and moved query string & body addition along with fix for span name for `Sentry.configuration.send_default_pii = false` case to the new PR https://github.com/getsentry/sentry-ruby/pull/1642. Every change is split per-commit, so hope two PRs are ok.

ojab:
Hey, any changes to be done here?

st0012:
@ojab hey sorry I'm quite busy recently so haven't got time to check all the PRs yet.

st0012 (approved):
Thanks for the fix!

PR #1640, opened 2021-12-12
Add tracing support to `ActionCable` integration

st0012:
1. Add tracing support.
截圖 2021-12-12 20 21 06

2. Use `contexts` instead of `extras` for the additional data.
截圖 2021-12-12 20 19 58

### DIFF
--- a/sentry-rails/lib/sentry/rails/action_cable.rb
+++ b/sentry-rails/lib/sentry/rails/action_cable.rb
@@ -2,19 +2,39 @@ module Sentry
   module Rails
     module ActionCableExtensions
       class ErrorHandler
-        def self.capture(env, transaction_name:, extra_context: nil, &block)
-          Sentry.with_scope do |scope|
-            scope.set_rack_env(env)
-            scope.set_extras(action_cable: extra_context) if extra_context
-            scope.set_transaction_name(transaction_name)
+        class << self
+          def capture(env, transaction_name:, extra_context: nil, &block)
+            Sentry.with_scope do |scope|
+              scope.set_rack_env(env)
+              scope.set_extras(action_cable: extra_context) if extra_context
+              scope.set_transaction_name(transaction_name)
+              transaction = start_transaction(env, scope.transaction_name)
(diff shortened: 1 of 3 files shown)


sentry-rails/lib/sentry/rails/action_cable.rb
```diff
@@ -2,20 +2,40 @@ module Sentry
...
+              transaction = start_transaction(env, scope.transaction_name)
+              scope.set_span(transaction) if transaction
 
-            begin
-              block.call
-            rescue Exception => e # rubocop:disable Lint/RescueException
-              Sentry::Rails.capture_exception(e)
+              begin
+                block.call
+                finish_transaction(transaction, 200)
```
sl0thentr0py, line 16: This is more an open question. I'm not sure if assigning HTTP request semantics (here, status) to all websocket lifecycle events makes sense. 
I'm gonna check what python ASGI does for websockets and get back, but this can be merged and we can iterate later.

sl0thentr0py (approved):
LGTM!

sentry-rails/lib/sentry/rails/action_cable.rb
sl0thentr0py: ok nvm, python ASGI doesn't even set the status and for all practical purposes, websockets are currently treated similarly to HTTP requests so you can ignore this. :)

sentry-rails/lib/sentry/rails/action_cable.rb
```diff
@@ -2,20 +2,40 @@ module Sentry
...
+                finish_transaction(transaction, 500)
 
-              raise
+                raise
+              end
             end
           end
+
+          def start_transaction(env, transaction_name)
+            sentry_trace = env["HTTP_SENTRY_TRACE"]
```
lewispb, line 27: It's not possible to define HTTP headers on WebSocket connections from Javascript to specify the trace ID. And the Sentry JS library only seems to append the trace id to XHR's. So I'm not sure how to actually use this feature?

PR #1650, opened 2021-12-24
Be able to add event's tags to exception object

benoittgt:
Like for `sentry_context` we want to be able to define a method in our exception to provide additional tags to the Sentry event.

### DIFF
--- a/sentry-ruby/spec/sentry/client_spec.rb
+++ b/sentry-ruby/spec/sentry/client_spec.rb
@@ -6,6 +6,13 @@ class ExceptionWithContext < StandardError
       foo: "bar"
     }
   end
+
+  def sentry_tags
+    {
+      a: 'b',
+      c: 'd'
+    }
+  end
 end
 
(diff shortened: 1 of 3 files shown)


sentry-ruby/spec/sentry/client_spec.rb
```diff
@@ -6,6 +6,13 @@ def sentry_context
       foo: "bar"
     }
   end
+
```
benoittgt, line 9: Maybe `ExceptionWithContext` is not longer relevant?

st0012:
@benoittgt hey thanks for the PR! there's also a discussion around this in https://github.com/getsentry/sentry-ruby/pull/1101. can you try the approach I mentioned in this comment? https://github.com/getsentry/sentry-ruby/pull/1101#pullrequestreview-839905527

benoittgt:
Instead of this PR, I think it's better to follow the example you provided with the `before_send`.

I think this PR can be closed.

PR #1662, opened 2022-01-02
Explicitly passing Rack related configurations

st0012:
Since `RequestInterface` is directly initialized from `Event`, it's entirely possible to pass everything it needs with method calls. 

This change will make `RequestInterface`'s dependencies clearer and shouldn't break any user program (interfaces shouldn't be initialized manually).

### DIFF
--- a/sentry-ruby/spec/sentry/interfaces/request_interface_spec.rb
+++ b/sentry-ruby/spec/sentry/interfaces/request_interface_spec.rb
@@ -4,15 +4,11 @@ require 'spec_helper'
 
 RSpec.describe Sentry::RequestInterface do
   let(:env) { Rack::MockRequest.env_for("/test") }
+  let(:send_default_pii) { false }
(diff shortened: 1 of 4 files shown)


sentry-ruby/spec/sentry/interfaces/request_interface_spec.rb
```diff
@@ -4,15 +4,11 @@
...
   let(:env) { Rack::MockRequest.env_for("/test") }
+  let(:send_default_pii) { false }
+  let(:rack_env_whitelist) { Sentry::Configuration::RACK_ENV_WHITELIST_DEFAULT }
 
   subject do
-    described_class.build(env: env)
-  end
-
-  before do
-    Sentry.init do |config|
```
st0012, line 13: It also makes tests cleaner, as we don't need to initialize the entire SDK for testing an interface class.

sl0thentr0py (approved):

PR #1665, opened 2022-01-04
Merge back 4.8.2 changes

st0012:
I'm not sure if this is the right way to port `4-8` branch (currently `4.8.2`) specific changes back to master, which are:

- Versions
- Changelog Entries

I thought about rebasing master on `4-8`, which will yield a more natural git history. But it requires force pushing the master branch, which sounds like a big no-no to me. 

And the other way is this: merging `4-8` back to master with this intermediate branch to resolve conflicts. It don't require any force-push but the commit history **might** look ugly (old commits marked as recently committed..etc.). This is the first time I do branch-based release so I'm not really sure what'd happen next.

Any idea? @sl0thentr0py

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,18 +11,11 @@ When `config.send_default_pii` is set as `true`, `:http_logger` will include que
 
 ### Bug Fixes
 
-- Use prepended method instead of `around_perform` for `ActiveJob` integration [#1631](https://github.com/getsentry/sentry-ruby/pull/1631)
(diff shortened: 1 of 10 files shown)


st0012:
@sl0thentr0py I'm going to merge this now in order to sync changes and patch out https://github.com/getsentry/sentry-ruby/pull/1667

sl0thentr0py:
sry I just got up and saw this. I'm not sure about the process in this case either tbh.
I think having a few weird commits temporarily is fine in special cases.

st0012:
@sl0thentr0py Because we use squash and merge, so all the weird commits seen in this PR were actually ignored as they don't introduce any real change. And we just need to remove them from the squashed commit message as well.

PR #1667, opened 2022-01-05
Correctly return JobClass#perform's return value

### LINKED ISSUE #1666: Regression in 4.8.2: ActiveJobs return nil even if they have a return value
### Issue Description

When upgrading from 4.8.1 for sentry-{ruby, rails, sidekiq} to 4.8.2 all ActiveJob calls return nil.

I am not 100% sure why we have return values from these jobs and they are only used in tests.

### Reproduction Steps

```ruby
class SentryJob < ActiveJob::Base
  def perform
    "hello"
  end
end
```

```ruby
require "test_helper"

class SentryJobTest < ActiveSupport::TestCase
  test "#perform" do
    value = SentryJob.perform_now
    assert_equal("hello", value)
  end
end
```

### Expected Behavior

Should return the value from `perform`

### Actual Behavior

Returns `nil`

### Ruby Version

2.6.9

### SDK Version

4.8.2

- sentry-rails 4.8.2
- sentry-sidekiq 4.8.2
- Rails 5.2.6

### Sentry Config

```ruby
SENTRY_FILTER = %i[password more_params]

Sentry.init do |config|
  config.environment = "appname"
  config.traces_sample_rate = 0.1
  config.propagate_traces = false

  filter = if Rails::VERSION::MAJOR >= 6
    # Rails 6
    ActiveSupport::ParameterFilter.new(
      SENTRY_FILTER
    )
  else
    # Rails 5
    ActionDispatch::Http::ParameterFilter.new(
      SENTRY_FILTER
    )
  end

    filter.filter(event.to_hash)
  end
end
```

st0012:
Fixes #1666

### DIFF
--- a/sentry-rails/lib/sentry/rails/active_job.rb
+++ b/sentry-rails/lib/sentry/rails/active_job.rb
@@ -24,9 +24,11 @@ module Sentry
 
         scope.set_span(transaction) if transaction
 
-        block.call
+        return_value = block.call
 
         finish_sentry_transaction(transaction, 200)
+
+        return_value
       rescue Exception => e # rubocop:disable Lint/RescueException
         finish_sentry_transaction(transaction, 500)
 
--- a/sentry-rails/spec/sentry/rails/activejob_spec.rb
+++ b/sentry-rails/spec/sentry/rails/activejob_spec.rb
@@ -1,6 +1,12 @@
 require "spec_helper"
 require "active_job/railtie"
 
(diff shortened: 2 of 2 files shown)


davidwessman:
Thank you! Took the same time for me to refactor our use of the return value 🙂

st0012:
@davidwessman No problem. Thank you for reporting the issue 😄  
I'll need to sync changes between master and `4-8` branch though so it could take a bit longer to patch out a new version.

PR #1672, opened 2022-01-07
Remove incorrect backtrace attribute from Event

st0012:
Backtrace is never a standalone attribute of Event. It's either attached with a `ThreadsInterface` or a `ExceptionInterface`.

Given that this attribute never works and is misleading, I consider the removal a bug fix instead of breaking change.


sl0thentr0py (approved):

st0012:
@sl0thentr0py thanks for the quick review!

PR #1673, opened 2022-01-07
Refactor RequestInterface

st0012:
This PR removes the 2 class methods that I think are redundant: `.build` and `.clean_env`, as they can just happen in `RequestInterface#initialize`.

I also don't think this is a breaking change as `Event#rack_env=` should be the API for users to attaching request data to an event.


st0012:
@sl0thentr0py need to merge this for other work. but feel free to add comments and I'll address them in other PRs.

sl0thentr0py:
lgtm :)

PR #1678, opened 2022-01-07
Goodbye faraday 👋 

### LINKED ISSUE #1663: Faraday 2.0 requires locking to ~> 1.0 or switching to faraday-net_http
### Issue Description

Faraday 2.0 was released with major changes: https://github.com/lostisland/faraday/blob/main/UPGRADING.md#thats-great-what-should-i-change-in-my-code-immediately-after-upgrading

I'd suggest switching to `faraday-net_http` and setting `Faraday.default_adapter = :net_http` since this will remove a number of dependencies. Alternatively, it seems you could lock to ~> 1.0, or consider using net_http directly instead of relying on Faraday to sidestep this issue entirely. 

I'm not sure what's best... locking Faraday to ~> 1.0 might be easiest in the short term, but may cause issues later as users may want to upgrade to Faraday ~> 2.0 for other reasons. Perhaps `sentry-ruby` could detect which version is being used somehow, and switch based on that, but I don't believe you can customize the gem dependencies, so you may need to pick a "winner" between the two in the long term

I ran into a similar problem with gem dependencies here: https://github.com/geokit/geokit/pull/250

So perhaps dropping the dependency might be best, but that may result in users needing code changes in order to use Sentry, for example if they're setting `transport_configuration.http_adapter` etc, and of course even that change could significant work on your end. 

For now, I believe I can just manually lock Faraday to ~> 1.0 in my Gemfile, even though I'm only using Faraday because of Sentry. 

Anyway, this is a tricky problem -- I'm happy to help if I'm able. Thank you for your attention!

st0012:
1. Removed `faraday_builder` and `http_adapter` transport configs.
2. Rewrote `HTTPTransport` with pure `Net::HTTP` implementation. ([faraday-net_http](https://github.com/lostisland/faraday-net_http) provided pretty good guidance)
3. Removed `faraday` from dependencies.

## Todo

- ~~Test against different versions of `Net::HTTP` to see if the implementation can work from Ruby 2.4~3.1~~
- Update migration guide and documentation.

Closes #1663

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,132 @@
+## 5.0.0
+
+### Breaking Change - Goodbye `faraday` 👋 
+
+This version removes the dependency of [faraday](https://github.com/lostisland/faraday) and replaces related implementation with the `Net::HTTP` standard library. 
+
+(If you didn't notice it, the SDK previously used `faraday` as the HTTP client to send events to Sentry. You can find the implementation in [HTTPTransport](https://github.com/getsentry/sentry-ruby/blob/v4-9/sentry-ruby/lib/sentry/transport/http_transport.rb).)
+
+#### Why
+
+Since the old `sentry-raven` SDK, we've been using `faraday` as the SDKs' HTTP client for years. It's an amazing tool and saved us many work, allowing us to focus on SDK features. 
+
+But because many users also use `faraday` in their own apps and have their own version requirements, managing this dependency has become harder and harder over the past few years. Just to list a few related issues:
+
+- [#944](https://github.com/getsentry/sentry-ruby/issues/944)
+- [#1424](https://github.com/getsentry/sentry-ruby/issues/1424)
+- [#1524](https://github.com/getsentry/sentry-ruby/issues/1524)
+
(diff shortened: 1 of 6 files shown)


st0012:
@sl0thentr0py It seems to work 🎉

ojab:
btw, are other transports welcome? We're using [httpx](https://gitlab.com/honeyryderchuck/httpx/) for API clients, because it has nice API and HTTP/2 + persistent connections support. It wouldn't hurt to use it for interactions with sentry, curling DSN says that it's HTTP/2. The same question about adding breadcrumbs for it.

Sorry for offtopic, dunno where to ask.

st0012:
@ojab You can always use `config.transport.transport_class` to assign your custom transports. But I don't plan to introduce other built-in transports, especially not the ones that relies on 3rd-party libraries (so we don't need to go through a similar issue as https://github.com/getsentry/sentry-ruby/issues/1663 a few years later).
With that being said, I'd be happy to help if you want to introduce something like `sentry-httpx` to support both the transport and breadcrumb support 😄

ojab:
You mean as `sentry-rails` works currently, still in the same repo, just as a different gem?

st0012:
Oh no I mean you can write your own `sentry-httpx` gem (or whatever you want to call it). I can give you pointers on how to write breadcrumb loggers and transports if you need any.

PR #1681, opened 2022-01-10
feat(performance): Sync activerecord and net-http span names

### LINKED ISSUE #1674: Sync span operation names
Update all relevant span operation names to match the docs here
https://develop.sentry.dev/sdk/performance/span-operations/

This is necessary for users to use some of the newer Performance dashboard features like slow db and http ops.

image

sl0thentr0py:
The span prefixes are needed as documented [here](https://develop.sentry.dev/sdk/performance/span-operations/) for the discover slow ops dashboards to be populated successfully.

Note that this won't break any existing user queries since we currently don't allow filtering/searching by span ops.

Fixes #1674


AbhiPrasad (approved):
See the current accepted span operation prefixes: https://github.com/getsentry/sentry/blob/809b7fe54c6f06cc1e4c503cf83ded896472a011/src/sentry/projectoptions/defaults.py#L74

st0012 (commented):
Thanks for the fix! Can you also add a changelog entry for it?

st0012:
@sl0thentr0py sorry I missed one thing: perhaps we can also add `template.` prefix to ActionView's tracers?

sl0thentr0py:
Yea I added those then removed it because I wasn't able to make it clean. The db one works with prefixes, if we go with a prefix for views it'll look like `template.render_template.action_view`. 

Nicer would be `template.render.action_view`? In that case I'll go with a mapping hash instead of a prefix. What do you think?

sl0thentr0py:
nvm I just added the prefix, it's not the nicest looking with repetition but it's fine for now..

st0012:
Since it's to support a feature, I think the duplication is fine.

st0012 (approved):
Thanks for the feature! I'll put this to `5.1.0` release, which will be released soon after `5.0`

PR #1687, opened 2022-01-14
Don't initialize Event objects when they won't be sent

### LINKED ISSUE #1683: Sentry has RuntimeError when not enabled in current environment
### Issue Description

When running Capybara tests, and an app request raises an exception when my system hits its `ulimit` from using `SecureRandom`, Sentry also causes the same error when capturing an app exception and wrapping it, here: https://github.com/getsentry/sentry-ruby/blob/6d312f89d09db2e5a1f2cb5169047b4bff0eb839/sentry-ruby/lib/sentry/event.rb#L48

What happens then is my test failure trace is for the Sentry `capture_exception` trace, not the original app exception.

It was very confusing because `test` is not in my `config.enabled_environments` array.

### Reproduction Steps

- Set Sentry `config.enabled_environments` to an array of some environment names
- Make sure `SecureRandom` raises an error – I don't know how to cause this, maybe it can be mocked? Or the underlying system call can be mocked somehow? I can 75% reliably reproduce this on my current system (on which `ulimit -n` retunrs `256`) by running 15 feature tests that contain ~20 app requests each.
  - or make something in Sentry's code raise an error, and make your rack app raise any error, so this line captures an exception: https://github.com/getsentry/sentry-ruby/blob/6d312f89d09db2e5a1f2cb5169047b4bff0eb839/sentry-ruby/lib/sentry/rack/capture_exceptions.rb#L16
- Run your app in an environment that's not in `config.enabled_environments`, e.g. in a Rails app set `config.enabled_environments = ['production']` and run the app in `RAILS_ENV=test`
- Make an app request to hit the app code that will raise an error
- Check where app request exceptions are logged
- See that the trace is for the error raised in Sentry, not the app error

### Expected Behavior

Sentry code does not try to capture exceptions for an app who's environment is not in `config.enabled_environments`.

### Actual Behavior

Sentry runs `capture_exception` and tries to create an event for it, and raises an error of its own, replacing the app exception trace with its own.

### Ruby Version

2.6.9

### SDK Version

4.7.3

Rails 6.0.4.1

### Sentry Config

```rb
Sentry.init do |config|
  config.dsn = Rails.configuration.sentry.fetch(:sentry_dsn)
  config.enabled_environments = %w[production staging]
end
```

st0012:
Fixes #1683

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Add workaround for ConnectionStub's missing interface [#1686](https://github.com/getsentry/sentry-ruby/pull/1686)
   - Fixes [#1685](https://github.com/getsentry/sentry-ruby/issues/1685)
+- Don't initialize Event objects when they won't be sent [#1687](https://github.com/getsentry/sentry-ruby/pull/1687)
+  - Fixes [#1683](https://github.com/getsentry/sentry-ruby/issues/1683)
 
 ## 4.9.0
 
--- a/sentry-ruby/lib/sentry/client.rb
+++ b/sentry-ruby/lib/sentry/client.rb
@@ -76,8 +76,9 @@ module Sentry
     # @param hint [Hash] the hint data that'll be passed to `before_send` callback and the scope's event processors.
     # @return [Event, nil]
     def event_from_exception(exception, hint = {})
+      return unless @configuration.sending_allowed? && @configuration.exception_class_allowed?(exception)
+
       integration_meta = Sentry.integrations[hint[:integration]]
-      return unless @configuration.exception_class_allowed?(exception)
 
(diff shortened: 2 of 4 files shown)


sl0thentr0py (approved):
hmm ideally this check should be done just once in a centralized place, but I can't see a good place for it without changing too many things, so it's fine for now :)

st0012:
@sl0thentr0py Yeah I agree with you. I think it depends on whether we expect users to initialize events themselves and call `capture_event` manually.

If we do, then I think the current approach (2 levels of checks) is necessary.
If we don't, we need to prohibit such usages, which should have its own PR.

sl0thentr0py:
`Event.new` should be considered as internal api, users shouldn't really use it directly. Only methods on hub or those exposed directly on the top-level `Sentry` module are our 'contract'.

PR #1688, opened 2022-01-14
Support serializing ActiveRecord job arguments in global id form

### LINKED ISSUE #1643: Serialize ActiveRecord arguments to ActiveJob instances as GlobalIDs
**Describe the idea**

Ideally, this:

image

would instead show something like:

```
[
  {
    user : "gid://app/User/1"
  }
]
```

Currently it renders the `.to_s` or `.inspect` of the object which contains no useful information for diagnosing the issue.

**Possible implementation**

I am not sure how to do this cleanly. Obviously overriding [`Sentry::Event#to_hash`](https://github.com/getsentry/sentry-ruby/blob/7848d59450ae6bcb8dea6c141abdcdfe7b218567/sentry-ruby/lib/sentry/event.rb#L111-L119) could work but seems a bit iffy for projects without GlobalID.

Perhaps, it should happen in https://github.com/getsentry/sentry-ruby/blob/9303b9ade03e98096feb67784f8df1c129bdd496/sentry-rails/lib/sentry/rails/active_job.rb#L55-L64

st0012:
Thanks for @bjeanes's proposal and on the idea and implementation in #1643 

Closes #1643

### DIFF
--- a/sentry-rails/lib/sentry/rails/active_job.rb
+++ b/sentry-rails/lib/sentry/rails/active_job.rb
@@ -37,7 +37,7 @@ module Sentry
           extra: sentry_context,
           tags: {
             job_id: job_id,
-            provider_job_id:provider_job_id
+            provider_job_id: provider_job_id
           }
         )
         raise e
@@ -57,12 +57,25 @@ module Sentry
       def sentry_context
         {
           active_job: self.class.name,
-          arguments: arguments,
+          arguments: sentry_serialize_arguments(arguments),
           scheduled_at: scheduled_at,
           job_id: job_id,
           provider_job_id: provider_job_id,
           locale: locale
         }
+      end
+
+      def sentry_serialize_arguments(argument)
+        case argument
+        when Hash
+          argument.transform_values { |v| sentry_serialize_arguments(v) }
+        when Array, Enumerable
+          argument.map { |v| sentry_serialize_arguments(v) }
+        when ->(v) { v.respond_to?(:to_global_id) }
+          argument.to_global_id.to_s rescue argument
+        else
+          argument
+        end
       end
     end
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Features
+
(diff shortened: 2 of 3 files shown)


sentry-rails/lib/sentry/rails/active_job.rb
```diff
@@ -57,13 +57,36 @@ def already_supported_by_sentry_integration?
...
+        when Array, Enumerable
+          argument.map { |v| sentry_serialize_argument(v) }
+        when ->(v) { v.respond_to?(:to_global_id) }
+          argument.to_global_id.to_s
+        else
+          argument
+        end
+      rescue StandardError
+        argument
+      end
```
lewispb, line 89: Presuming we have Active Support available here in sentry-rails, we could perhaps simplify the implementation with [#deep_transform_values](https://edgeapi.rubyonrails.org/classes/Hash.html#method-i-deep_transform_values)

bjeanes: Godo point. I think this code is roughly based on my initial issue wherein I didn't want to assume ActiveSuport, but given this is would be in the `-rails` gem, seems fine to assume it's there.

st0012: Well I did simplified the implementation to:

```rb
      def sentry_serialize_arguments(argument)
        case argument
        when Hash
          argument.deep_transform_values { |v| sentry_serialize_arguments(v) }
        when Array, Enumerable
          argument.map { |v| sentry_serialize_arguments(v) }
        when ->(v) { v.respond_to?(:to_global_id) }
          argument.to_global_id.to_s
        else
          argument
        end
      rescue StandardError
        argument
      end
```

But it turns out `deep_transform_values` is added after Rails 6.0. And we're still supporting Rails 5.*, so....

lewispb: Ah ok 😅

st0012: @lewispb fyi `transform_values` also works 😄

st0012: @lewispb fyi `transform_values` also works 😄

PR #1692, opened 2022-01-17
Directly execute ActionCable's action if the SDK is not initialized

### LINKED ISSUE #1691: Doesn't execute blocks of capture() in tests: doesn't call subscribed from an action cable channel
### Issue Description

I'm trying to reproduce a bug with my action cable channel in my tests (rspec 3) (rails 7) and couldn't get the `subscribed` method to execute

### Reproduction Steps

have a channel:

```ruby
class MessagesChannel < ApplicationCable::Channel
  def subscribed
    raise NotImplementedError
  end
end
```

Test it:

```ruby
require "rails_helper"

RSpec.describe MessagesChannel do
  before do
    stub_connection
  end

  it "rejects subscriptions without a current user" do
    expect { subscribe }.to raise_error(NotImplementedError)
  end
end
```

### Expected Behavior

executes the subscribed block

### Actual Behavior

doesn't execute the subscribed block

### Ruby Version

ruby 3.1.0p0 (2021-12-25 revision fb4df44d16) [x86_64-darwin21]

### SDK Version

sentry-ruby-core (4.9.1)

sentry-rails (4.9.1)

### Sentry Config

_No response_

st0012:
Fixes #1691

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+### Bug Fixes
+
+- Directly execute ActionCable's action if the SDK is not initialized [#1692](https://github.com/getsentry/sentry-ruby/pull/1692)
+  - Fixes [#1691](https://github.com/getsentry/sentry-ruby/issues/1691)
+
 ## 4.9.1
 
 ### Bug Fixes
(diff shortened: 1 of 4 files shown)


sl0thentr0py (approved):

st0012:
@renchap Thanks for the feedback. I was waiting for a day to see if there're other issues to go along with the release. But it looks like there's not! I'll make the release shortly.

Issue #1695, opened 2022-01-23
Can't send transactions with SDK 5.0.0, frozen OpenSSL::SSL::SSLContext error


yorch:
### Issue Description

I have a Rails 6.1 app that I'm developing, where I have Sentry SDK integrated for a few weeks without problems. After upgrading SDK to `v5.0.0`, I've started seen `Transaction sending failed: can't modify frozen OpenSSL::SSL::SSLContext` message in the Rails logs. See below for a complete error message.

Gemfile:
```
gem 'sentry-ruby'
gem 'sentry-rails'
```

Logs:
```
Sending envelope [transaction] 7c1e1d08ccc076c8b85138ea2bcd5cfa to Sentry
Transaction sending failed: can't modify frozen OpenSSL::SSL::SSLContext: #<OpenSSL::SSL::SSLContext:0x0000000111d593c0 @verify_mode=1, @verify_hostname=true, @max_proto_version=nil, @min_proto_version=769, @cert_store=#<OpenSSL::X509::Store:0x000000010f7938c0 @verify_callback=nil, @error=nil, @error_string=nil, @chain=nil, @time=nil>, @session_new_cb=#<Proc:0x0000000111e83c00 /Users/user/.rvm/rubies/ruby-3.1.0/lib/ruby/3.1.0/net/http.rb:1038>>
Unreported Transaction: CustomerController#new
exception happened in background worker: can't modify frozen OpenSSL::SSL::SSLContext: #<OpenSSL::SSL::SSLContext:0x0000000111d593c0 @verify_mode=1, @verify_hostname=true, @max_proto_version=nil, @min_proto_version=769, @cert_store=#<OpenSSL::X509::Store:0x000000010f7938c0 @verify_callback=nil, @error=nil, @error_string=nil, @chain=nil, @time=nil>, @session_new_cb=#<Proc:0x0000000111e83c00 /Users/user/.rvm/rubies/ruby-3.1.0/lib/ruby/3.1.0/net/http.rb:1038>>
```

### Reproduction Steps

Upgraded SDK from `4.9.1` to `5.0.0` on a Rails 6.1 app using Ruby 3.1.0, and started seeing the error message described.

### Expected Behavior

Transactions being sent to Sentry as in previous SDK versions.

### Actual Behavior

Getting error messages while Sentry trying to send transactions.

### Ruby Version

3.1.0

### SDK Version

5.0.0

### Integration and Its Version

Rails

st0012:
[...]

PR #1696, opened 2022-01-23
Don't reuse Net::HTTP objects in `HTTPTransport`

st0012:
When investigating #1695, I found that `Net::HTTP` object is not supposed to be used as `Faraday::Connection` that can be held and shared. Because it'll cause error `HTTP session already opened` when multiple threads try to send events with the same object.

I also assume #1695 is caused by a race condition between threads that attempt to update the same `OpenSSL::SSL::SSLSocket` instance's session [around here](https://github.com/ruby/net-http/blob/v0.2.0/lib/net/http.rb#L1030-L1039)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 5.0.1
+
+- Don't reuse Net::HTTP objects in `HTTPTransport` [#1696](https://github.com/getsentry/sentry-ruby/pull/1696)
+
 ## 5.0.0
 
 ### Breaking Change - Goodbye `faraday` 👋 
--- a/sentry-ruby/lib/sentry/transport/http_transport.rb
+++ b/sentry-ruby/lib/sentry/transport/http_transport.rb
@@ -14,12 +14,11 @@ module Sentry
     RATE_LIMIT_HEADER = "x-sentry-rate-limits"
     USER_AGENT = "sentry-ruby/#{Sentry::VERSION}"
 
-    attr_reader :conn
-
     def initialize(*args)
       super
-      @conn = set_conn
       @endpoint = @dsn.envelope_endpoint
+
+      log_debug("Sentry HTTP Transport will connect to #{@dsn.server}")
     end
 
     def send_data(data)
@@ -127,10 +126,8 @@ module Sentry
       @transport_configuration.encoding == GZIP_ENCODING && data.bytesize >= GZIP_THRESHOLD
     end
 
-    def set_conn
+    def conn
       server = URI(@dsn.server)
-
-      log_debug("Sentry HTTP Transport connecting to #{server}")
 
       use_ssl = server.scheme == "https"
       port = use_ssl ? 443 : 80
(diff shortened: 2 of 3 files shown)


st0012:
@sl0thentr0py this is kind of an urgent issue so I'll merge and prepare the release first. And hopefully this can be released ASAP.

sl0thentr0py:
hmm ok, let's release this because it's a bug but some comments for later.

How does this affect performance/overhead? I assume reusing the connection if possible does have some efficiency over a fresh connection every time, but I'm not sure about the exact details.

Another discrepancy with python also appears here.
In ruby, we have the `BackgroundWorker` instance on the top level `Sentry` module while in python, we have the worker on the [transport itself](https://github.com/getsentry/sentry-python/blob/6d89f4f50b5e9f8a8dcef4efdd73f99e050ea410/sentry_sdk/transport.py#L145). If possible, I would like to resolve these inconsistencies in architecture longer term so we have the same high level logic across SDKs.

st0012:
@sl0thentr0py sorry that I should have mentioned this clearer: `faraday` also initializes new instances for every request. So it's the removal of `faraday` that changed the behavior. This fix just reverts it.

Here's the stacktrace that should help you understand how that's done:

```
/Users/st0012/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/faraday-1.9.3/lib/faraday/adapter.rb:47:in `connection'
/Users/st0012/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/faraday-net_http-1.0.1/lib/faraday/adapter/net_http.rb:66:in `call'
/Users/st0012/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/faraday-1.9.3/lib/faraday/middleware.rb:18:in `call'
/Users/st0012/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/faraday-1.9.3/lib/faraday/rack_builder.rb:154:in `build_response'
/Users/st0012/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/faraday-1.9.3/lib/faraday/connection.rb:516:in `run_request'
/Users/st0012/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/faraday-1.9.3/lib/faraday/connection.rb:281:in `post'
/Users/st0012/projects/sentry-ruby/sentry-ruby/lib/sentry/transport/http_transport.rb:33:in `send_data'
```

>  In ruby, we have the BackgroundWorker instance on the top level Sentry module while in python, we have the worker on the transport itself. If possible, I would like to resolve these inconsistencies in architecture longer term so we have the same high level logic across SDKs.

When I implemented it, there wasn't any guidance around how/where to build the worker and that's why it's done this way. I'm not against moving it but I won't give it a high priority either. Can you open an issue for that then we can move the discussion there? thx

sl0thentr0py:
> faraday also initializes new instances for every request. 

ah ok thx, ignore then! 

>  there wasn't any guidance around 

yes absolutely! And it works absolutely fine. :)
Moving further discussion outside to not pollute this issue more.

Issue #1693, opened 2022-01-20
Redis tracing


lewispb:
**Describe the idea**

Similar to the existing Sentry Performance support for Active Record database tracing, we've found it useful to view the time taken to execute Redis commands.

I've created a simple custom Redis instrumenter and tracing subscriber. With this enabled, we can inspect traces including Redis commands:

<img width="1432" alt="sentry-performance-with-redis" src="https://user-images.githubusercontent.com/1773614/150431894-e0277ab3-0f92-488a-a94b-0e8422eb3406.png">

**Why do you think it's beneficial to most of the users**

Redis is a popular and widely-used database.

**Possible implementation**

We would add the following to the "sentry-rails" gem:

- A custom instrumenter, inspired by the [Redis Rails Instrumentation gem](https://github.com/lautis/redis-rails-instrumentation)
- A subscriber to the newly generated Active Support Notifications

I considered a new gem "sentry-redis", but it appears that the Performance product with tracing is only available for Rails apps. 

### Instrumentation example

```rb
class Redis
  module Instrumentation
    def logging(commands, &block)
      ActiveSupport::Notifications.instrument("command.redis", commands: commands, server: "#{host}:#{port}/#{db}") do
        return super(commands, &block)
      end
    end
  end
end

Redis::Client.prepend(Redis::Instrumentation)
```

### Subscriber example

The first command and key is instrumented, but not the value.

```rb
require "sentry/rails/tracing/abstract_subscriber"

module Sentry
  module Rails
    module Tracing
      class RedisSubscriber < AbstractSubscriber
        def self.subscribe!
          subscribe_to_event([ "command.redis" ]) do |event_name, duration, payload|
            description = payload.dig(:commands, 0).take(2).join(" ")

            record_on_current_span(
              op: event_name,
              start_timestamp: payload[START_TIMESTAMP_NAME],
              description: description,
              duration: duration
            ) do |span|
              span.set_data(:redis_server, payload[:server])
            end
          end
        end
      end
    end
  end
end
```

I'm happy to implement this if the approach is acceptable. Thanks!

st0012:
Thanks for the proposal, I think it'd be a great feature to add!

> I considered a new gem "sentry-redis", but it appears that the Performance product with tracing is only available for Rails apps.

[...]

PR #1697, opened 2022-01-23
Redis support for sentry-ruby

lewispb:
## Description

Implements https://github.com/getsentry/sentry-ruby/issues/1693

Patches a Redis client to capture breadcrumb and span data when calls are made to Redis.

#### Todo
- [x] Test scenario where Redis client isn't available
- [x] Test more complex Redis commands
- [x] Tests for breadcrumb logger
- [x] Manual testing
- [x] Changelog 

#### Outstanding questions
- Some folks might find it useful to also capture the values of the Redis commands. That would likely contain PII so we'd likely want to respect the `config.send_default_pii` setting. Is capturing these values something we need right away, or can it wait for a future PR?
- Should we add configuration to opt-out of the Redis tracing?
- What should the OP_NAME be? Right now I've set it to "[db.redis.command](https://github.com/getsentry/sentry-ruby/issues/1674#issuecomment-1007558239)", but could be "redis.command" or simply "redis"?


st0012:
@lewispb thanks for the work! I think you've done a fantastic job on this 👍 
but it may take another week or 2 for me to start merging changes for the `5.1.0` release. so I will give detailed review around then.

lewispb:
No problem @st0012, thanks for looking!

sentry-ruby/spec/sentry/breadcrumb/redis_logger_spec.rb
```diff
@@ -0,0 +1,81 @@
+require "spec_helper"
+require "fakeredis"
```
st0012, line 2: I think it's ok to put this in `spec_helper` and before `require "sentry-ruby"`. Then we don't need to force the `redis.rb` reload anymore. Wdyt?

st0012 (commented):
I think the implementation looks good and `logging` is probably the best place to patch for this 👍 

I left 2 comments but generally it's ready to merge. After merging I'll play a bit with it and perhaps add some screenshot to the changelog & improve some tests around SDK initialization check.

sentry-ruby/lib/sentry/redis.rb
```diff
@@ -0,0 +1,88 @@
...
+
+    def commands_description
+      parsed_commands.map do |statement|
+        statement.values.join(" ").strip
+      end.join(", ")
+    end
+
+    def parsed_commands
+      commands.map do |statement|
+        command, key, *_values = statement
```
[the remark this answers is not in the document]
st0012: Sorry I missed the questions in the description, please ignore this.

st0012:
> Some folks might find it useful to also capture the values of the Redis commands. That would likely contain PII so we'd likely want to respect the config.send_default_pii setting. Is capturing these values something we need right away, or can it wait for a future PR?

I can implement that in a separate PR and ask for your opinion, wdyt?

> Should we add configuration to opt-out of the Redis tracing?

I haven't received any request to opt-out specific tracing integration. So it's probably not needed now.
Even if we do receive that, it'll be a broader discussion around the entire SDK design. So it's out of this PR's scope.

> What should the OP_NAME be? Right now I've set it to "db.redis.command", but could be "redis.command" or simply "redis"?

I think `db.redis.command` matches the [recommended span names](https://develop.sentry.dev/sdk/performance/span-operations/). @sl0thentr0py can you confirm this is the case?

sl0thentr0py:
yes that span status is correct, only the `db.` prefix matters currently.

sl0thentr0py:
thanks for the contribution @lewispb! 
There is a high-level architecture discussion that I wanted to get into at some point and this new redis integration is a good opportunity to do that.

In the ruby SDK currently, we have separate gems encapsulating specific integrations (`rake` is a one-off here that has it's own `:skip_rake_integration` config param). In other sentry SDKs, we generally use a `config.integrations = [:foo, :bar]` list to enable/disable specific integrations. Ruby is of course different here because of how `Gemfile`s work and how autoloading works in rails, so this different architecture is certainly justified.

Now with redis, we're adding it to the main `sentry-ruby` gem, and so it lies in neither broader architecture. I don't want to start too many large changes now, but it would be nice if we come up with some kind of convention and all agree going forward on how to add new integrations to the ruby SDK.

sentry-ruby/spec/sentry/breadcrumb/redis_logger_spec.rb
lewispb: Good idea!

sentry-ruby/lib/sentry/redis.rb
```diff
@@ -0,0 +1,88 @@
...
+    end
+
+    private
+
+    attr_reader :commands, :host, :port, :db
+
+    def record_span
+      return yield unless (transaction = Sentry.get_current_scope.get_transaction) && transaction.sampled
+
+      transaction.start_child(op: OP_NAME, start_timestamp: Sentry.utc_now.to_f).then do |sentry_span|
```
lewispb, line 30: I use then here, which is Ruby 2.6. Or we could use yield_self (Ruby 2.5).

Do we need to support Ruby 2.4?

st0012:
> There is a high-level architecture discussion that I wanted to get into at some point and this new redis integration is a good opportunity to do that.

I think there are different categories in terms of "integrations":

1. Libraries that are opted-in very intentionally, like `Rails`, `Sidekiq`...etc.
2. Default-like libraries (90% of the apps have them), like `Rake` or `Rack`.
3. Low-level components like `net-http` and `redis`. They usually power the libraries like 1) and other gems, but not used directly.

For 1), the current approach is to ask users to install the related `sentry-*` gem. I think it's makes sense as they are aware of the library/framework they chose and should be motivated to install the related Sentry integration.

For 2), the current approach is to have their code in `sentry-ruby` but the integration is handled case by case:

- Because Rack middleware requires explicit API to inject a middleware, so we let user do that themselves.
- On the other hand, Rake integration is more like a patch-based approach. So we apply that by default.
    - However, some users are not happy about that so we provided options to opt-out

For 3), the current approach is to detect if those components exist in the app, and opt-in for the user.

Let me explain a bit on my decision on 3):

Most users don't spend time configuring their Sentry SDK. They copy and paste the configuration from documentation or setup wizard, check if errors go through, and then forget about it. I have 2 examples for this assumption:

1. We mention `config.traces_sampler` and `config.traces_sample_rate`, and tell users to pick one of them, which looks like this:
```
  config.traces_sample_rate = 0.5
  # or
    0.5
  end
```

And guess what happened, [many users](https://github.com/search?l=Ruby&q=config.traces_sample_rate+config.traces_sampler&type=Code) leave both of them uncommented 😂 

2. There were many problems on the ActiveSupport breadcrumb logger (not the logger code itself but they way it collects all the data without proper filtering). While the logger was added years ago to `sentry-raven`, we didn't receive much issues report until the release of `sentry-rails` a year ago. Why? Because most users didn't even know it's a thing and never used it before. My the other project is one of those.

So my concern is that most users may not aware of those lower-level components and thus won't intentionally make the effort to opt-in the integration. That's why I didn't want to make `sentry-redis` or `sentry-net_http`, because I know many users won't even know them exist even if they do use those libraries somewhere. And that'll be a shame because they also provide valuable information. You can also argue that it'll give users an impression that `sentry-ruby` provides much less feature when compare to other services' SDKs, where those integrations are opted-in by default.

sentry-ruby/lib/sentry/redis.rb
st0012: @lewispb Yeah we still haven't dropped Ruby 2.4 yet. So it should be supported.
Speaking of this, @sl0thentr0py do you think we can get some statistics about our users' Ruby version breakdown?

lewispb:
> I can implement that in a separate PR and ask for your opinion, wdyt?

Sounds great, thanks.

> I haven't received any request to opt-out specific tracing integration. So it's probably not needed now.

Cool, OK!

sl0thentr0py:
Completely agree on the zero-config front, that is also very much our design philosophy as well.
It's just that these 3 implicit categories add some cognitive overhead for the end-user too and make docs potentially confusing. Other SDKs just have them all in the same 'integration' category and a lot of them are also enabled by default so most users don't have to care.

Either way, I don't wanna bikeshed on this too much, but it's instructive that we had this conversation. Let's just go with the PR in its current form. :+1: We can agree on a simple rule like 'large' frameworks go in separate gems and everything else goes in `sentry-ruby`. If we add more optional things in the future, we will have to figure out how we let users switch them on/off.

st0012 (approved):
👍 Thanks for this great PR!

lewispb:
Sure, no problem! I could also take a look at it if you like, although probably not till next week?

PR #1701, opened 2022-01-26
Capture transaction tags

karlentwistle:
## Description

With the following code:

```ruby
transaction = Sentry.start_transaction(name: 'example transaction', description: "example transaction")
transaction.set_tag('host', 'www.example.org')
transaction.finish
```

I was expecting Sentry to create a clickable tag within the Tag Details section here:

Screenshot 2022-01-26 at 22 08 13

However - this doesn't currently work because the transaction's tags aren't shared here: https://github.com/getsentry/sentry-ruby/blob/16a13052b562b6a6a3b1f9747e69141a6eaed9f3/sentry-ruby/lib/sentry/client.rb#L105

With this change applied the tags from the transaction are captured and it creates a clickable tag within Event Details -> Tag Details

Screenshot 2022-01-26 at 22 13 34

I wondered if there might be a specific reason transaction tags aren't currently captured or might it be possible to introduce this change.

Please let me know your thoughts - thanks very much!

### DIFF
--- a/sentry-ruby/lib/sentry/client.rb
+++ b/sentry-ruby/lib/sentry/client.rb
@@ -108,6 +108,7 @@ module Sentry
         event.contexts.merge!(trace: transaction.get_trace_context)
         event.timestamp = transaction.timestamp
         event.start_timestamp = transaction.start_timestamp
+        event.tags = transaction.tags
 
         finished_spans = transaction.span_recorder.spans.select { |span| span.timestamp && span != transaction }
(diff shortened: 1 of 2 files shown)


st0012 (approved):
Yes it's something I missed. Sorry about that and thanks for the PR 🙌

st0012:
@sl0thentr0py FYI. And here is [sentry-python's implementation](https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/tracing.py#L596).

PR #1705, opened 2022-01-30
Register Sentry's ErrorSubscriber for Rails 7.0+ apps

### LINKED ISSUE #1630: Support the new Rails standard error reporting interface
Rails 7.0 will add a standard interface for reporting errors to error tracking services like Sentry (rails/rails#43625).

```ruby
# Report any error raised and continue execution:
Rails.error.handle do
  # ...
end

# Report and re-raise any error:
Rails.error.record do
  # ...
end

# Report an error directly:
Rails.error.report(error)
```

Additionally, Rails’s internal execution wrappers have been updated to report errors via `Rails.error.report`.

The Sentry SDK can opt in to the new interface via [`Rails.error.subscribe(subscriber)`](https://github.com/rails/rails/blob/89b1cc4818e3cb08ed403441001b8853502a637c/activesupport/lib/active_support/error_reporter.rb#L63-L67), where `subscriber` is an object implementing a `report` method. `report` takes a positional exception argument and `handled:` and `context:` keyword arguments.

```
report(Exception, handled: Boolean, context: Hash)
```

Supporting the new interface will allow Sentry users to use the standard `Rails.error.{handle,record,report}` methods. And in the future, when the standard interface is more widely deployed, it’ll make it easier for Rails users to try and adopt Sentry.

st0012:
This implements the integration proposed in #1630.

After Rails 7.0, you'll be able to use the standardized `Rails.error.record` or `Rails.error.handle` interfaces to report exceptions to Sentry. Please check [ActiveSupport::ErrorReporter](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/error_reporter.rb) for more information.

### Gotcha & Workaround

If an operation is wrapped inside `app.executor.wrap`, any exception raised will be captured by Rails' error reporter,
even if it has been reported by either the SDK integration or user. So in some cases (e.g. `ActionCable`'s channel actions), the same error could be reported twice.

I worked around this by marking if an exception has been captured by the SDK. And if it has, we don't capture it with the `ErrorSubscriber`.

(Dropping the original integration isn't an option because the contextual data may not be accessible from the error reporter)

Closes #1630

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Features
 
 - Support serializing ActiveRecord job arguments in global id form [#1688](https://github.com/getsentry/sentry-ruby/pull/1688)
+- Register Sentry's ErrorSubscriber for Rails 7.0+ apps [#1705](https://github.com/getsentry/sentry-ruby/pull/1705)
 
 ## 5.0.2
 
--- a/sentry-rails/lib/sentry/rails/error_subscriber.rb
+++ b/sentry-rails/lib/sentry/rails/error_subscriber.rb
@@ -1,0 +1,11 @@
+module Sentry
+  module Rails
+    # This is not a user-facing class. You should use it with Rails 7.0's error reporter feature and its interfaces.
+    # See https://github.com/rails/rails/blob/main/activesupport/lib/active_support/error_reporter.rb for more information.
+    class ErrorSubscriber
+      def report(error, handled:, severity:, context:)
+        Sentry::Rails.capture_exception(error, level: severity, contexts: { "rails.error" => context }, tags: { handled: handled })
+      end
+    end
+  end
+end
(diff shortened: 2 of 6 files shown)


st0012:
@georgeclaghorn I'd like to hear your opinion on the gotcha and the workaround I took 🙂

georgeclaghorn:
I don't feel familiar enough with this gem to give a thorough review but this all seems reasonable to me. :+1:

casperisfine:
> (Dropping the original integration isn't an option because the contextual data may not be accessible from the error reporter)

Could you expand on that? I'm not sure I follow.

ariccio:
Interesting. I *think* this means I can turn on [`config.active_record.db_warnings_action`](https://edgeguides.rubyonrails.org/configuring.html#config-active-record-db-warnings-action), setting it to `:report`, and it will automatically send them through to sentry? Awesome! I'll have to try it!

PR #1708, opened 2022-02-01
Isolate ActiveJob error handling methods

mbobin:

**Other Notes**

## Description
Describe your changes:

This pull request refactors the `ActiveJobExtensions` module to prevent its methods from being leaked into the `ActiveJob` object. The reason behind this change is that I'm trying to run `sentry-raven` and `sentry-rails` side by side while doing the upgrade and both gems define these methods.

### DIFF
--- a/sentry-rails/lib/sentry/rails/active_job.rb
+++ b/sentry-rails/lib/sentry/rails/active_job.rb
@@ -2,79 +2,89 @@ module Sentry
   module Rails
     module ActiveJobExtensions
       def perform_now
-        if !Sentry.initialized? || already_supported_by_sentry_integration?
-          super
-        else
-          Sentry.with_scope do |scope|
-            capture_and_reraise_with_sentry(scope) do
-              super
+        Adapter.new(self).perform { super }
+      end
+
+      class Adapter
+        attr_reader :job
+
+        def initialize(job)
+          @job = job
+        end
+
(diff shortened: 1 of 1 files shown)


mbobin:
@st0012 could you have a look at this PR, please? 🙏

st0012:
@mbobin I'm neutral for this type of change but I don't think this one is justified, because compatibility with `sentry-raven` isn't something we support.

If it's a temporary need, I'd recommend you to have a fork of `sentry-raven` with the similar changes. That way, once you finished testing you can just drop it.

mbobin:
@st0012 I thought this benefits this gem more because it stops polluting ActiveJob's descendants with Sentry specific methods.  Let's take for instance the [changes from v4.8.2](https://my.diffend.io/gems/sentry-rails/4.8.1/4.8.2), this is where `capture_and_reraise_with_sentry`'s signature was changed from 3 to 2 arguments and this can be seen as a breaking change because they are exposed as public methods in the job classes and people might use them.

st0012:
@mbobin ok I get the point. I'll make a PR for it 👍

st0012:
@mbobin I've opened https://github.com/getsentry/sentry-ruby/pull/1725 for the same purpose. So I'm going to close this now. Thanks for proposing this change.

PR #1713, opened 2022-02-05
Avoid causing NoMethodError for Sentry.* calls when the SDK is not inited

### LINKED ISSUE #1706: Don't cause `NoMethodError` when calling `Sentry.*` with an uninitialized SDK
It should be either no-op or raising an exception, as the SDK should always be initialized (this is a requirement across all Sentry SDKs).

I prefer raising exceptions for such condition and have proposed it in a recent discord discussion as well. But we also have methods like `Sentry.capture_*` that current does nothing instead of raising an exception. So my concerns are:

- If I make methods like `Sentry.set_tags` raise exceptions, should I also change `Sentry.capture_*`'s behaviors.
- If the answer is `yes`, should we consider that as a breaking change?

As maintainers we can argue that it's a bug fix so it's not a breaking change (can go out at any patch). But to some (probably few) users, it'll be a change that break their app without actually providing visible benefits to them (should be in a major release).

So I think the first step would be to stop causing the `NoMethodError`. And then we can discuss when to start raising the exception.

st0012:
Closes #1706

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,11 @@ the format "MULTI, SET mykey, INCR counter, EXEC".
 - Support serializing ActiveRecord job arguments in global id form [#1688](https://github.com/getsentry/sentry-ruby/pull/1688)
 - Register Sentry's ErrorSubscriber for Rails 7.0+ apps [#1705](https://github.com/getsentry/sentry-ruby/pull/1705)
 
+### Bug Fixes
+
+- Avoid causing NoMethodError for Sentry.* calls when the SDK is not inited [#1713](https://github.com/getsentry/sentry-ruby/pull/1713)
+  - Fixes [#1706](https://github.com/getsentry/sentry-ruby/issues/1706)
+
 ## 5.0.2
 
 - Respect port info provided in user's DSN [#1702](https://github.com/getsentry/sentry-ruby/pull/1702)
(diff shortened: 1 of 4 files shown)


sl0thentr0py (approved):
thx :)

lewispb:
This PR looks good as-is but I was wondering if it was possible to avoid the explicit `return unless initialized?` calls everywhere.

One idea I had to do that would be to append something like this to `sentry-ruby.rb` (not tested):

```rb
unless Sentry.initialized?
  klass_methods = Sentry.methods(false)
  Sentry.singleton_class.prepend(Module.new do
    klass_methods.each do |method_name|
      define_method(method_name) do
        nil
      end
    end
  end)
end
```

That way it'd automatically cover future changes too.

st0012:
@lewispb Yeah I've thought about that but I didn't think it's a good idea because:

1. The interfaces are pretty static. We haven't been introduce any new method for months. So the ease of adding new APIs is not the most important thing here.
2. Whichever meta-programming route we take, it won't be very expressive about its intention.
3. It's more common that users/maintainers debug or investigate individual API methods than examining `sentry-ruby.rb` altogether. So I think explicitly stating `return unless initialized?` in every method makes them more understandable in most cases.

lewispb:
@st0012 those are all good points 👍 thanks for explaining your thinking here.

st0012:
@lewispb np, thanks for the advice too. I sincerely appreciate that you check my code and provide suggestions from time to time!

Issue #1707, opened 2022-02-01
Reorganize `sentry-rails`' test apps


st0012:
`sentry-rails` supports a wide range of Rails versions (from 5.0 to 7.0). This means that it's test setup is complex and [full of compatibility workarounds](https://github.com/getsentry/sentry-ruby/blob/master/sentry-rails/spec/support/test_rails_app/app.rb#L155-L197). We should start thinking a more maintainable way to build test apps (perhaps  one app file for each Rails version).

st0012:
Closed via #1719

PR #1719, opened 2022-02-10
Implement #1707

st0012:
One tricky thing of maintaining Rails support across version 5.0~7.0 is to have a test app that can work with all of them. Right now it's achieved by putting many conditions (for dependencies, application code, and cleanup logic) inside the app file. But I think in the long-term that's not maintainable. So what I try to do in this PR is to give each version its own test app, while maximizing reused logic.

I know there's always one concern about this approach: what if we need to add something to the test app? wouldn't we need to add them n times?

I thought about that and I think it rarely happen because:
- Most of the time we update the test app for new features support from `sentry-ruby` or supporting a new Rails version.
- All the recently added features only work after a certain version of Rails. For example, none of them would effect Rails 5.0 and 5.1 apps. So under this structure, only the related version files would be updated. `5-0.rb` would probably stay as it is until we drop Rails 5.0 support.
- Many cleanup code I added recently were for Rails 7.0 only. So they won't be applied across versions either.

### DIFF
--- a/sentry-rails/spec/support/test_rails_app/app.rb
+++ b/sentry-rails/spec/support/test_rails_app/app.rb
@@ -1,6 +1,12 @@
 ENV["RAILS_ENV"] = "test"
 
 require 'rails'
+
+if Rails.version.to_f < 5.2
+  require "support/test_rails_app/apps/5-0"
+  return
+end
+
 require "active_record"
 require "active_job/railtie"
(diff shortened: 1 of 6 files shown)


sl0thentr0py:
@st0012 this is great! the more explicit the better. I'm generally fine with duplication when necessary.

PR #1721, opened 2022-02-11
Refactor sentry-rails test app

st0012:
- I find the directory name `support` a bit odd and `dummy` is probably better.
- Some configs can actually be centralized.


sentry-rails/spec/fixtures/test_rails_app/app.rb
```diff
@@ -24,15 +25,15 @@ class TestApp < Rails::Application
 
 case Gem::Version.new(Rails.version)
 when -> (v) { v < v5_2 }
-  require "support/test_rails_app/apps/5-0"
+  require "fixtures/test_rails_app/apps/5-0"
```
lewispb, line 28: Fixtures reads to me as [Rails fixtures](https://api.rubyonrails.org/v7.0.2/classes/ActiveRecord/FixtureSet.html). How about using `/dummy`? I think that's the standard across Rails gems.

st0012: `dummy` is a better name, thanks 👍

PR #1723, opened 2022-02-11
Nested childs for Active Record SQL queries

pbernery:
Currently, when nesting children (using `start_child`), SQL queries, Redis commands, are not attached to the created span but are attached to the root span (or the transaction).

This makes the span graph looks like this:

not-nested

while I expect to have "db.sql.active_record" spans nested in parent spans.

I made some change in this PR that leads to the expected behaviour:

nested

but I am not sure than calling `set_span` is the right approach, it changes the "current" span globally. This works when there is one nested branch but if there are multiple, this may not work as expected.

The Redis spans have the same issue:
with-redis

And I guess the issue is everywhere `Sentry.get_current_scope.get_transaction` is used.

There was a PR https://github.com/getsentry/sentry-ruby/pull/1382/ that fixed nested custom child. It fixed things for children created manually, and proposes to use a child to create another child, but how could events handled by Sentry use those child?

### DIFF
--- a/sentry-rails/lib/sentry/rails/tracing/abstract_subscriber.rb
+++ b/sentry-rails/lib/sentry/rails/tracing/abstract_subscriber.rb
@@ -41,14 +41,18 @@ module Sentry
             return unless options[:start_timestamp]
 
             scope = Sentry.get_current_scope
-            transaction = scope.get_transaction
-            return unless transaction && transaction.sampled
+            current_span = scope.get_span
+            return unless current_span && current_span.sampled
 
-            span = transaction.start_child(**options)
+            span = current_span.start_child(**options)
+            Sentry.get_current_scope.set_span(span)
+
             # duration in ActiveSupport is computed in millisecond
             # so we need to covert it as second before calculating the timestamp
             span.set_timestamp(span.start_timestamp + duration / 1000)
             yield(span) if block_given?
+
+            Sentry.get_current_scope.set_span(current_span)
           end
         end
       end


sentry-rails/lib/sentry/rails/tracing/abstract_subscriber.rb
```diff
@@ -41,14 +41,18 @@ def record_on_current_span(duration:, **options)
             return unless options[:start_timestamp]
 
             scope = Sentry.get_current_scope
-            transaction = scope.get_transaction
-            return unless transaction && transaction.sampled
+            current_span = scope.get_span
+            return unless current_span && current_span.sampled
+
+            span = current_span.start_child(**options)
+            Sentry.get_current_scope.set_span(span)
```
st0012, line 48: I think this and the `get_current_scope` are unnecessary. Aren't they the same scope as the `scope` local?

sentry-rails/lib/sentry/rails/tracing/abstract_subscriber.rb
```diff
@@ -41,14 +41,18 @@ def record_on_current_span(duration:, **options)
             return unless options[:start_timestamp]
 
             scope = Sentry.get_current_scope
-            transaction = scope.get_transaction
-            return unless transaction && transaction.sampled
+            current_span = scope.get_span
+            return unless current_span && current_span.sampled
+
+            span = current_span.start_child(**options)
```
st0012, line 47: We can use `with_child_span(&block)` here:

https://github.com/getsentry/sentry-ruby/blob/341b66e5f3af6643cdb9458611fdacff8f60584d/sentry-ruby/lib/sentry/span.rb#L151-L166

st0012 (commented):
I generally agree with this idea. But I'm not sure if it could cause problems to some users?

In my other project, there could be many n+1 queries across different abstraction layers in problematic endpoints. In those cases, I'm worry that nesting will make the spans harder to view? (Of course, you can also argue that it's an UI design issue).

@sl0thentr0py Do you think the above would be an issue? Is there any recommendation on nesting spans in integrations?

pbernery:
Thanks for both your feedbacks.

So, with @st0012 suggestions applied, do you believe this PR could be merged as-is then, or should more work be done?

The changes have made works for SQL transactions, but not for Redis transactions, so I guess the equivalent should be applied on the Redis side. And maybe more services.

But we could advance step by step: nest SQL first, then later Redis, etc.

st0012:
@pbernery the proposal is good but we can't merge the current code and without tests.

st0012:
Let me know if you plan to finish the PR, or I can make another one to address the issue as well.

st0012:
I think the SDK should actually provide a top-level helper like `Sentry.with_child_span { |span| ... }` to do all the checks and heavy lifting work. So I will open another PR and make the change.

pbernery:
Hi Stan, sorry for the delayed response.
Alright, sounds good, and looking forward to this update.

st0012:
@pbernery I've added #1784 to address the issue. Can you give it a try when have time? thx
I also introduced the `Sentry.with_child_span` API in #1783, which I think you'd want to take a look to 🙂

pbernery:
Hey @st0012, just tested the Sentry.with_child_span and it looks like it's working great on our side.
sentry-redis may need the same change to nest calls the same way.

Thank you very much for this update :)

st0012:
@pbernery Thanks for catching that 👍 I've added another PR to address it.

PR #1724, opened 2022-02-11
Fix context merging #1722

### LINKED ISSUE #1722: set_context doesn't allow to overwrite values
### Issue Description

When calling `set_context` multiple times with the same context key and the same keys within the value, the value hashes will be merged into one another. But if the same keys are used within these hashes, the new values won't overwrite the old ones.

Example:
When iterating over objects and doing possibly problematic things, I'd like to recognize which object caused the error in the resulting sentry issue. But only the first value put into the context will be submitted. In my opinion (and the usecase above), it would be more intuitive to let the new values overwrite the old ones.

### Reproduction Steps

```ruby
object_ids = [1, 2, 3]

object_ids.each do |id|
  Sentry.configure_scope { |scope| scope.set_context(:objects, object_id: id) }
  Sentry.capture_message("cannot handle object") if id == 2
end
```

### Expected Behavior

The context `:objects` should be `{ object_id: 2 }`.

### Actual Behavior

The  context `:objects` is `{ object_id: 1 }`, always the first item in the list.

This is due to the merge direction:

https://github.com/getsentry/sentry-ruby/blob/c65088f988890573365a44526606020c7d18d738/sentry-ruby/lib/sentry/scope.rb#L173-L178

In line 176, the merge direction should be reversed:

```ruby
old.merge(new)
```

### Ruby Version

3.0.0

### SDK Version

5.1.0

Sentry::Rack::CaptureExceptions

### Sentry Config

```ruby
Sentry.init do |config|
  config.dsn = "..."
  config.environment = "..."
  config.release = "..."
end
```

thomasjachmann:
When using the same key in a context hash, the new value should
overwrite the old value. Before, the value having been set first would
always win.

fixes #1722

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Unreleased
+
+### Bug Fixes
+
+- Allow overwriting of context values [#1724](https://github.com/getsentry/sentry-ruby/pull/1724)
+  - Fixes [#1722](https://github.com/getsentry/sentry-ruby/issues/1722)
+
 ## 5.1.0
 
 ### Features
--- a/sentry-ruby/lib/sentry/scope.rb
+++ b/sentry-ruby/lib/sentry/scope.rb
@@ -173,7 +173,7 @@ module Sentry
     def set_contexts(contexts_hash)
       check_argument_type!(contexts_hash, Hash)
       @contexts.merge!(contexts_hash) do |key, old, new|
-        new.merge(old)
+        old.merge(new)
       end
     end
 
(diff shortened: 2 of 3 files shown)


st0012 (approved):
Yes, that's a bug. Thank you for reporting it and making the PR! 😄

rahul342:
Thank for you fixing! Curious, is there a timeline by which this goes live in new release?

st0012:
This should go out with version `5.1.1` next week 🙂

Issue #1718, opened 2022-02-10
Record redis command arguments in breadcrumbs and tracing spans.


st0012:
**Describe the idea**

Record redis command arguments in breadcrumbs and tracing spans.
The value should only be recorded if `send_default_pii == true`

**Why do you think it's beneficial to most of the users**

It provides more information to help users debug.

PR #1726, opened 2022-02-13
Log Redis command arguments if send_default_pii = true

lewispb:
Closes https://github.com/getsentry/sentry-ruby/issues/1718

## Description
When config.send_default_pii = true we will also log any Redis command arguments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased
 
+### Features
+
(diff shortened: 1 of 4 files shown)


st0012:
@lewispb Thanks for the work! Again! The PR looks great but I need to do one or two patch releases before merging `5.2.0` features. So I need to let it park here for a while.

lewispb:
No worries at all @st0012! I just pushed a test case I missed for the span description.

st0012 (approved):
Thanks again for the follow up work 👍

PR #1734, opened 2022-02-17
Avoid using Resque::Worker#work in tests

st0012:
`Resque::Worker#work` calls `#startup`, which does tons of things for working on jobs:

https://github.com/resque/resque/blob/085f77e87927e3a04bc1c951ca8e36c62ceaef1a/lib/resque/worker.rb#L363-L368

But basically none of them are needed for processing a single job in
tests. So it's a waste of time for performing them.

Some of the methods may cause unstable state for tests, like `#start_heartbeat`.

https://github.com/resque/resque/blob/085f77e87927e3a04bc1c951ca8e36c62ceaef1a/lib/resque/worker.rb#L513-L527

The `#start_heartbeat` method would create a thread for heartbeating. And
it randomly causes build failures like

https://github.com/getsentry/sentry-ruby/runs/5201864477?check_suite_focus=true

I'm not sure if it's a direct or indirect cause of the problem as I'm
not able to reproduce it consistently. But I think by avoiding the `#startup` call 
can surely make the tests both faster and more stable.

### DIFF
--- a/sentry-resque/spec/sentry/resque_spec.rb
+++ b/sentry-resque/spec/sentry/resque_spec.rb
@@ -1,5 +1,9 @@
 require "spec_helper"
 
+def process_job(worker)
+  worker.work_one_job(worker.reserve)
+end
+
 RSpec.describe Sentry::Resque do
   before do
     perform_basic_setup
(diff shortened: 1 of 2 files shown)


st0012:
@sl0thentr0py since this doesn't affect the SDK, I'm merging it now to prevent more random failures, which is annoying.

PR #1738, opened 2022-02-20
Avoid duplicated capturing on the same exception object

### LINKED ISSUE #1731: Duplicate reporting for Sidekiq errors
### Issue Description

After the update to version `5.1.0` I started to get `Sidekiq::JobRetry::Skip` errors in Sentry. After close examination, I found that for each of these errors I also had their cause reported as a separate error in Sentry.

It seems the Sidekiq error handler:
https://github.com/getsentry/sentry-ruby/blob/623d0464ce8120cd2ae3bfdb12113a265ad09d44/sentry-sidekiq/lib/sentry-sidekiq.rb#L28

is creating the sentry error for the cause, and the new error subscription for version `5.1.0`:

https://github.com/getsentry/sentry-ruby/blob/623d0464ce8120cd2ae3bfdb12113a265ad09d44/sentry-rails/lib/sentry/rails/railtie.rb#L123

is creating the sentry error for the `Sidekiq::JobRetry::Skip`

### Reproduction Steps

Use the `5.1.0` version of Sentry and raise an error within a sidekiq job.

### Expected Behavior

The raised error should be logged in Sentry

### Actual Behavior

Two errors are logged in sentry:
* The raised error
* The Sidekiq internal `Sidekiq::JobRetry::Skip`

### Ruby Version

3.1.0

### SDK Version

5.1.0

Rails (7.0.2) + Sidekiq (6.4.1)

### Sentry Config

I don't have any custom error handling configured

st0012:
As described in #1731, the SDK can have duplicated exception report in Rails 7+ applications because of the new `Rails.error` support introduced in version `5.1.0`. 

After a [discussion with the feature author](https://github.com/rails/rails/pull/43625/files#r809532572), we decided to let whichever mechanism captures the exception first report it (could be either `Rails.error` or Sentry's integration in different scenarios). Although it's possible that we'd loss some contextual data with `Rails.error`, the difference is acceptable.

Also, `sentry-python` already has [a similar mechanism](https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/integrations/dedupe.py) to de-dup error reports, so it's been proven to work.

Fixes #1731

### DIFF
--- a/sentry-ruby/lib/sentry/hub.rb
+++ b/sentry-ruby/lib/sentry/hub.rb
@@ -94,6 +94,8 @@ module Sentry
     def capture_exception(exception, **options, &block)
       check_argument_type!(exception, ::Exception)
 
+      return if Sentry.exception_captured?(exception)
+
       return unless current_client
 
       options[:hint] ||= {}
@@ -104,7 +106,7 @@ module Sentry
 
       capture_event(event, **options, &block).tap do
         # mark the exception as captured so we can use this information to avoid duplicated capturing
-        exception.instance_variable_set(:@__sentry_captured, true)
+        exception.instance_variable_set(Sentry::CAPTURED_SIGNATURE, true)
       end
     end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Allow overwriting of context values [#1724](https://github.com/getsentry/sentry-ruby/pull/1724)
   - Fixes [#1722](https://github.com/getsentry/sentry-ruby/issues/1722)
+- Avoid duplicated capturing on the same exception object [#1738](https://github.com/getsentry/sentry-ruby/pull/1738)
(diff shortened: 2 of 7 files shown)


sentry-ruby/lib/sentry/hub.rb
```diff
@@ -104,7 +106,7 @@ def capture_exception(exception, **options, &block)
 
       capture_event(event, **options, &block).tap do
         # mark the exception as captured so we can use this information to avoid duplicated capturing
-        exception.instance_variable_set(:@__sentry_captured, true)
+        exception.instance_variable_set(Sentry::CAPTURED_SIGNATURE, true)
```
st0012, line 109: I feel like the logic of `Sentry.exception_captured?` may be used by users or integrations in the future for checking purpose so I extracted it. But marking exception as captured should only be done here so I didn't add a method for it.

sl0thentr0py (approved):
thx @st0012, looks good!

PR #1742, opened 2022-02-21
ref(envelopes): Move rate limiting logic to each item in envelope

sl0thentr0py:
This makes the envelope logic a bit more consistent to other SDKs because the rate limits can apply to each item in the envelope separately. This is necessary to be able to work with session envelopes in #1715.

### DIFF
--- a/sentry-ruby/lib/sentry/transport.rb
+++ b/sentry-ruby/lib/sentry/transport.rb
@@ -53,7 +53,7 @@ module Sentry
     end
 
     def send_envelope(envelope)
-      process_rate_limits(envelope)
+      reject_rate_limited_items(envelope)
 
       return if envelope.items.empty?
 
@@ -169,7 +169,7 @@ module Sentry
       [item_header, item_payload]
     end
 
-    def process_rate_limits(envelope)
+    def reject_rate_limited_items(envelope)
       envelope.items.reject! do |item|
         if is_rate_limited?(item.type)
           log_info("[Transport] Envelope item [#{item.type}] not sent: rate limiting")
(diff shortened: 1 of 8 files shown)


sentry-ruby/lib/sentry/transport.rb
```diff
@@ -173,6 +168,19 @@ def fetch_pending_client_report
 
       [item_header, item_payload]
     end
+
+    def process_rate_limits(envelope)
```
st0012, line 172: I think `reject_rate_limited_items` gives a better intention here. Or it's possible to extend this method in the future?

sentry-ruby/lib/sentry/transport.rb
```diff
@@ -173,6 +168,19 @@ def fetch_pending_client_report
...
+      envelope.items.reject! do |item|
+        if is_rate_limited?(item.type)
+          log_info("[Transport] Envelope item [#{item.type}] not sent: rate limiting")
```
st0012, line 175: Not related to this PR: I'm thinking downgrading most of the info logs to `debug` level. Because from my experience, we rarely need to know the sent events, not even in debugging. Wdyt?

st0012:
@sl0thentr0py btw we'll need a changelog entry for this as well.

sentry-ruby/lib/sentry/transport.rb
sl0thentr0py: hmm let's do this in a separate PR, I'll need to change a lot of specs otherwise :)

st0012: Yeah I didn't mean we should change it here. And we should also change some other places instead of just here..

st0012 (approved):

Issue #1745, opened 2022-02-24
exception happened in background worker: ActiveRecord::ConnectionNotEstablished


sdogruyol:
### Issue Description

This started to happen after we upgraded our application from Rails 6.1 to Rails 7.0.2.2.

Sentry SDK just crashes in the middle of the request with an unrelated stacktrace and the following error

```
ERROR -- sentry: exception happened in background worker: ActiveRecord::ConnectionNotEstablished
```

The weird part is that our application uses *Sequel* and not ActiveRecord.

### Reproduction Steps

Not available

### Expected Behavior

Should not crash the request

### Actual Behavior

Crashes the request with 500 error

### Ruby Version

3.0.3

### SDK Version

5.0.2

### Integration and Its Version

5.0.2

### Sentry Config

```ruby
# frozen_string_literal: true

Sentry.init do |config|
  config.dsn = ENV.fetch('SENTRY_RAVEN_KEY')
  config.send_default_pii = true
end
```

st0012:
@sdogruyol Thanks for reporting the issue. 

The root cause is that we have [this code](https://github.com/getsentry/sentry-ruby/blob/master/sentry-rails/lib/sentry/rails/background_worker.rb) in the SDK's background worker to make sure we release `ActiveRecord`'s connection if it acquired one. The code was written assuming the client would use `ActiveRecord`, which is not the case for you (and perhaps some other users).

So in #1746, I tried to replace the code with a higher-level helper. But because I don't know how you set up Sequal with your Rails app, I can't test it for you. Can you try the branch and let me know if it fixes the issue?

sdogruyol:
Hey @st0012 thanks a lot for the response.

I just tried your branch and it doesn't seem to fix the problem. Still getting the same `exception happened in background worker: ActiveRecord::ConnectionNotEstablished` exception.

st0012:
@sdogruyol can you set `config.debug = true` in your initializer and provide me the error output? (which should include backtrace)

ajparise:
> The weird part is that our application uses Sequel and not ActiveRecord.

I am experiencing this issue after upgrading sentry-rails from 4.2.2 to 5.1.0.
The app is running Rails 5.2.6.2, it uses Mongoid and AR, but it does not have a persistent or default AR connection setup.

As a workaround I've resorted to synchronous event sending.

```ruby
config.background_worker_threads = 0
```

st0012:
[...]

PR #1746, opened 2022-02-25
Replace manual connection control with abstracted execution wrapper

st0012:
As described in [this document](https://github.com/rails/rails/blob/1d9d53831600c974fcccd4b5574f0e03d7e1513c/guides/source/threading_and_code_execution.md#executor), `Rails.application.executor.wrap` should provide a higher-level of abstraction to achieve concurrent execution, which includes returning acquired ActiveRecord connection.

This prevents us from locking in a particular ORM's database connection mechanism.

### DIFF
--- a/sentry-rails/lib/sentry/rails/background_worker.rb
+++ b/sentry-rails/lib/sentry/rails/background_worker.rb
@@ -2,7 +2,7 @@ module Sentry
   class BackgroundWorker
     def _perform(&block)
       # make sure the background worker returns AR connection if it accidentally acquire one during serialization
-      ActiveRecord::Base.connection_pool.with_connection do
+      ::Rails.application.executor.wrap do
         block.call
       end
     end
--- a/sentry-rails/spec/dummy/test_rails_app/app.rb
+++ b/sentry-rails/spec/dummy/test_rails_app/app.rb
@@ -2,7 +2,7 @@ ENV["RAILS_ENV"] = "test"
 
 require "rails"
 
-require "active_record"
+require "active_record/railtie"
 require "active_job/railtie"
 require "action_view/railtie"
 require "action_controller/railtie"


sentry-rails/lib/sentry/rails/background_worker.rb
```diff
@@ -2,7 +2,7 @@ module Sentry
   class BackgroundWorker
     def _perform(&block)
       # make sure the background worker returns AR connection if it accidentally acquire one during serialization
-      ActiveRecord::Base.connection_pool.with_connection do
+      ::Rails.application.executor.wrap do
```
st0012, line 5: After Rails 7.0, it'll capture the block's exception and report it, which will enter here again and cause infinite loop. So now I think this may be a bad idea 😕

sl0thentr0py: but our `@__sentry_captured` fix will avoid the loop right?

st0012: Not, because the loop will be caused by the exception from SDK (if there's any), instead of the one we capture.
For example, if an event for `ZeroDivisionError` caused a `NoMethodError` during serialization, that `NoMethodError` will be captured and became a new event. If that `NoMethodError` is somehow a problem of the SDK code and/or problematic contextual data, it'll be an infinite loop.

sl0thentr0py: right but then we can just add another rescue here, SDK should fail gracefully internally.
Or move [this rescue](https://github.com/getsentry/sentry-ruby/blob/87b50d78fdfdc4e6946d2d71c84819cf6ed54386/sentry-ruby/lib/sentry/background_worker.rb#L55-L57) over to both `_perform` methods.

st0012: Yeah but that feels ugly. And if this doesn't actually solve #1745, I don't see the necessity to implement it like this.

sl0thentr0py: ah ok just read the follow up in the other issue

Issue #1603, opened 2021-11-05
delayed job integration: payload reached its size limit


lacco:
### Issue Description

We recently switched from `sentry-raven` to the new version of the SDK.
Everything seems to work for Rails HTTP requests, but exception tracking for DelayedJob causes some issues.

From time to time, we see the following log entries:

```
Transaction sending failed: the server responded with status 413
body: {"detail":"failed to read request body","causes":["payload reached its size limit"]}

Event sending failed: the server responded with status 413
body: {"detail":"failed to read request body","causes":["payload reached its size limit"]}
```

This only happens for tracing/ exception tracking in our delayed jobs.

I tried to narrow down the issue in two ways:
1. `config.debug = true` didn't get any additional information
2. `Rails.logger.error(event.inspect)` in `config.before_send` is never logging anything

Any suggestions about what I can do to make sure that none of our exceptions are getting lost?

### Reproduction Steps

Hard to reproduce on our non-production systems

### Expected Behavior

It should send all exceptions that are happening to Sentry

### Actual Behavior

Some exceptions (and probably traces) are missing

### Ruby Version

2.6.3

### SDK Version

sentry-delayed_job=4.7.3, sentry-ruby=4.7.3

### Integration and Its Version

Rails, DelayedJob

### Sentry Config

```
Sentry.init do |config|
  config.breadcrumbs_logger = [:active_support_logger]

  filter = ActiveSupport::ParameterFilter.new(Rails.application.config.filter_parameters)
  config.before_send = lambda do |event, _hint|
    filter.filter(event.to_hash)
  end

  config.traces_sample_rate = 0.2
end
```

st0012:
@lacco it's probably because the breadcrumb generated for the delayed job query. can you try disabling `breadcrumbs_logger` (remove the line)?

lacco:
@st0012  Breadcrumbs are disabled now, exceptions are no longer showing them.

But the issue is still there, is there something else that I can try?

st0012:
@lacco can you try using the `log-event-in-debug-mode` branch with `debug` mode? it should print events in json format (notice: it may be verbose) to help you collect the related event. you can then paste a processed version here or send me privately [on discord](https://discord.com/invite/PXa5Apfe7K).

lacco:
@st0012  After turning off the performance monitoring, we were not able to reproduce this error anymore.
This is very strange, but it also means that we can just close this ticket for the time being.

[...]

PR #1747, opened 2022-02-25
Check envelope size before sending it

st0012:
Because we now need to check if a serialized envelope item is oversized (see https://github.com/getsentry/sentry-ruby/issues/1603#issuecomment-1019216106), envelope serialization is better performed by Transport to have more control to filter oversized items.


st0012:
@sl0thentr0py I know the code could still be refactored (which I'll do later), but can you give it a quick scan and see if the approach looks good?

sl0thentr0py (approved):
logic looks ok :+1:

sl0thentr0py (approved):

PR #1762, opened 2022-03-14
fix: TransactionEvent should not set level to :error

sl0thentr0py:
Currently, since we derive from `Event`, all transactions are marked with a level `:error` which is slightly confusing for the end user.


st0012:
@sl0thentr0py ping ^
also, do you think this should have a changelog entry?

sl0thentr0py:
also no changelog entry, just a small fix

PR #1767, opened 2022-03-18
fix: Set last_event_id only for error events

### LINKED ISSUE #1766: `last_event_id` references transaction event
### Issue Description

When I [call `Sentry.last_event_id` to expose the event ID for the last exception to my event handler](https://docs.sentry.io/clients/ruby/usage/#referencing-events), it includes the ID of a transaction event.

### Reproduction Steps

1. Initialize the client to sample every transaction (see below).
2. Trigger an exception.
3. Call `Sentry.last_event_id`.

### Expected Behavior

`Sentry.last_event_id` should return the event ID for the exception.

### Actual Behavior

`Sentry.last_event_id` returns the event ID for a transaction (`Sentry::TransactionEvent`).

### Ruby Version

3.1.1

### SDK Version

5.2.1

sentry-rails v5.2.1 (Rails v7.0.2.3)

### Sentry Config

```
Sentry.init do |config|
end
```

sl0thentr0py:
`last_event_id` only has a valid use case for cross-referencing errors on the end-user UI, updating it for transactions doesn't make sense and [is consistent with python](https://github.com/getsentry/sentry-python/blob/4410e284ad362e31447e060cb07caae9fb4b64a3/sentry_sdk/hub.py#L323-L324).

Fixes #1766

### DIFF
--- a/sentry-ruby/lib/sentry/hub.rb
+++ b/sentry-ruby/lib/sentry/hub.rb
@@ -150,7 +150,7 @@ module Sentry
         configuration.log_debug(event.to_json_compatible)
       end
 
-      @last_event_id = event&.event_id
+      @last_event_id = event&.event_id unless event.is_a?(Sentry::TransactionEvent)
       event
     end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Unreleased
+
+### Bug Fixes
+
+- Set `last_event_id` only for error events [#1767](https://github.com/getsentry/sentry-ruby/pull/1767)
+  - Fixes [#1766](https://github.com/getsentry/sentry-ruby/issues/1766)
+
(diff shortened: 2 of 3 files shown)


sentry-ruby/lib/sentry/hub.rb
```diff
@@ -150,7 +150,7 @@ def capture_event(event, **options, &block)
         configuration.log_debug(event.to_json_compatible)
       end
 
-      @last_event_id = event&.event_id
+      @last_event_id = event&.event_id unless event.is_a?(Sentry::TransactionEvent)
```
st0012, line 153: @sl0thentr0py this is the reason I want to do the inheritance-removal refactor. they are basically 2 types of events but yet they're both `Sentry::Event`.

st0012 (approved):
can you add a changelog entry for this? thx

Issue #1765, opened 2022-03-18
Dealing with thousands of errors (Or, event_from_exception is slow. Or, Rails 7's error_reporter is awkward)


jdelStrother:
We recently had a cache server go down, which caused our servers to become overwhelmed and start dropping requests like crazy. This seemed unusual: previously we've been fine with a cache server disappearing - we might get a handful of errors, but Rails' cache normally behaves fairly sensibly, just behaving like a cache-miss & silently failing on writes.

I think the new problem lies in a mix of Rails 7's error reporting behaviour, and Sentry's event_from_exception being a relatively slow method.

In Rails 7, every single cache error gets sent to the error_reporter (https://github.com/rails/rails/blob/0169d15bc7ec4557971d6ac6120e48b2cac9c407/activesupport/lib/active_support/cache/redis_cache_store.rb#L460-L466)

These then get sent to Sentry. We're using Sentry's BackgroundWorker to process these asynchronously (and drop events that exceed the `max_queue` size), but before that happens, Sentry calls event_from_exception on the current thread. This seems like quite a heavy method, mostly due to StackTraceBuilder - 

<img width="1452" alt="image" src="https://user-images.githubusercontent.com/2377/158985235-26455533-f177-4b97-ac4a-5cf3db5255bc.png">

and worse, it's called for every single error-report, regardless if BackgroundWorker is about to discard the error because the queue is full. `config.before_send` is called after event_from_exception, so there's not an obvious way for me to control whether that work gets done.

I'm not sure what a good fix would look like here - 
* Possibly the fault lies with Rails 7 - for now, I think I'm going to patch the RedisCacheStore#failsafe method to not call `error_reporter&.report`.  It does seem unfortunate that even if you provide your own `error_handler` callback, ActiveSupport.error_reporter gets called regardless.
* It would be nice if `event_from_exception` happened asynchronously in BackgroundWorker - that way a) it's not blocking the current thread, and b) it would benefit from not doing unnecessary work when we're just about to discard the exception.  But that looks like quite a big change from the current flow, especially if it's keeping compatibility with the other `config.async` options.
* Maybe Sentry.capture_exception could have its own throttle that just discards exceptions if there's been more than X in the past Y seconds ?

Any other suggestions?

st0012:
Thanks for reporting this, but I have a few questions before I can provide suggestions:

[...]

Issue #1754, opened 2022-03-04
(Active-)Job errors not send (sentry-resque)


Flixt:
### Issue Description

Errors that happen within Jobs are not sent to Sentry when using `sentry-rails` and `sentry-resque`

The problem was encountered after upgrading our app to Rails 7 and Ruby 3.0.2

We upgraded:

`sentry-ruby` from `'~> 4.6.5'`, to `'~> 5.1.1'`
`sentry-rails` from `'~> 4.6.5'`, to `'~> 5.1.1'`

And we switched from our own fork of `resque-sentry` to `sentry-resque', '~> 5.1.1'`
We noticed after a while, errors from whithin our jobs are no longer reported to Sentry.

I tried to hunt down the bug and here is what I found so far. Some facts beforehand:

- Rails 7 introduced a global error handling https://github.com/rails/rails/issues/43472
- sentry-ruby does already support this via https://github.com/getsentry/sentry-ruby/blob/master/sentry-rails/lib/sentry/rails/error_subscriber.rb#L7
- When using Sentry with `config.background_worker_threads > 0` the option `{hint: {background: false}} is needed for capture_exception/capture_message
- We run Resque via the `resque:workers` rake task

### Reproduction Steps

I forked the repository and changed the Rails 7.0 example app of sentry-rails in a way to make it possible to reproduce the issue described above.
The diff of the fork can be found here (I did not open a PR as this does not contain a fix):

https://github.com/getsentry/sentry-ruby/compare/master...flowdgmbh:no-resque-events-example

- Clone the fork
- `cd sentry-ruby/sentry-rails/examples/rails-7.0`
- `bundle install`
- Change the dsn in `sentry-rails/examples/rails-7.0/config/initializers/sentry.rb`
- RAILS_ENV=development COUNT=1 QUEUE=* bundle exec rake resque:workers
- In another terminal schedule a job with `bundle exec rails runner 'ErrorJob.perform_later'`

### Expected Behavior

The error is reported to sentry

### Actual Behavior

The following happens for us when an error occours in the job (in the context of the separate Reqsue worker):

- The error is raised in the job
- `Sentry::Rails::ActiveJobExtensions` decides the error is handled by an integration (sentry-resque) in https://github.com/getsentry/sentry-ruby/blob/master/sentry-rails/lib/sentry/rails/active_job.rb#L5
[...]

PR #1771, opened 2022-03-21
Add config.rails.register_error_subscriber to control error reporter integration

### LINKED ISSUE #1731: Duplicate reporting for Sidekiq errors

st0012:
As explained in https://github.com/rails/rails/pull/43625#issuecomment-1072514175, we decided to make the integration disabled by default.

Closes #1731, #1754, #1765

### DIFF
--- a/sentry-rails/spec/sentry/rails_spec.rb
+++ b/sentry-rails/spec/sentry/rails_spec.rb
@@ -209,31 +209,51 @@ RSpec.describe Sentry::Rails, type: :request do
   end
 
   describe "error reporter integration", skip: Rails.version.to_f < 7.0 do
-    before do
-      make_basic_app
-    end
-
-    it "registers Sentry::Rails::ErrorSubscriber to Rails" do
-      Rails.error.report(Exception.new, handled: false)
-
-      expect(transport.events.count).to eq(1)
-
-      ActiveSupport.error_reporter.report(Exception.new, handled: false)
-
-      expect(transport.events.count).to eq(2)
-    end
-
-    it "sets correct contextual data to the reported event" do
-      Rails.error.handle(severity: :info, context: { foo: "bar" }) do
-        1/0
+    context "when config.register_error_subscriber = false (default)" do
+      before do
+        make_basic_app
       end
 
-      expect(transport.events.count).to eq(1)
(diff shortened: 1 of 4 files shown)


sentry-rails/spec/sentry/rails_spec.rb
```diff
@@ -209,31 +209,51 @@
...
-      make_basic_app
-    end
+    context "when config.register_error_subscriber = false (default)" do
+      before do
+        make_basic_app
+      end
 
-    it "registers Sentry::Rails::ErrorSubscriber to Rails" do
-      Rails.error.report(Exception.new, handled: false)
+      it "registers Sentry::Rails::ErrorSubscriber to Rails" do
```
sl0thentr0py, line 217: doesn't register*

sl0thentr0py (approved):

sentry-rails/spec/sentry/rails_spec.rb
st0012: 👍

Issue #1680, opened 2022-01-10
Should the SDK provide test helpers?


st0012:
Since error reporting is usually enabled in staging/production environments, users may not write tests for it. And to be honest, there isn't a clear way to test it either (probably only verifying the arguments are sent to `Sentry.capture_exception` correctly?).

As both a user and a maintainer of the SDK, I sometimes hope there are helpers to verify event payloads when developing apps. Ideally, it'll be something like:

```ruby
# in addition to inject test helpers, it'll also
# - enable the SDK in a test-only mode (haven't decided what that means yet)
# - replace HTTPTransport with DummyTransport
include Sentry::TestHelper

it "reports the exception to Sentry" do
  expect do
    do_something
  end.to have_sent_event_to_sentry(transaction: "PostsController#create", exception: "ZeroDivisionError")
  
  # or for a more detailed verification
  # captured_sentry_events is defined in Sentry::TestHelper
  last_event = captured_sentry_events.last

  expect(last_event.contexts).to match(....)
end
```

st0012:
BTW, to use the helpers, users may need to move `sentry-ruby` out of the production group and make it universal in all environments (test & production group sounds rare). I wonder if that'll cause concern or even issues for some users.

benoittgt:
I am in favor of this proposal. 

I didn't implement this. But we use this in our apps:

```ruby
# ~/code~/app/spec/support/sentry.rb
# frozen_string_literal: true

module SentryHelpers

  def sentry_events
    sentry_transport.events
  end

  def last_sentry_event
    sentry_events.last
  end

  def reset_sentry_events
    sentry_transport.events = []
  end

  private

    def sentry_transport
      Sentry.get_current_hub.current_client.transport
    end

end

RSpec.configure do |config|
  config.include SentryHelpers

  config.around(:each) do |example|
    expect(sentry_events.size).to eq 0

    example.run

    reset_sentry_events
  end
end
```

```ruby
# spec/support/matchers/sentry_matchers.rb
# frozen_string_literal: true

RSpec::Matchers.define :raise_sentry_event do |expected_exception_attributes|
  supports_block_expectations

  match(notify_expectation_failures: true) do |actual|
    actual.call
    @expected_exception_attributes = expected_exception_attributes

    if (expected_exception_attributes.keys - authorized_attributes_keys).present?
      @unauthorized_keys_error = true
      return false
    end

    if sentry_events.size != 1
      @sentry_events_size_error = true
      return false
    end

[...]

PR #1773, opened 2022-03-24
Add `Sentry::TestHelper`

st0012:
As explained in #1680, the SDK should encourage testing the error capturing logic by providing useful tools to users. So as a first step, we decided to add `Sentry::TestHelper` for the following functionalities:

- `setup_sentry_test` - Alters the SDK configuration to for testing. Based on how user sets up the test environment, this may be called once before all tests, or before every test.
- `sentry_events` - An easy way to access all the event objects that have been captured.
- `last_sentry_event` - Basically `sentry_events.last`.
- `teardown_sentry_test` - Clears all captured events.

### Example

```rb
require 'rails_helper'

RSpec.describe "Welcomes", type: :request do
  before do
    setup_sentry_test
  end

  after do
    teardown_sentry_test
  end

  describe "GET /" do
    it "captures and sends exception to Sentry" do
      get "/"
      expect(response).to have_http_status(500)
      expect(sentry_events.count).to eq(2)

      error_event = sentry_events.first
      expect(error_event.transaction).to eq("WelcomeController#index")
      error = extract_sentry_exceptions(error_event).first
      expect(error.type).to eq("ZeroDivisionError")
      expect(error_event.tags).to match(counter: 1, request_id: anything)

      transaction_event = sentry_events.last
      expect(transaction_event.spans.count).to eq(3)
    end
  end
end
```


st0012:
@sl0thentr0py If we're going to do this, I think the `Sentry::TransactionEvent` refactor needs to be done as well. Otherwise the extra interfaces on it (e.g. it still has `#exception`) will be confusing to users.
Another reason is that "conceptually", both `Sentry::Event` and `Sentry::TransactionEvent` will become public to users as they'll be poking with it in tests. And that'll make the "refactor" a breaking change. 

Wdyt?

st0012:
I spent an hour trying the separation and I think the ideal modeling would be:

- `Event = breadcrumbs + request + contexts + tags....etc.`
- `ErrorEvent = Event + threads + exception + level`
- `TransactionEvent = Event + spans`

But that'll require changing the SDK's modeling standard. So I'll leave it for the SDK team to decide and just add comments about the `Sentry::Event` & `Sentry::TransactionEvent` relationship.

sl0thentr0py:
@st0012 I kinda agree with you that this inheritance causes coupling and confusion.

But the other side is that even in the relay protocols we have a single [`Event` type](https://github.com/getsentry/relay/blob/29a9211696f31d3b3682d3a3abb3d05d729edffe/relay-general/src/protocol/event.rs#L172) that then branches based on the `type` field for further processing. So this inheritance/derivation of both `Error` and `Transaction` from `Event` exists throughout Sentry. 

Just another point to consider, I still haven't made up my mind / thought through completely on what I'd like to have here tbh.

st0012:
Oh cool. So on the server side it's actually modeled better, but the SDK side's spec and implementation are the "dirty" ones? 
If that's the case, I think we (or you) can initiate the modeling change from the Ruby SDK?

sl0thentr0py:
Even between SDKs there's a fair amount of inconsistency.
* python doesn't have any `Event` class, just a hash and `event_from_exception` and `transaction.finish` that makes the corresponding event hash.
* js actually has the [cleanest modeling](https://github.com/getsentry/sentry-javascript/blob/8035b14285a34d3b387319f01911d4cc9c741109/packages/types/src/event.ts#L17-L49) here which fits with [your comment](https://github.com/getsentry/sentry-ruby/pull/1773#issuecomment-1081133376) and also is more or less one-one with the relay protocol.

So yes I think basically your comment is the way to go. I'll put it on our backlog but dunno when we'll actually get to it. :)

st0012:
@sl0thentr0py Since that looks like the right direction, I'd be happy to do it this week. Otherwise, it'd become another debt if we exposed them with all the test helpers.

sl0thentr0py:
ok so we do a breaking release then?

sl0thentr0py:
sounds good! I'm slightly worried because the specs will also change but we'll just have to review carefully. But you're right that no end-user API is changing. 👍

sl0thentr0py:
@st0012 do you know what we wanna do with this? Should it go out with the next release or do we wait till later?

st0012:
Since some users start asking for the `close` API for their testing, I think test helpers like these should also benefit them?
I like the idea that SDK provides some helpers for testing and encourage users to test the SDK on CI. In the long-term it should reduce the chance for SDK bug to break on production and will also help issue reports.
But it also comes back to if Sentry wants to support them?

sl0thentr0py (approved):
@st0012 I merged master and updated some specs, do you want to add a changelog before merging?

st0012:
@sl0thentr0py can we wait for the close api first? I want to refresh the implementation here with it.

sl0thentr0py (approved):
lgtm!

PR #1774, opened 2022-03-28
Register SentryContextClientMiddleware on sidekiq workers

jeromepl:
This configuration was missing in order to support the scenario where a sidekiq job itself pushes new jobs.

This meant that although the Sentry user was propagated from the client to the sidekiq server when running a job, it was not propagated again to any new jobs launched from this worker.

This is what is recommend as per [sidekiq's documentation](https://github.com/mperham/sidekiq/wiki/Middleware#client-middleware-registered-in-both-places)


st0012 (approved):
Thank you!

PR #1777, opened 2022-03-29
Correct inaccurate event model relationships

### LINKED ISSUE #1776: Clean up `Event` and `TransactionEvent` inheritance logic
The ideal modeling would be:

- `Event = breadcrumbs + request + contexts + tags....etc.`
- `ErrorEvent = Event + threads + exception + level`
- `TransactionEvent = Event + spans`

_Originally posted by @st0012 in https://github.com/getsentry/sentry-ruby/pull/1773#issuecomment-1081133376_

See also [js implementation](https://github.com/getsentry/sentry-javascript/blob/8035b14285a34d3b387319f01911d4cc9c741109/packages/types/src/event.ts#L17-L49) for reference.

st0012:
As described in #1776 and discussed in https://github.com/getsentry/sentry-ruby/pull/1773, error events shouldn't be the parent of transaction events. They're both variants of the core event type with their unique attributes. So this PR corrects the currently inaccurate inheritance relationship.

Close #1776


sentry-ruby/lib/sentry/event.rb
```diff
@@ -9,7 +9,10 @@
 require 'sentry/utils/custom_inspection'
 
 module Sentry
+  # This is an abstract class that defines the shared attributes of an event.
+  # Please don't use it directly. The user-facing classes are its child classes.
   class Event
+    TYPE = "event"
```
st0012, line 15: @sl0thentr0py By giving all events a default `"event"` type, I think we can remove many fallback code that deals with `nil` type. I think the events can still be accepted without any issue. Let me know if that'll cause any problem?

sentry-ruby/lib/sentry/event.rb
```diff
@@ -71,8 +69,6 @@ def initialize(configuration:, integration_meta: nil, message: nil)
       @rack_env_whitelist = configuration.rack_env_whitelist
 
       @message = (message || "").byteslice(0..MAX_MESSAGE_SIZE_IN_BYTES)
```
st0012, line 71: Ideally, `message` should be moved to `ErrorEvent` as well. But since that'll require changing `Event#initialize` as well and doesn't provide much benefit. I decided not to go this far.

sentry-ruby/lib/sentry/event.rb
sl0thentr0py: oh yea that's perfect, this always felt like a code smell

sl0thentr0py (approved):
nice and clean! don't see any obvious problems

PR #1783, opened 2022-04-04
Add `Sentry.with_child_span` for easier span recording

st0012:
This PR adds a new `Sentry.with_child_span` API. It'll allow users to record a child span with a block without the need to getting the current span from scope, checking its presence...etc. A simple block wrapping will do the trick:

```rb
Sentry.with_child_span(op: "my op") do |child_span|
  # my operation
  # operation result will be the return value
end
```

Note that if there's no span in the current scope, the block will still be executed. But the yield `child_span` will be `nil` instead.

With this API, it's also easier to fix the child span nesting issue reported in #1723.

### DIFF
--- a/sentry-ruby/lib/sentry-ruby.rb
+++ b/sentry-ruby/lib/sentry-ruby.rb
@@ -385,6 +385,40 @@ module Sentry
       get_current_hub.start_transaction(**options)
     end
 
+    # Records the block's execution as a child of the current span.
+    # If the current scope doesn't have a span, the block would still be executed but the yield param will be nil.
+    # @param attributes [Hash] attributes for the child span.
+    # @yieldparam child_span [Span, nil]
+    # @return yield result
+    #
+    # @example
+    #   Sentry.with_child_span(op: "my operation") do |child_span|
+    #     child_span.set_data(operation_data)
+    #     child_span.set_description(operation_detail)
+    #     # result will be returned
+    #   end
+    #
+    def with_child_span(**attributes, &block)
+      current_span = get_current_scope.get_span
+
+      if current_span
+        result = nil
+
+        begin
+          current_span.with_child_span(**attributes) do |child_span|
+            get_current_scope.set_span(child_span)
+            result = yield(child_span)
+          end
+        ensure
+          get_current_scope.set_span(current_span)
+        end
+
+        result
+      else
+        yield(nil)
+      end
+    end
+
     # Returns the id of the lastly reported Sentry::Event.
     #
     # @return [String, nil]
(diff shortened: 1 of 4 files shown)


st0012:
ping @sl0thentr0py

sentry-ruby/lib/sentry-ruby.rb
```diff
@@ -385,6 +385,40 @@ def start_transaction(**options)
...
+    #
+    def with_child_span(**attributes, &block)
+      current_span = get_current_scope.get_span
+
+      if current_span
+        result = nil
+
+        begin
+          current_span.with_child_span(**attributes) do |child_span|
+            get_current_scope.set_span(child_span)
```
sl0thentr0py, line 409: Just an observation, all the current usages of `set_span` set it to the transaction and we don't really use `scope#get_span` either so this will be the first time the span on the scope gets used right?

```
➜  sentry-ruby git:(master) ✗ ag --ignore 'spec' set_span
sentry-rails/lib/sentry/rails/action_cable.rb
18:              scope.set_span(transaction) if transaction

sentry-rails/lib/sentry/rails/active_job.rb
31:                scope.set_span(transaction) if transaction

sentry-resque/lib/sentry/resque.rb
28:              scope.set_span(transaction) if transaction

sentry-delayed_job/lib/sentry/delayed_job/plugin.rb
22:            scope.set_span(transaction) if transaction

sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
21:        scope.set_span(transaction) if transaction

sentry-ruby/lib/sentry/scope.rb
132:    def set_span(span)

sentry-ruby/lib/sentry/rack/capture_exceptions.rb
23:            scope.set_span(transaction) if transaction
```

sl0thentr0py (commented):
sorry for the delay, busy week :/

can you maybe add an integration test too where you add a child span with this new api in an existing requests/response cycle transaction?

sentry-ruby/lib/sentry-ruby.rb
st0012: Yes, because most the usages in the SDK are to create the top-level transaction. But as described in #1784, `sentry-rails`'s subscriber should use spans instead. And users will generally use `get_span` and `set_span` for spans more than transactions.

sentry-ruby/spec/sentry/rack/capture_exceptions_spec.rb
```diff
@@ -405,6 +405,37 @@ def will_be_sampled_by_sdk
...
+          expect(transaction.contexts.dig(:trace, :op)).to eq("rack.request")
+          expect(transaction.spans.count).to eq(2)
+
+          first_span = transaction.spans.first
+          expect(first_span[:op]).to eq("first level")
+          expect(first_span[:parent_span_id]).to eq(transaction.contexts.dig(:trace, :span_id))
+
+          second_span = transaction.spans.last
+          expect(second_span[:op]).to eq("second level")
+          expect(second_span[:parent_span_id]).to eq(first_span[:span_id])
```
st0012, line 436: @sl0thentr0py I've added the integration test here.

sl0thentr0py (approved):
thx!

PR #1788, opened 2022-04-08
Allow config.transport.proxy to be a String or URI

### LINKED ISSUE #1782: Proxy configuration in version 5+ no longer accepts a URL as a string
### Issue Description

We have recently been migrating Rails 5.2 apps from `sentry-raven` to `sentry-ruby`. We have 2 services that are being migrated, one that is on using `sentry-raven` 4.x (v4.8 to be exact, using Faraday). And another that we tried to upgrade to `sentry-ruby` v5.0.2. In the app using v4.8, we were able to use the steps in the (https://docs.sentry.io/platforms/ruby/migration/#renamedrelocated) to rename the proxy configuration option, and pass it a URL string as follows:

```
config.proxy #=> config.transport.proxy
# Example: config.transport.proxy = "http://myproxy.example.com:port"
```

However, doing the same when moving to v5.x throws an error when a proxy string is provided. This caused a production deployment issue for us because we only use a proxy in our production environment, and were not able to test for this in our development/testing environments.

### Reproduction Steps

1. Configure Sentry Ruby using v5+, setting the `transport.proxy` option to a URL string based on the migration guide from [sentry-raven](https://docs.sentry.io/platforms/ruby/migration/#renamedrelocated).
 ```
Sentry.init do |config|
  config.transport.proxy = "http://myproxy.example.com:port"
  ...
end
```
2. Force triggering of an exception in a Rails controller method, such as `Sentry.capture_exception(StandardError.new('This is an exception'))`
  

### Expected Behavior

One of the following 
1. The Sentry Ruby HTTPTransport class should be able to handle the proxy as a URL string, and seamlessly convert the parse the URL string, OR 
2.  Sentry documentation should be updated, including the migration guide, to indicate the configuration value for `config.transport.proxy` should be a `hash`, and a string is **NOT** acceptable.

### Actual Behavior

A `no implicit conversion of Symbol into Integer` exception is thrown when the proxy is set as a string instead of a hash. See the HTTPTransport class [here](https://github.com/getsentry/sentry-ruby/blob/master/sentry-ruby/lib/sentry/transport/http_transport.rb#L132-L137). Therefore, events are not sent to Sentry.

See log

```
2022-04-01T15:20:58Z Sending envelope [event] 081d827b4aca47ecb25328d611027179 to Sentry
2022-04-01T15:20:58Z Event sending failed: no implicit conversion of Symbol into Integer
2022-04-01T15:20:58Z exception happened in background worker: String does not have #dig method
2022-04-01T15:20:58Z Sending envelope [transaction] 7193c21c36534fda8e4030d4e9a2dbd9 to Sentry
2022-04-01T15:20:58Z Transaction sending failed: no implicit conversion of Symbol into Integer
2022-04-01T15:20:58Z Unreported Transaction: V1::XXXXXController#create
2022-04-01T15:20:58Z exception happened in background worker: no implicit conversion of Symbol into Integer
```

### Ruby Version

2.6.6

### SDK Version

5+

Rails/Sidekiq

### Sentry Config

```
Sentry.init do |config|
  config.transport.proxy = "http://myproxy.example.com:port"
  ...
end
```

jknipp:
## Description
Fixes #1782

Update `config.transport.proxy` to allow String and URI values as
previously supported by `sentry-ruby` versions <= 4.8 using Faraday.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
   - [Discussion thread and explanation on the decision](https://github.com/rails/rails/pull/43625#issuecomment-1072514175)
 - Check if ActiveRecord connection exists before calling AR connection pool [#1769](https://github.com/getsentry/sentry-ruby/pull/1769)
   - Fixes [#1745](https://github.com/getsentry/sentry-ruby/issues/1745)
+- Update `config.transport.proxy` to allow String and URI values as previously supported by `sentry-ruby` versions <= 4.8 using Faraday
+  - Fixes [#1782](https://github.com/getsentry/sentry-ruby/issues/1782)
 
 ### Refactoring
 
@@ -79,8 +81,8 @@
 
   <img width="80%" src="https://user-images.githubusercontent.com/6536764/157057827-2893527e-7973-4901-a070-bd78a720574a.png">
 
-  The SDK now supports [automatic session tracking / release health](https://docs.sentry.io/product/releases/health/) by default in Rack based applications.  
-  Aggregate statistics on successful / errored requests are collected and sent to the server every minute.  
+  The SDK now supports [automatic session tracking / release health](https://docs.sentry.io/product/releases/health/) by default in Rack based applications.
+  Aggregate statistics on successful / errored requests are collected and sent to the server every minute.
   To use this feature, make sure the SDK can detect your app's release. Or you have set it with:
 
   ```ruby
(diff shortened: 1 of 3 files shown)


jknipp:
I added a simplified version for [Faraday's proxy normalization](https://github.com/lostisland/faraday/blob/1f496a7f1786c674cb7d9998856678f336439163/lib/faraday/options/proxy_options.rb#L11-L24). I'm not sure there is much additional value in going deeper here, but I can update if necessary.

st0012 (approved):
@jknipp This is an awesome fix. Thank you 👍

jknipp:
Thanks @st0012!  🙌

PR #1797, opened 2022-04-25
Don't mask real error when converting an event to JSON fails

drcapulet:
If the event is unable to be converted to JSON, trying to key into `event_hash` errors because it is `nil`. I'm going to look into how to prevent that issue in a later PR.


st0012:
It looks like such error only happens at the line of `event_hash = event.to_json_compatible`? And thus cause `event_hash` to be `nil`.

If that's the case, perhaps we should rescue that method call to make the case explicit? Something like:

```rb
begin
  event_hash = event.to_json_compatible
rescue => e
  log_error("Converting #{event.inspect} to JSON compatible hash failed", e, debug: configuration.debug)
end
```

st0012:
@drcapulet Do you want to continue on this one? Or you're ok with me opening another PR to finish it?

st0012:
Closing in favor of #1853

PR #1802, opened 2022-04-29
Fix: Don't expect `.backtrace` to be present

jamietanna:
## Description

If a non-`Error` type is passed to the `Raven.capture_exception` method,
for instance when ActiveRecord errors are returned, and we've set up
`backtrace`s to be recorded, Raven itself raises an exception because
the `.backtrace` isn't found.

By checking that `:backtrace` is present first, we can make sure we
don't hit these error conditions.

### DIFF
--- a/sentry-raven/CHANGELOG.md
+++ b/sentry-raven/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - fix: user_context with block does not reset context after block exits
+- fix: Don't expect `.backtrace` to be present
 
 ## 3.1.2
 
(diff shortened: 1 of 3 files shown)


st0012:
I think it's clear that `capture_exception` expects an exception object. Is there any reason you want to pass those objects as exception events, instead of converting them into strings and use `capture_message`?

jamietanna:
Hey @st0012 that's a very fair shout 😁 

This is a codebase I've inherited, so when I spotted we'd had some issues I thought I'd jump into a fix. I didn't consider that we weren't doing the right thing, so thanks for challenging it!

I'll discuss with the team, but `capture_message` is much likely a better solution

Issue #1522, opened 2021-08-01
`async` option will be dropped.


st0012:
The `async` option is unique to the Ruby SDK. It was designed to help send events asynchronously through different backends (e.g. a Ruby thread, Sidekiq worker...etc.). Depends on the backend, it can pose thread to the system due to its additional extra memory consumption. So it's an option with some trade-offs.

But since version `4.1`, the SDK now has its own background worker managed (implemented with the famous `concurrent-ruby` library). It can handle most of the async

## The `async` Option Approach

1. The SDK serializes the event and event hint into json-compatible Ruby hashes.
2. It passes the event payload and hint to the block.
3. In general, the block would enqueue a background job with the above data.
	- Some earlier apps use a new Ruby thread to send the data. This is unrecommended.
	- With background job libraries like Sidekiq or Resque, this means adding objects into Redis.
	- With delayed_job, this means adding a new delayed_job record.
4. A background worker (e.g. Sidekiq worker) then picks the event and hint and send it.

### Pros

Users can customize their event sending logic. But generally it's just a worker with `Sentry.send_event(event, hint)`.

### Cons

- The event payload (usually dozens of kbs) could be copied twice: first copied to the medium storage and then allocates the background worker process.
- When there is an event spike, it can flood the medium storage (Redis) and take down the entire system.

## The Background Worker

1. The SDK passes the event and its hint to the background worker (a pool of threads managed by `concurrent-ruby`).
2. A worker then picks the event, serializes it, and sents it.

### Pros

- It doesn't allocate extra memory other than the original event payload.
- It's faster.
- It doesn't require any user code.
- The background worker doesn't queue more than 30 events. So even when there's a spike, it's unlikely to consume all the memory.

### Cons

- ~~Unsent events will die with the process. Generally speaking, the queue time in background worker is very low. And the chance of missing events due to this reason is small in web apps. But for script programs, the process often leaves before the worker is able to send the event. This is why `hint: { background: false }` is required in rake integrations.~~ 
  - ~~However, I don't think this problem can be solved with the `async` option.~~
  
This drawback has been addressed in https://github.com/getsentry/sentry-ruby/pull/1617.
  
### Missing Events During A Spike Because of Queue Limit

I know many users have concern about the background worker's 30 events queue limit will make them lose events during a spike. But as the maintainer and a user of this SDK, I don't worry about it because:

[...]

PR #1803, opened 2022-04-30
Warn users about config.async's deprecation

st0012:
Since we're going to drop `config.async` in `6.0` (as explained in #1522), I think we should start warning users about the change.

I did my best to make the message clear and compact, but it'd be great if we can have real professionals to polish it a bit 🙂

### DIFF
--- a/sentry-ruby/lib/sentry/configuration.rb
+++ b/sentry-ruby/lib/sentry/configuration.rb
@@ -290,6 +290,16 @@ module Sentry
     def async=(value)
       check_callable!("async", value)
 
+      log_warn <<~MSG
+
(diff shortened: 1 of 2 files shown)


sl0thentr0py (approved):

sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -290,6 +290,16 @@ def dsn=(value)
     def async=(value)
       check_callable!("async", value)
 
+      log_warn <<~MSG
+
+        sentry-ruby now sends events asynchronously by default with its backgrounrd worker (supported since 4.1.0).
```
sl0thentr0py, line 295: typo
```suggestion
        sentry-ruby now sends events asynchronously by default with its background worker (supported since 4.1.0).
```

PR #1811, opened 2022-05-06
explain init explicit

Garllon:
When i used the `sentry-ruby` standalone in my Sinatra project it didn't work, when i only add the gem. 
The `Sentry.init` command needs to be executed once, as you can see in the screenshots, so that the configuration is loaded.

Without init:
Screenshot 2022-05-06 at 10 05 51

With the init command:
Screenshot 2022-05-06 at 10 06 36

### DIFF
--- a/sentry-ruby/README.md
+++ b/sentry-ruby/README.md
@@ -55,13 +55,17 @@ gem "sentry-resque"
 
 ### Configuration
 
-You can use `Sentry.init` to initialize and configure your SDK:
-
+You need to initialize Sentry once with:
 ```ruby
 Sentry.init do |config|
   config.dsn = "MY_DSN"
 end
+```
 
+The DSN can be also set via the `SENTRY_DSN` environment variable, then you
+only need to run:
+```ruby
+Sentry.init
 ```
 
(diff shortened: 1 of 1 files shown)


sl0thentr0py (changes requested):
Hi @Garllon, thx for the suggestion, but to keep it consistent with other SDKs, the README only has the most basic version of the init block. All other ways of configuring are explained in the docs linked below.

For now, I'd suggest just changing the following to clarify that you _need_ to init the SDK.

> You can use `Sentry.init` to initialize and configure your SDK:

to 

> You need to use `Sentry.init` to initialize and configure your SDK:

Garllon:
> Hi @Garllon, thx for the suggestion, but to keep it consistent with other SDKs, the README only has the most basic version of the init block. All other ways of configuring are explained in the docs linked below.
> 
> For now, I'd suggest just changing the following to clarify that you _need_ to init the SDK.
> 
> > You can use `Sentry.init` to initialize and configure your SDK:
> 
> to
> 
> > You need to use `Sentry.init` to initialize and configure your SDK:

Also fine with that. :)

Issue #1808, opened 2022-05-04
Cannot share a single DB connection between a unicorn worker and sentry worker


jakeonfire:
### Issue Description

It is not possible to have a single database connection for a process that then reports a sentry error asynchronously without opening an additional connection.

we have an RDS DB with many connections, but also a huge fleet of sidekiq workers that can scale up and down in the thousands. so doubling the connection count requirement is not currently possible.

### Reproduction Steps

Create a single unicorn worker with a single DB connection that generates an error to report to sentry.

### Expected Behavior

The process reports the error using a single DB connection.

### Actual Behavior

The process requires two DB connections to report the error.

### Ruby Version

2.6.5

### SDK Version

5.3.0

jakeonfire:
i would file this under optimizations. we have many workers with autoscaling and limited DB connections. if sentry could re-use a DB connection if needed instead of requiring a new one that would have a big impact for the number of required connections for largely horizontally scaled applications. even making it optional instead of required would be a win in some cases.

jakeonfire:
one idea: https://stackoverflow.com/a/26733385

jakeonfire:
@st0012 i think my PR solves the original issue without requiring an additional DB connection

st0012:
Closed via #1812

PR #1812, opened 2022-05-07
Don't require a DB connection, but release one if it is acquired

jakeonfire:
solves https://github.com/getsentry/sentry-ruby/issues/1808 which evolved from https://github.com/getsentry/sentry-ruby/pull/1320. tip from https://stackoverflow.com/a/26733385.


st0012 (approved):
Just checked that `#with_connection` uses `#release_connection` underneath, so calling it directly is even better. Thanks for the fix 👍

PR #1824, opened 2022-05-20
Move envelope item processing/trimming logic to the `Item` class

st0012:
I think size-control logic should live inside the `Item` class and `Transport#serialize_envelope` only needs to decide whether the result should be included in the envelope.


sl0thentr0py (approved):
nice, I agree!

PR #1834, opened 2022-06-19
Prepare for Rails 7.1's error reporter API change

st0012:
Rails 7.1 will introduce a new `source` keyword argument to `ErrorReporter` APIs. So we should support that as well.

Related PR: https://github.com/rails/rails/pull/44998

### DIFF
--- a/sentry-rails/lib/sentry/rails/error_subscriber.rb
+++ b/sentry-rails/lib/sentry/rails/error_subscriber.rb
@@ -3,13 +3,13 @@ module Sentry
     # This is not a user-facing class. You should use it with Rails 7.0's error reporter feature and its interfaces.
     # See https://github.com/rails/rails/blob/main/activesupport/lib/active_support/error_reporter.rb for more information.
     class ErrorSubscriber
-      SKIP_SOURCES = [/.*_cache_store.active_support/]
+      SKIP_SOURCES = Regexp.union([/.*_cache_store.active_support/])
 
       def report(error, handled:, severity:, context:, source: nil)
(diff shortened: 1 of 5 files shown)


st0012:
cc @casperisfine

sentry-rails/lib/sentry/rails/error_subscriber.rb
```diff
@@ -3,8 +3,17 @@ module Rails
...
     class ErrorSubscriber
-      def report(error, handled:, severity:, context:)
-        Sentry::Rails.capture_exception(error, level: severity, contexts: { "rails.error" => context }, tags: { handled: handled })
+      SKIP_SOURCES = [/.*_cache_store.active_support/]
+
+      def report(error, handled:, severity:, context:, source: nil)
+        tags = { handled: handled }
+
+        if source
+          return if SKIP_SOURCES.any? { |skip_source| skip_source.match?(source) }
```
casperisfine, line 12:
```suggestion
      SKIP_SOURCES = Regexp.union([/.*_cache_store.active_support/])

      def report(error, handled:, severity:, context:, source: nil)
        tags = { handled: handled }

        if source
          return if SKIP_SOURCES.match?(source)
```

casperisfine (commented):
❤️

sentry-rails/lib/sentry/rails/error_subscriber.rb
st0012: Never knew this usage before, thanks!

sl0thentr0py (approved):

PR #1844, opened 2022-07-12
Implement close API that cleans up state and reverses init logic

sl0thentr0py:
The only logic not reversed is `apply_patches` which is fine since they already check if the patch is already applied and also no-op if `initialized?` is false.

### DIFF
--- a/sentry-ruby/lib/sentry-ruby.rb
+++ b/sentry-ruby/lib/sentry-ruby.rb
@@ -60,11 +60,11 @@ module Sentry
     end
 
     # @!attribute [rw] background_worker
-    #   @return [BackgroundWorker]
+    #   @return [BackgroundWorker, nil]
     attr_accessor :background_worker
 
     # @!attribute [r] session_flusher
-    #   @return [SessionFlusher]
+    #   @return [SessionFlusher, nil]
     attr_reader :session_flusher
 
     ##### Patch Registration #####
@@ -215,8 +215,31 @@ module Sentry
 
       at_exit do
         @session_flusher&.kill
-        @background_worker.shutdown
+        @background_worker&.shutdown
       end
+    end
+
+    # Flushes pending events and cleans up SDK state.
+    # SDK will stop sending events and all top-level APIs will be no-ops after this.
+    #
+    # @return [void]
+    def close
+      if @background_worker
+        @background_worker.shutdown
+        @background_worker = nil
+      end
+
+      if @session_flusher
+        @session_flusher.kill
+        @session_flusher = nil
+      end
+
+      if config.capture_exception_frame_locals
+        exception_locals_tp.disable
+      end
+
+      @main_hub = nil
+      Thread.current.thread_variable_set(THREAD_LOCAL, nil)
     end
 
     # Returns true if the SDK is initialized.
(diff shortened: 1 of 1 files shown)


st0012:
Can we have these test cases?

- It can clean up the initialized SDK
- The SDK can be reinitialized after cleaned up
- It doesn't cause issues when called on the uninitialized SDK
- It doesn't cause issues when called on the cleaned up SDK

sentry-ruby/spec/sentry_spec.rb
```diff
@@ -784,4 +784,141 @@
...
+        described_class.close
+        expect(described_class.session_flusher).to eq(nil)
+      end
+
+      it "disables Tracepoint" do
+        perform_basic_setup do |config|
+          config.capture_exception_frame_locals = true
+        end
+
+        expect(described_class.exception_locals_tp).to receive(:disable)
```
st0012, line 825: This blocks the TP to actually receive `disable` so it'll be left enabled and break other tests.

sentry-ruby/spec/sentry_spec.rb
```diff
@@ -784,4 +784,141 @@
...
+        expect(described_class.configuration).to eq(nil)
+        expect(described_class.csp_report_uri).to eq(nil)
+        expect(described_class.get_main_hub).to eq(nil)
+        expect(described_class.get_current_hub).to eq(nil)
+        expect(described_class.get_current_client).to eq(nil)
+        expect(described_class.get_current_scope).to eq(nil)
+        expect(described_class.last_event_id).to eq(nil)
+      end
+
+      it "no-ops top level API actions" do
```
st0012, line 840: I think we don't need to test these here. We already have [tests](https://github.com/getsentry/sentry-ruby/blob/master/sentry-ruby/spec/initialization_check_spec.rb) testing uninitialized no-op calls.

st0012 (changes requested):

sentry-ruby/spec/sentry_spec.rb
sl0thentr0py: ahh thanks, turns out I still don't get rspec

st0012: You can add `.and_call_original` to let it still perform the call.

sentry-ruby/CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
 
 Individual gem's changelog has been deprecated. Please check the [project changelog](https://github.com/getsentry/sentry-ruby/blob/master/CHANGELOG.md).
 
+## Unreleased
+
+### Features
```
st0012, line 7: These should be at the top-level `CHANGELOG.md`. This one has been deprecated.

sentry-ruby/lib/sentry-ruby.rb
```diff
@@ -215,8 +215,31 @@ def init(&block)
 
       at_exit do
         @session_flusher&.kill
```
st0012, line 217: Perhaps we can use `close` here too?

sl0thentr0py: indeed we can!

PR #1849, opened 2022-07-20
Set sentry.error_event_id in request env if the middleware captures errors

### LINKED ISSUE #1846: rails: allow showing link to sentry page on error page
**Describe the idea**
image

- useful for internal apps to be able to link directly to the error
- useful for public apps to show an error code that engineers can later hunt down

**Possible implementation**
https://github.com/grosser/sentry-user_informer

st0012:
It will allow users/libraries to implement mechanisms displaying relevant error events' ids. Which will help their users report issues.

Closes #1846

### DIFF
--- a/sentry-rails/lib/sentry/rails/capture_exceptions.rb
+++ b/sentry-rails/lib/sentry/rails/capture_exceptions.rb
@@ -32,7 +32,9 @@ module Sentry
           current_scope.set_transaction_name(original_transaction)
         end
 
-        Sentry::Rails.capture_exception(exception)
+        Sentry::Rails.capture_exception(exception).tap do |event|
+          env[ERROR_EVENT_ID_KEY] = event.event_id if event
+        end
       end
 
       def start_transaction(env, scope)
--- a/sentry-ruby/lib/sentry/rack/capture_exceptions.rb
+++ b/sentry-ruby/lib/sentry/rack/capture_exceptions.rb
@@ -3,6 +3,8 @@
 module Sentry
   module Rack
     class CaptureExceptions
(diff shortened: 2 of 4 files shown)


grosser (approved):

sentry-rails/lib/sentry/rails/capture_exceptions.rb
```diff
@@ -32,7 +32,9 @@ def capture_exception(exception, env)
           current_scope.set_transaction_name(original_transaction)
         end
 
-        Sentry::Rails.capture_exception(exception)
+        Sentry::Rails.capture_exception(exception).tap do |event|
+          env[ERROR_EVENT_ID_KEY] = event.event_id if event
```
grosser, line 36: this should also go into the rack middleware [here](https://github.com/getsentry/sentry-ruby/blob/master/sentry-ruby/lib/sentry/rack/capture_exceptions.rb#L57) ?

sentry-ruby/lib/sentry/rack/capture_exceptions.rb
```diff
@@ -53,8 +55,10 @@ def transaction_op
         "rack.request".freeze
       end
 
-      def capture_exception(exception, _env)
-        Sentry.capture_exception(exception)
+      def capture_exception(exception, env)
+        Sentry.capture_exception(exception).tap do |event|
```
st0012, line 59: @grosser This also has the same logic.
I had to replicate them in 2 places because the receivers are different: `Sentry` and `Sentry::Rails` and I don't want to introduce another abstraction just for the 2 occurrences.

sl0thentr0py (approved):
nice, very useful!

PR #1861, opened 2022-08-02
Include Delayed::Job instance in hints

c960657:
It would be useful to include the `Delayed::Job` instance in `hint`, so it can be used for filtering in `before_send`.


st0012:
Sorry for the delay. I think this is helpful but may cause issue with the `async` option, which we plan to drop in `v6.0.0`. 
I'll revisit this after it's dropped.

PR #1867, opened 2022-08-14
Manually require `sidekiq/api` for RetrySet class

st0012:
After https://github.com/mperham/sidekiq/commit/345ba002bc7035d8135d58f4c5dad6dadef298dc
`sidekiq/api` is not required with server logic anymore, which means RetrySet is also not loaded by default. So to use it in tests, we need to require `sidekiq/api` manually.

[Relate failure](https://github.com/getsentry/sentry-ruby/runs/7802900637?check_suite_focus=true)


sl0thentr0py (approved):

st0012:
@sl0thentr0py wow that's quick, thanks!

PR #1869, opened 2022-08-16
Dynamic Sampling Context Continuation / Baggage

sl0thentr0py:
This PR adds support for the first part of our Dynamic Sampling product, i.e.
* parsing incoming `W3C Baggage` headers and adding them as a `Baggage` object on the transaction
* adding this `baggage` header to other outgoing HTTP requests in our `net/http` patch
* converting the `sentry-*` items in the baggage to a DSC (Dynamic Sampling Context) and attaching them to outgoing envelope headers in the `trace` field

now contains #1898 and #1902 

closes #1835

PR #1871, opened 2022-08-18
Remove webmock

sl0thentr0py:
While working on https://github.com/getsentry/sentry-ruby/pull/1869, I noticed that `webmock` [removed ` Net::WebMockNetBufferedIO`](https://github.com/bblimke/webmock/pull/991) in the newer versions.

It seems we have our own stubs anyway so we don't need `webmock` anymore I believe? Let me know if I missed something.


st0012 (approved):
Good catch 👍

PR #1876, opened 2022-08-26
Add basic OpenTelemetry support

mjq-sentry:
Closes https://github.com/getsentry/sentry-ruby/issues/1907

---

For users who already use OpenTelemetry instrumentation or would prefer to use it instead of Sentry's own tracing API, provide a means to disable Sentry's automatic trace instrumentation and a span processor that exports OpenTelemetry spans to Sentry.

## Usage

```ruby
Sentry.init do |config|
  # ...

  # Enable trace collection
  config.traces_sample_rate = 1.0  # or use traces_sampler

  # Disable automatic span instrumentation
  config.auto_instrument_traces = false
end

OpenTelemetry::SDK.configure do |c|
  # ...

  c.add_span_processor(Sentry::OpenTelemetry::SpanProcessor.new)
end
```

Setting `auto_instrument_traces` to `false` in `Sentry.init` stops Sentry from automatically creating any transactions or spans. (Tracing must still be enabled with `traces_sample_rate` or `traces_sampler`.)

`Sentry::OpenTelemetry::SpanProcessor` is notified whenever an OTel span is started or stopped. In reaction, it starts or stops a Sentry transaction (if the span has no parent) or span. When the transaction is finished it is picked up and processed by Sentry's worker thread like any other transaction would be.

If the child service uses the `Sentry::Rack::CaptureExceptions` middleware (see order caveat below), then its transactions will be linked inside Sentry with the parent service's transaction.

## Caveats

* sentry-rails is not yet supported (haven't looked into it yet).
* `Sentry.init` must be run before `OpenTelemetry::SDK.configure` (Sentry's `Net::HTTP` monkey patch to add the `sentry-trace` header must run after OTel's monkey-patch that adds an HTTP span)
* `Sentry::Rack::CaptureExceptions` must be before `OpenTelemetry::Instrumentation::Rack::Middlewares::TracerMiddleware` in the middleware stack (the hub and scope must be set up before OTel creates any spans)
* Since Sentry spans are started and stopped in reaction to OTel spans starting and stopping, the durations reported might be slightly different. (This should be fixable by overriding the timestamps on the Sentry spans we create to match OTel's).
* In the case where an OTel root span has a child span that starts after the root span has finished, this will be represented in Sentry as two transactions (one for the original root span, and one for the late child).


sentry-ruby/lib/sentry/scope.rb
```diff
@@ -7,7 +7,7 @@ module Sentry
   class Scope
     include ArgumentCheckingHelper
 
-    ATTRIBUTES = [:transaction_names, :contexts, :extra, :tags, :user, :level, :breadcrumbs, :fingerprint, :event_processors, :rack_env, :span, :session]
+    ATTRIBUTES = [:transaction_names, :contexts, :extra, :tags, :user, :level, :breadcrumbs, :fingerprint, :event_processors, :rack_env, :span, :session, :trace]
```
mjq-sentry, line 10: Needed some way to get the Sentry Trace (the trace ID/span ID/sample flag we store in the `sentry-trace` header) from the Rack middleware that reads it to the SpanProcessor that actually uses it (to create the linked transaction). Alternatives very welcome.

sentry-ruby/lib/sentry/open_telemetry/span_processor.rb
```diff
@@ -0,0 +1,63 @@
...
+
+      def shutdown(timeout: nil)
+        # no-op: we rely on Sentry.close being called for the same reason as
+        # whatever triggered this shutdown.
+      end
+
+      private
+
+      def from_sentry_sdk?(otel_span)
+        caller.any? { |line| line =~ /lib[\\\/]sentry[\\\/]background_worker.rb/ }
```
mjq-sentry, line 59: This is here mostly to stop transactions from being made out of the HTTP spans that get created when the background worker sends traces to Sentry. I think `caller` is expensive. We could do this from properties of the OTel span instead, but note that `HTTP CONNECT` spans don't have URL info (so we can't check against the dsn like we do in the `Net::HTTP` monkey patch).

sentry-ruby/lib/sentry/open_telemetry/span_processor.rb
```diff
@@ -0,0 +1,63 @@
...
+        return unless sentry_span
+
+        sentry_span.set_op(otel_span.name)
+        current_scope.set_transaction_name(otel_span.name) if sentry_span.is_a?(Sentry::Transaction)
+
+        otel_span.attributes&.each do |key, value|
+          sentry_span.set_data(key, value)
+          if key == "db.statement"
+            sentry_span.set_description(value)
+          end
```
mjq-sentry, line 39: We could apply more transformations here [like the collector does](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/1fc794dfa65e55ab0faadeb16448f7c3daaf5c6f/exporter/sentryexporter/sentry_exporter.go#L226-L337). This was just an easy one to get us started.

dcramer:
Are span processors common in otel sdks? aka could we do this everywhere?

AbhiPrasad:
It's a part of the core spec - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#span-processor, and supported by every major SDK

sentry-ruby/lib/sentry/scope.rb
st0012: We store request env in `rack_env`, so perhaps reading the header from it could also work?

sentry-ruby/lib/sentry/open_telemetry/span_processor.rb
st0012: `caller` [allocates a new string for each frame on the stack](https://github.com/ruby/ruby/blob/58c8b6e86273ccb7a1b903d9ab35956b69b3b1bf/vm_backtrace.c#L402-L430). And in Rails projects, it's easy to have 100+ frames because all the frameworks/libraries loaded. So using `caller` here will have a huge performance impact.

mjq-sentry: Thank you for the source! That's what I feared. Rather than trying to detect whether it's a Sentry span, we could perhaps instead prevent HTTP client spans from turning into transactions. For the majority of use cases I imagine you would only really want to collect them as part of a larger transaction (something associated with a framework).

sl0thentr0py:
ok I'm taking this up now

so let's talk about this `auto_instrument_traces` stuff first.
> Setting `auto_instrument_traces` to false in Sentry.init stops Sentry from automatically creating any transactions or spans. 

Currently only net/http and redis span creation is disabled by this flag. What do we want from this flag? Do we want to turn off ALL instrumentation by sentry if otel is present? To prevent otel/sentry from stepping on each others' toes, we should just no-op most of the performance part of the SDK (except sampling) and that's a bit more work. If that's what we want, I can think of what the best way to do that is.

AbhiPrasad:
Ideally we should turn off all automatic instrumentation that Sentry does with this flag - but still keep user's manual instrumentation. The reason I say this is because if we turn off all instrumentation, then we'll have API issues because we need to access `startTransaction` inside the collector itself.

Turning off all instrumentation is maybe cleaner though, because we can just instruct folks to switch to OpenTelemetry usage completely. Perhaps it's easier to start with turning off all automatic instrumentation? And then we can see how that looks/feels?

sentry-ruby/lib/sentry/open_telemetry/span_processor.rb
```diff
@@ -0,0 +1,63 @@
...
+
+        scope = Sentry.get_current_scope
+        parent_sentry_span = scope.get_span
+
+        sentry_span = if parent_sentry_span
+          parent_sentry_span.start_child(op: otel_span.name)
+        else
+          options = { name: otel_span.name, op: otel_span.name }
+          transaction = Sentry::Transaction.from_sentry_trace(scope.trace, **options) if scope.trace
+          scope.set_trace(nil)
```
sl0thentr0py, line 19: @mjq-sentry is this `set_trace(nil)` necessary for some logic?

sentry-ruby/lib/sentry/scope.rb
sl0thentr0py: done

sl0thentr0py:
ok this is backwards compat now with a new `config.instrumenter` option which defaults to `:sentry`.

sentry-ruby/lib/sentry/open_telemetry/span_processor.rb
sl0thentr0py: I would be fine with just skipping HTTP CONNECT in general.

sl0thentr0py: this has been solved in https://github.com/getsentry/sentry-ruby/pull/1876/commits/26127e5244fcbf377152c0054fd56c02dacd2287#diff-a614943cb316a2889baab64fd91e7d69bfe10bbff44717fcf985655d16988d17R75-R89

basically we can always check `net.peer.name` set [here for CONNECT](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/ea896ed70cc6f6ce1b674c89376c90dfd70f769d/instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb#L57) and [here for the others](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/ea896ed70cc6f6ce1b674c89376c90dfd70f769d/instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb#L27) against our dsn so it seems to work.

sl0thentr0py:
@st0012, @AbhiPrasad  this is very big now but it's ready to ship I think. We can fix small things and iterate on them later after shipping. Please let me know if you have questions! I will comment on the files themselves to explain some decisions for @st0012.

I still wanna write some integration tests but you guys can start reviewing whenever you have time.


Summary: 
* new gem `sentry-opentelemetry` to be used as described [here](https://github.com/getsentry/sentry-ruby/blob/mjq/hackweek-otel/sentry-opentelemetry/README.md#configuration)
* all sentry instrumentation will be disabled by a new `config.instrumeneter` option so that sentry and otel do not step on each other's toes
* the 2 main things are
   * a [`SpanProcessor`](https://github.com/getsentry/sentry-ruby/blob/mjq/hackweek-otel/sentry-opentelemetry/lib/sentry/opentelemetry/span_processor.rb) that hooks into `on_start/on_finish` whenever OpenTelemetry starts/finishes a span. This in turn creates sentry transactions/spans as necessary and keeps track of them in an internal `span_map` hash.
   * a [`Propagator`](https://github.com/getsentry/sentry-ruby/blob/mjq/hackweek-otel/sentry-opentelemetry/lib/sentry/opentelemetry/propagator.rb) to make sure distributed tracing and dynamic sampling work by hooking into `inject/extract` for incoming/outgoing sentry-trace/baggage header management.

sentry-opentelemetry/lib/sentry/opentelemetry/span_processor.rb
```diff
@@ -0,0 +1,159 @@
+require 'singleton'
+
+module Sentry
+  module OpenTelemetry
+    TraceData = Struct.new(:trace_id, :span_id, :parent_span_id, :parent_sampled, :baggage)
+
+    class SpanProcessor < ::OpenTelemetry::SDK::Trace::SpanProcessor
+      include Singleton
```
sl0thentr0py, line 8: I made it a singleton because `span_map` needs to be accessed in the propagator and I didn't want to add a global variable.

sentry-opentelemetry/lib/sentry/opentelemetry/span_processor.rb
```diff
@@ -0,0 +1,159 @@
...
+      def on_start(otel_span, parent_context)
+        return unless Sentry.initialized? && Sentry.configuration.instrumenter == :otel
+        return unless otel_span.context.valid?
+        return if from_sentry_sdk?(otel_span)
+
+        trace_data = get_trace_data(otel_span, parent_context)
+
+        sentry_parent_span = @span_map[trace_data.parent_span_id] if trace_data.parent_span_id
+
+        sentry_span = if sentry_parent_span
```
sl0thentr0py, line 29: the root span from otel becomes a transaction and all other child spans become child spans of the transaction in sentry

sentry-opentelemetry/lib/sentry/opentelemetry/propagator.rb
```diff
@@ -0,0 +1,73 @@
...
+        sentry_trace_data = Transaction.extract_sentry_trace(sentry_trace)
+        return context unless sentry_trace_data
+
+        context = context.set_value(SENTRY_TRACE_KEY, sentry_trace_data)
+        trace_id, span_id, _parent_sampled = sentry_trace_data
+
+        span_context = ::OpenTelemetry::Trace::SpanContext.new(
+          trace_id: [trace_id].pack('H*'),
+          span_id: [span_id].pack('H*'),
+          # we simulate a sampled trace on the otel side and leave the sampling to sentry
```
sl0thentr0py, line 45: sampling the transaction is left to sentry, so we enforce sampling on the otel side, we might iterate on this later for better sampling compatibility with otel.

sentry-rails/lib/sentry/rails/tracing/abstract_subscriber.rb
```diff
@@ -43,7 +43,7 @@ def record_on_current_span(duration:, **options)
             Sentry.with_child_span(**options) do |child_span|
               # duration in ActiveSupport is computed in millisecond
               # so we need to covert it as second before calculating the timestamp
-              child_span.set_timestamp(child_span.start_timestamp + duration / 1000)
+              child_span&.set_timestamp(child_span.start_timestamp + duration / 1000)
```
st0012, line 46: In `sentry-rails`'s case, I think we may just remove all the subscribers (either not subscribe them or unsubscribe them, depends on the timing), instead of making them run and deal with `nil` span?

sentry-ruby/lib/sentry/hub.rb
```diff
@@ -92,7 +93,9 @@ def start_transaction(transaction: nil, custom_sampling_context: {}, **options)
       transaction
     end
 
-    def with_child_span(**attributes, &block)
+    def with_child_span(instrumenter: :sentry, **attributes, &block)
```
st0012, line 96: Could users use both instrumenters (`sentry` and `otel`) at the same time? Or it's simpler like: if users chose `otel`, they'll give up `sentry` tracing completely.

sentry-opentelemetry/lib/sentry/opentelemetry/span_processor.rb
st0012: By globally I assume you mean within the threads of the same process? 

When reading the doc, I see

> This means that the hub used in onStart is the same hub used in onEnd and it means that hubs fork properly in async contexts.

But in Ruby SDK, every thread would have their own `Hub`. So I'm not sure if a global `SpanProcessor` can achieve that when it'll probably always use the main hub.

st0012 (commented):
I'm not familiar with `OpenTelemetry` so I'm learning as I review this and have some questions:

- Is the goal to let users use [opentelemetry-ruby](https://github.com/open-telemetry/opentelemetry-ruby) but send the events to Sentry? If that's true, I assume we want them to opt-in to one side completely? e.g. Not tracing some parts with open-telemetry and the rest with Sentry's current tracing APIs/integrations.
- What concurrency model does `opentelemetry-ruby` support and does it fit with our a Hub/thread design?
- The PR description says `... disable Sentry's automatic trace instrumentation and a span processor that exports OpenTelemetry spans to Sentry.`
    - What'd happen if some Sentry's automatic traces slip in?


Suggestions:

- We probably want to add something in `sentry-opentelemetry` to enforce/warn users about

    > The SDK is initialized before the OpenTelemetry SDK is initialized.

- I feel a helper like `Sentry.with_native_tracing?` could be helpful. Like in [sentry-rails's railties](https://github.com/getsentry/sentry-ruby/blob/master/sentry-rails/lib/sentry/rails/railtie.rb#L48), we can just NOT register tracing subscribers if it's `false`. Then we won't need to touch the subscribers' logic.

- Most of the changes in `sentry-ruby` could be extracted to a separate PR.

sentry-opentelemetry/lib/sentry/opentelemetry/span_processor.rb
sl0thentr0py: I think this part of the spec is outdated, since we no longer use scope to do anything, it shouldn't matter.
@abhiprasad

sentry-ruby/lib/sentry/hub.rb
sl0thentr0py: for now, it's one or the other. I don't know what direction we will take on this on the future, depends on a lot of parameters like adoption/future of SDKs etc

sentry-rails/lib/sentry/rails/tracing/abstract_subscriber.rb
sl0thentr0py: yea, good point, I wasn't sure where to do this. Will just disable the subscribers.

sl0thentr0py: done

sl0thentr0py:
> Is the goal to let users use [opentelemetry-ruby](https://github.com/open-telemetry/opentelemetry-ruby) but send the events to Sentry?

Yes, exactly, we want to make it easier for users using otel to switch over to sentry for performance while continuing to use sentry for errors.


> If that's true, I assume we want them to opt-in to one side completely? e.g. Not tracing some parts with open-telemetry and the rest with Sentry's current tracing APIs/integrations.

Yes, this is what the `instrumenter` option does, it's a switch between either sentry/otel. We might generalize this later depending on adoption but for now, this is what we will ship.

> What concurrency model does opentelemetry-ruby support and does it fit with our a Hub/thread design?

The initial design was using scope but we removed that so our hub/scope management should not matter here anymore.
All concurrency handling in otel is done by the [`Context`](https://github.com/open-telemetry/opentelemetry-ruby/blob/18bfd391f2bda2c958d5d6935886c8cba61414dd/api/lib/opentelemetry/context.rb#L126-L128) and we just rely on them to do the right thing.

> What'd happen if some Sentry's automatic traces slip in?

It's fine, not the worst. This is very much an MVP/experiment so we'll just fix it if someone reports a bug. 
I've tried to make sure that doesn't happen though.

> Most of the changes in sentry-ruby could be extracted to a separate PR.

I'll split this up into smaller PRs later today. 👍

sentry-opentelemetry/lib/sentry/opentelemetry/span_processor.rb
AbhiPrasad: Yup the spec is outdated, will fix in develop!

sl0thentr0py:
Tracking list of new PRs:
* https://github.com/getsentry/sentry-ruby/pull/1944

sl0thentr0py:
closing in favour of split PRs

PR #1877, opened 2022-08-29
Pin redis to <5 for sidekiq

sl0thentr0py:
See https://github.com/mperham/sidekiq/issues/5498

### DIFF
--- a/sentry-rails/Gemfile
+++ b/sentry-rails/Gemfile
@@ -29,6 +29,9 @@ gem "mini_magick"
 gem "sprockets-rails"
 
 gem "sidekiq"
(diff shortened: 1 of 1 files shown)


sl0thentr0py:
`resque` is also failing, opened issue upstream
https://github.com/resque/resque/issues/1821

st0012:
For this locking to work, we need to do that in specific integration's folders. And it looks like `sidekiq` has merged the PR, so maybe we only need to do that in `sentry-resque`?

sl0thentr0py:
yea I was going to close this tomorrow anyway, let's just wait it out for upstream.

PR #1879, opened 2022-09-05
bugfix: sentry crash when exeception message is an array.

xiaoronglv:
## Problem

when the exception message is not a string, the sentry will crash and sentry SDK will cause main application to crash.

Here is an example to crash sentry.

For example

```
raise StandardError.new(message: ["msg1", "msg2"])
```

I know the developer should use a standard string as the exception
message, but not all developers are well educated. Sentry as a service
provider can't always assume the exception message is a string.

Sometimes it's a hash or array.

## Solution

if the exception message is not a string, the `byteslice` should not be called.

### DIFF
--- a/sentry-ruby/lib/sentry/interfaces/single_exception.rb
+++ b/sentry-ruby/lib/sentry/interfaces/single_exception.rb
@@ -15,7 +15,7 @@ module Sentry
 
     def initialize(exception:, stacktrace: nil)
       @type = exception.class.to_s
-      @value = (exception.message || "").byteslice(0..Event::MAX_MESSAGE_SIZE_IN_BYTES)
+      @value = (exception.message || "").try(:byteslice, 0..Event::MAX_MESSAGE_SIZE_IN_BYTES).to_s
       @module = exception.class.to_s.split('::')[0...-1].join('::')
       @thread_id = Thread.current.object_id
       @stacktrace = stacktrace


st0012:
Which Ruby version are you using? In the versions of Ruby I tested (as early as Ruby 2.4), `Exception#message` only returns String even if it's initialized with non-string objects:

```
irb(main):001:0> StandardError.new(message: [1, 2]).message
=> "{:message=>[1, 2]}"
irb(main):003:0* RUBY_VERSION
=> "2.4.10"
```

xiaoronglv:
Hi @st0012 ,

I have updated the script to reproduce this issue.

```
class CustomError < StandardError
  attr_accessor :message

  def initialize(message)
    @message = message
  end
end


e = CustomError.new(["dasdf", "asfd"])


Sentry.capture_exception e
```

Then, you will get this error.

```
/var/www/shared/bundle/ruby/3.0.0/gems/sentry-ruby-5.4.1/lib/sentry/interfaces/single_exception.rb:18:in `initialize': undefined method `byteslice' for ["dasdf", "asfd"]:Array (NoMethodError)
```

st0012:
It's the author of `CustomError`'s responsibility to make sure the `message` override maintains the same behavior. So in this case, the `checkr` gem should maintain that compatibility.

st0012:
In the `checkr`'s case, it overrides an API that it shouldn't and probably don't need to override. And in my experience this doesn't happen often. Also, the root cause can be easily addressed.

It's not like we expect users to follow all Ruby best practices to use the SDK. The question here imo is "whether the SDK should prepare for unconventional patches from 3rd-party code". And if the answer is yes, then our codebase will need to handle all the edge cases that could be produced out of our control, which isn't practical.

PR #1882, opened 2022-09-06
Remove resque redis pin since upstream is fixed

sl0thentr0py:
https://github.com/resque/resque/pull/1828
https://github.com/resque/resque/pull/1822
https://github.com/resque/resque/releases/tag/v2.4.0

### DIFF
--- a/sentry-resque/Gemfile
+++ b/sentry-resque/Gemfile
@@ -17,6 +17,3 @@ gem "sentry-rails", path: "../sentry-rails"
 
 gem "pry"
 gem "debug", github: "ruby/debug", platform: :ruby if RUBY_VERSION.to_f >= 2.6
-
-# remove this after https://github.com/resque/resque/pull/1828 is merged and released
-gem "redis", "< 5.0"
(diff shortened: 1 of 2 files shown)


sl0thentr0py:
hmm still segfaulting somewhere else :(

st0012:
This is actually an issue between Ruby 2.6 & `debug` & `redis-client`. I think it's probably Ruby 2.6's issue. I can reproduce it with:

```rb
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  ruby "~> 2.6.6"

  gem "debug"
  gem "redis-client"
end
```

This change should workaround the issue:

```diff
--- a/sentry-resque/spec/spec_helper.rb
+++ b/sentry-resque/spec/spec_helper.rb
@@ -1,6 +1,6 @@
 require "bundler/setup"
 require "pry"
-require "debug" if RUBY_VERSION.to_f >= 2.6
+require "debug" if RUBY_VERSION.to_f >= 2.7

 require "resque"
```

st0012:
I opened an issue for the crash here: https://github.com/redis-rb/redis-client/issues/47
And the latest failure seems to be caused by Rack 3. We'll need to start testing multiple Rack versions I think.

sl0thentr0py:
ah yes Rack 3 makes sense, thx for the heads up.
Will try to fix in another PR in a bit.

st0012:
`redis-client` has fixed the issue so we probably wont need the workaround either. Perhaps we can drop the version pin in the Rack PR instead?

PR #1898, opened 2022-09-15
Baggage creation for head SDK

sl0thentr0py:
WIP


sentry-ruby/lib/sentry/transaction.rb
```diff
@@ -78,7 +85,12 @@ def self.from_sentry_trace(sentry_trace, baggage: nil, hub: Sentry.get_current_h
           sampled_flag != "0"
         end
 
+      # If there's an incoming sentry-trace but no incoming baggage header,
+      # for instance in traces coming from older SDKs,
+      # baggage will be empty and frozen and won't be populated as head SDK.
       baggage = Baggage.from_incoming_header(baggage) if baggage
```
st0012, line 91: Looks like this part can be reduced to just 1 condition:

```rb
baggage = if baggage
  Baggage.from_incoming_header(baggage)
else
  Baggage.new({})
end

baggage.freeze!
```

sentry-ruby/lib/sentry/transaction.rb
```diff
@@ -139,6 +154,12 @@ def set_initial_sample_decision(sampling_context:)
         return
       end
 
+      @effective_sample_rate = if [true, false].include?(sample_rate)
```
st0012, line 157: I feel this part can be merged with the above `unless` condition (line 151)?

sentry-ruby/lib/sentry/transaction.rb
```diff
@@ -205,6 +234,34 @@ def generate_transaction_description
...
 
+    def populate_head_baggage
+      sentry_items = {}
+      sentry_items["trace_id"] = trace_id
+
+      # TODO-neel filter for high cardinality / low quality tx source here
+      sentry_items["transaction"] = name
+      sentry_items["sample_rate"] = effective_sample_rate&.to_s
+
+      configuration = Sentry.configuration
```
st0012, line 245: If configuration is needed in these new methods, we can use `@configuration` and remove its todo comment.

sentry-ruby/lib/sentry/transaction.rb
```diff
@@ -205,6 +234,34 @@ def generate_transaction_description
...
+        sentry_items["public_key"] = configuration.dsn&.public_key
+      end
+
+      user = Sentry.get_current_scope&.user
+      sentry_items["user_segment"] = user["segment"] if user && user["segment"]
+
+      # there's an existing baggage but it was mutable,
+      # which is why we are creating this new baggage.
+      # However, if by chance the user put some sentry items in there, give them precedence.
+      sentry_items.merge!(@baggage.sentry_items) if @baggage && !@baggage.sentry_items.empty?
```
st0012, line 259: Do we cover this scenario in tests?

sentry-ruby/lib/sentry/transaction.rb
sl0thentr0py: `from_incoming_header` also returns `nil` so we also need to handle that
https://github.com/getsentry/sentry-ruby/blob/bd4c9dbb32c816149a17909c91db40c1b3714f80/sentry-ruby/lib/sentry/baggage.rb#L46

sl0thentr0py: or I can move that check outside if you prefer that.

sentry-ruby/lib/sentry/transaction.rb
sl0thentr0py: not really, but it's also some 'possible future support' type edge-case that I added in because other SDKs added it. I'm partially debating not having it at all.

sentry-ruby/lib/sentry/transaction.rb
sl0thentr0py: we should decide on a way going forward, I'm fine with using the global since that's what python does too. But if you think having the local variable is cleaner, we can go with that too. What do you prefer?

sentry-ruby/lib/sentry/transaction.rb
st0012: Yeah I think that'll reduce the complexity here (esp. don't need to remember that `from_incoming_header` also returns `nil`) and the method.

sentry-ruby/lib/sentry/transaction.rb
st0012: I'm keen to reduce global state because it caused me some problems when taking over `sentry-raven`, especially with its tests.

sl0thentr0py: I just added all the variables explicitly in initialize

sentry-ruby/lib/sentry/transaction.rb
sl0thentr0py: I just removed it since JS also simplified their implementation recently and removed some cruft logic.

st0012 (approved):

PR #1900, opened 2022-09-17
Make sure test helper clears the current scope before/after a test

st0012:
Because the `setup_sentry_test` helper provides a clean client, the scope should be clean too. I also imagine `teardown_sentry_test` includes clearing the current scope's contextual data.

And because there's not a `bind_client` equivalent API for scope, we clear the scope's value with `Scope#clear` instead of creating a new scope. But in practice that both of them should work the same.


francois-ferrandis (approved):
Looks good, thank you!

PR #1901, opened 2022-09-17
Configure Rails 7.1's test app separately

st0012:
Rails 7.1 stores the error reporter directly under the `ActiveSupport` class (PR: https://github.com/rails/rails/pull/46029). So we need to make sure the subscriber is not subscribed unexpectedly before any tests.

And because both the change and the `unsubscribe` API is introduced in Rails 7.1, we should start having a dedicated test setup for it now.


st0012:
ping @sl0thentr0py

sl0thentr0py (approved):

PR #1902, opened 2022-09-19
Add transaction source annotation

### LINKED ISSUE #1866: Transaction source
Annotate all transaction naming exhaustively with sources as in 
https://develop.sentry.dev/sdk/event-payloads/transaction/#transaction-annotations

sl0thentr0py:
See https://develop.sentry.dev/sdk/event-payloads/transaction/#transaction-annotations for spec.

This is necessary going forward because we need to separate high cardinality / low quality (for us, only raw urls from rack) to enable ingestion to do metrics aggregation without blowing up.

Note: This is independent in principle, but I'm basing it off the baggage changes since all of this needs to go out together anyway and some of the `Transaction` constructors changed and would cause merge conflicts otherwise.

closes #1866

### DIFF
--- a/sentry-rails/lib/sentry/rails/action_cable.rb
+++ b/sentry-rails/lib/sentry/rails/action_cable.rb
@@ -13,8 +13,8 @@ module Sentry
             Sentry.with_scope do |scope|
               scope.set_rack_env(env)
               scope.set_context("action_cable", extra_context) if extra_context
-              scope.set_transaction_name(transaction_name)
-              transaction = start_transaction(env, scope.transaction_name)
+              scope.set_transaction_name(transaction_name, source: :view)
+              transaction = start_transaction(env, scope.transaction_name, scope.transaction_source)
               scope.set_span(transaction) if transaction
 
               begin
@@ -29,11 +29,11 @@ module Sentry
             end
           end
 
-          def start_transaction(env, transaction_name)
+          def start_transaction(env, transaction_name, transaction_source)
             sentry_trace = env["HTTP_SENTRY_TRACE"]
             baggage = env["HTTP_BAGGAGE"]
 
-            options = { name: transaction_name, op: "rails.action_cable".freeze }
+            options = { name: transaction_name, source: transaction_source, op: "rails.action_cable".freeze }
             transaction = Sentry::Transaction.from_sentry_trace(sentry_trace, baggage: baggage, **options) if sentry_trace
             Sentry.start_transaction(transaction: transaction, **options)
           end
--- a/sentry-rails/lib/sentry/rails/action_cable.rb
+++ b/sentry-rails/lib/sentry/rails/action_cable.rb
@@ -14,7 +14,7 @@ module Sentry
               scope.set_rack_env(env)
               scope.set_context("action_cable", extra_context) if extra_context
               scope.set_transaction_name(transaction_name, source: :view)
-              transaction = start_transaction(env, scope.transaction_name, scope.transaction_source)
+              transaction = start_transaction(env, scope)
               scope.set_span(transaction) if transaction
 
               begin
@@ -29,11 +29,11 @@ module Sentry
             end
           end
 
-          def start_transaction(env, transaction_name, transaction_source)
+          def start_transaction(env, scope)
             sentry_trace = env["HTTP_SENTRY_TRACE"]
             baggage = env["HTTP_BAGGAGE"]
 
-            options = { name: transaction_name, source: transaction_source, op: "rails.action_cable".freeze }
+            options = { name: scope.transaction_name, source: scope.transaction_source, op: "rails.action_cable".freeze }
             transaction = Sentry::Transaction.from_sentry_trace(sentry_trace, baggage: baggage, **options) if sentry_trace
             Sentry.start_transaction(transaction: transaction, **options)
           end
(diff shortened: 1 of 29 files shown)


sentry-ruby/lib/sentry/scope.rb
```diff
@@ -74,6 +89,7 @@ def dup
       copy.tags = tags.deep_dup
       copy.user = user.deep_dup
       copy.transaction_names = transaction_names.deep_dup
+      copy.transaction_sources = transaction_sources.deep_dup
```
st0012, line 92: I think this and the `transaction_names` don't need `deep_dup`, just `dup` should be enough. Especially in the `transaction_sources`'s case, the content are all `Symbol` and are not mutable.



I can't recall why I used `deep_dup` at `transaction_names`, but given that strings are generally frozen nowadays and there's no point mutating them in practice, I think it can switch to `dup` too.

sentry-rails/lib/sentry/rails/action_cable.rb
```diff
@@ -29,11 +29,11 @@ def capture(connection, transaction_name:, extra_context: nil, &block)
             end
           end
 
-          def start_transaction(env, transaction_name)
+          def start_transaction(env, transaction_name, transaction_source)
```
st0012, line 32: Let's align this with `CaptureExceptions#start_transaction` to just pass a scope?

sentry-ruby/lib/sentry/transaction.rb
```diff
@@ -14,12 +14,19 @@ class Transaction < Span
...
+    SOURCES = %i(custom url route view component task)
+
     include LoggingHelper
 
     # The name of the transaction.
     # @return [String]
     attr_reader :name
 
+    # The source of the transaction name.
+    # @return [String]
```
st0012, line 27: It looks like in most cases it'll be `Symbol` instead?

sentry-ruby/lib/sentry/transaction.rb
```diff
@@ -43,10 +50,18 @@ class Transaction < Span
...
+      name: nil,
+      source: :custom,
+      parent_sampled: nil,
+      baggage: nil,
+      **options
+    )
       super(**options)
 
       @name = name
+      @source = SOURCES.include?(source) ? source : :custom
```
st0012, line 64: We should call a `to_sym` on `source` just in case.

sentry-ruby/lib/sentry/event.rb
```diff
@@ -18,7 +18,7 @@ class Event
       event_id level timestamp
       release environment server_name modules
       message user tags contexts extra
-      fingerprint breadcrumbs transaction
```
st0012, line 21: It looks like a `TransactionEvent`-only attribute? If that's the case, can we put it there?

sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
```diff
@@ -37,8 +37,8 @@ def build_tags(tags)
         Array(tags).each_with_object({}) { |name, tags_hash| tags_hash[:"sidekiq.#{name}"] = true }
       end
 
-      def start_transaction(transaction_name, sentry_trace)
-        options = { name: transaction_name, op: "sidekiq" }
+      def start_transaction(transaction_name, transaction_source, sentry_trace)
```
st0012, line 40: It feels like we can also just take `scope` here.

sentry-ruby/lib/sentry/transaction.rb
```diff
@@ -242,20 +264,26 @@ def generate_transaction_description
...
-        "transaction" => name,# TODO-neel filter for high cardinality tx source
         "sample_rate" => effective_sample_rate&.to_s,
         "environment" => @environment,
         "release" => @release,
         "public_key" => @dsn&.public_key
       }
 
+      items["transaction"] = name unless source_low_quality?
+
       user = Sentry.get_current_scope&.user
```
st0012, line 275: Can use `@hub.current_scope` here

sentry-rails/lib/sentry/rails/action_cable.rb
sl0thentr0py: yea good idea, thanks

sentry-ruby/lib/sentry/event.rb
sl0thentr0py: I had it there first, but then moved it here because it's always coupled to the `transaction` field.

sentry-ruby/spec/sentry/transaction_spec.rb
```diff
@@ -479,9 +488,19 @@
           "environment" => "development",
           "public_key" => "12345",
           "trace_id" => subject.trace_id,
-          "transaction"=>"foo"
+          "transaction"=>"foo",
+          "sample_rate" => "1.0"
```
antonpirker, line 492: Why is here not checked for the source? What am I missing?

sentry-ruby/spec/sentry/transport_spec.rb
```diff
@@ -13,7 +13,7 @@
 
   let(:client) { Sentry::Client.new(configuration) }
   let(:hub) do
-    Sentry::Hub.new(client, subject)
+    Sentry::Hub.new(client, Sentry::Scope.new)
```
st0012, line 16: 🤦‍♂️ my bad

st0012 (approved):
👍

sentry-ruby/spec/sentry/transaction_spec.rb
antonpirker: nevermind

PR #1909, opened 2022-10-03
Prepare changelog for 5.5.0

sl0thentr0py:


st0012:
Is there any side-effect about the dynamic sampling feature that we may want to mention? Like potential increase in quota consumption or additional transactions?

st0012 (approved):

PR #1912, opened 2022-10-05
Release connection pool only when active record is connected

matthewrfindley:
#1769 Fixed an issue where accessing the connection pool on a rails application that did not use active record, but had the active record gem installed.

#1812 Attempted to ensure connection pools were released, but introduced a regression for rails apps without active record connections established.

This change re-introduces the guard to ensure active record is connected before attempting to access the connection pool.

If you access the connection_pool attribute without active record being connected, the result is an `ActiveRecord::ConnectionNotEstablished` error.

Testing this bug-fix is a bit tricky. It would require setting up a rails application with active record installed, but does not establish a connection. AFAICT there is no such testing done in this project to allow for testing this easily. I am open to suggestions on a direction for introducing a new testing strategy within this project. Please advise if you feel it is necessary to add coverage for this bug fix.


st0012:
@matthewrfindley Thanks for the PR 👍 

Notes for future readers of this change:

Both [`ActiveRecord::Base.connected?`](https://github.com/rails/rails/blob/e24fbf71fa0a10f63346b9b42604451958740844/activerecord/lib/active_record/connection_adapters/abstract/connection_handler.rb#L248-L251) and [`ActiveRecord::Base.connection_pool`](https://github.com/rails/rails/blob/e24fbf71fa0a10f63346b9b42604451958740844/activerecord/lib/active_record/connection_handling.rb#L282-L284) ends up calling [`ActiveRecord::ConnectionAdapters::ConnectionHandler#retrieve_connection_pool`](https://github.com/rails/rails/blob/e24fbf71fa0a10f63346b9b42604451958740844/activerecord/lib/active_record/connection_adapters/abstract/connection_handler.rb#L262-L265) to get the connection pool, which would return `nil` if there's no connection pool config. But when that happens, `ActiveRecord::Base.connection_pool` [raises the `ConnectionNotEstablished` exception](https://github.com/rails/rails/blob/e24fbf71fa0a10f63346b9b42604451958740844/activerecord/lib/active_record/connection_handling.rb#L283) and `connected?` does not.

st0012 (approved):

PR #1915, opened 2022-10-09
Allow users to configure their asset-skipping pattern

st0012:
Some users have customized asset paths, so the SDK's default pattern may not work for them. In such cases, they can use the new configuration option to set their own path pattern.

```rb
Sentry.init do |config|
  config.rails.assets_regexp = /my_regexp/
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 ## Unreleased
 
+### Features
+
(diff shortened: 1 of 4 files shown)


sl0thentr0py (approved):

st0012:
@sl0thentr0py I'll add a changelog before merging it

PR #1920, opened 2022-10-20
Use `Sentry.with_child_span` in redis and net/http patches

sl0thentr0py:
Move away from using `span.start_child` explicitly and use the top level `Sentry.with_child_span`.

This is necessary going forward for having some top-level api control for disabling sentry instrumentation if we're using open telemetry.

Note: I had to remove one net/http test for the unsampled case since that behaviour changed now. But this is fine as in principle the `sampled` flag just controls the final `hub.capture_event` and not the actual span recording.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
     config.rails.assets_regexp = /my_regexp/
   end
   ```
+- Use `Sentry.with_child_span` in redis and net/http instead of `span.start_child` [#1920](https://github.com/getsentry/sentry-ruby/pull/1920)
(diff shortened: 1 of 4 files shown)


AbhiPrasad (approved):

st0012 (approved):
This may be nitpicking, but if there's a behaviour change, is it still just a refactoring?
Should the user notice anything from the changes? e.g. different span nesting

sl0thentr0py:
@st0012 yea fair point, the nesting might change (be more accurate).
I'll move the changelog entry.

st0012 (approved):

PR #1922, opened 2022-10-21
Fix setting the sentry-trace header in net/http

sl0thentr0py:
Sorry I made a mistake in #1920 that wasn't caught by the tests that just check the request object after everything happens.
The header should be set before the request actually runs in `super`.


st0012:
@sl0thentr0py Good catch 👍 Can we add a regression test for that case?

st0012 (approved):

PR #1924, opened 2022-10-23
Use `Exception#detailed_message` when generating exception message if applicable 

st0012:
Ruby 3.1 provided more detailed exception messages like

```
undefined method `[]' for nil:NilClass

          {}[:foo][:bar]
                  ^^^^^^
```

But the additional information was put directly input `Exception#message`, which caused some problems (as described in [this ticket](https://bugs.ruby-lang.org/issues/18438)).

So in Ruby 3.2, the additional message will be put into the `Exception#detailed_message` ([proposal](https://bugs.ruby-lang.org/issues/18438)), which as a error monitoring service is what we should use.

### DIFF
--- a/sentry-rails/spec/sentry/rails_spec.rb
+++ b/sentry-rails/spec/sentry/rails_spec.rb
@@ -103,7 +103,7 @@ RSpec.describe Sentry::Rails, type: :request do
         expect(response.status).to eq(500)
 
         expect(event["exception"]["values"][0]["type"]).to eq("RuntimeError")
-        expect(event["exception"]["values"][0]["value"]).to eq("An unhandled exception!")
+        expect(event["exception"]["values"][0]["value"]).to match("An unhandled exception!")
       end
     end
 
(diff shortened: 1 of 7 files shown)


sentry-rails/spec/sentry/rails_spec.rb
```diff
@@ -103,7 +103,7 @@
         expect(response.status).to eq(500)
 
         expect(event["exception"]["values"][0]["type"]).to eq("RuntimeError")
-        expect(event["exception"]["values"][0]["value"]).to eq("An unhandled exception!")
+        expect(event["exception"]["values"][0]["value"]).to match("An unhandled exception!")
```
st0012, line 106: `#detailed_message` always includes `(ExceptionClass)`, which I think is beneficial and should be kept. 
But I also don't want to check Ruby version before every message check, so I just switch to `match` instead.

st0012:
cc @mame

sentry-ruby/lib/sentry/interfaces/single_exception.rb
```diff
@@ -15,7 +15,15 @@ class SingleExceptionInterface < Interface
 
     def initialize(exception:, stacktrace: nil)
       @type = exception.class.to_s
-      @value = (exception.message || "").byteslice(0..Event::MAX_MESSAGE_SIZE_IN_BYTES)
+      exception_message =
+        if exception.respond_to?(:detailed_message)
+          exception.detailed_message(highlight: false)
```
st0012, line 20: - `highlight: false` removes ANSI escape code that's mainly for standard output.
- `#detailed_message` always returns a string. Even if the exception's message is `nil`, it still returns `"(Exception)"`.

sl0thentr0py (approved):
nice! the only (negligible) concern I have is if this somehow affects grouping?
But I think it should be fine, let's merge it.

PR #1926, opened 2022-10-25
Fix return in do..end block

sl0thentr0py:
bug introduced in #1919
see https://github.com/getsentry/sentry-ruby/pull/1919#discussion_r1003397568

#skip-changelog


gareth:
The change in this branch allows our tests to pass where the previous branch failed 👍🏻

st0012 (approved):

PR #1930, opened 2022-11-02
Make `sentry-sidekiq` compatible with Sidekiq 7

st0012:
Most changes are for testing, but it also fixes an incompatible call in the `ErrorHandler` class.

### DIFF
--- a/sentry-sidekiq/lib/sentry/sidekiq/error_handler.rb
+++ b/sentry-sidekiq/lib/sentry/sidekiq/error_handler.rb
@@ -3,6 +3,8 @@ require 'sentry/sidekiq/context_filter'
 module Sentry
   module Sidekiq
     class ErrorHandler
+      WITH_SIDEKIQ_7 = ::Gem::Version.new(::Sidekiq::VERSION) >= ::Gem::Version.new("7.0")
+
       def call(ex, context)
         return unless Sentry.initialized?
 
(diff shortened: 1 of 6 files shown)


sentry-sidekiq/lib/sentry/sidekiq/error_handler.rb
```diff
@@ -40,7 +42,13 @@ def retry_limit(context)
         when Integer
           limit
         when TrueClass
-          ::Sidekiq.options[:max_retries] || 25
+          max_retries =
```
sl0thentr0py, line 45: I haven't looked at this before, but why does sentry have to deal with the retry logic at all?

sentry-sidekiq/lib/sentry/sidekiq/error_handler.rb
```diff
@@ -40,7 +42,13 @@ def retry_limit(context)
...
+            if WITH_SIDEKIQ_7
+              ::Sidekiq.default_configuration[:max_retries]
```
sl0thentr0py, line 47: is `default_configuration` the right thing if the user overrides `max_retries`?

sentry-sidekiq/lib/sentry/sidekiq/error_handler.rb
st0012: Because with `config.sidekiq.report_after_job_retries = true`, the SDK should wait for the retry quota exhausted before reporting.

sentry-sidekiq/lib/sentry/sidekiq/error_handler.rb
st0012: From [the only invocation of `max_retries`](https://github.com/mperham/sidekiq/blob/ad9442f116086424948d60ec059755f9168a6bc8/lib/sidekiq/job_retry.rb#L73), I think users need to either assign the `default_configuration` value, or configure it at job level (which is handled in the `Integer` case).

So yeah, in this condition `default_configuration` is the right place to check.

sl0thentr0py (approved):
thx for explaining!

PR #1944, opened 2022-11-22
Add config.instrumenter to switch between sentry and otel instrumentation

sl0thentr0py:
The new `config.instrumenter` will act as a top-level switch to disable on sentry transactions/spans when we are in `:otel` mode.

* Disable rails subscribers when instrumenter is not `:sentry`
* Noop top level `start_transaction` and `with_child_span` APIs when the instrumenter don't match with the currently active instrumenter (defaults to `:sentry`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Unreleased
+
+### Features
+
+- Add OpenTelemetry support with new `sentry-opentelemetry` gem
+  - Add `config.instrumenter` to switch between sentry and otel instrumentation [#1944](https://github.com/getsentry/sentry-ruby/pull/1944)
+
 ## 5.6.0
 
 ### Features
--- a/sentry-delayed_job/spec/sentry/delayed_job_spec.rb
+++ b/sentry-delayed_job/spec/sentry/delayed_job_spec.rb
@@ -284,6 +284,30 @@ RSpec.describe Sentry::DelayedJob do
         event = transport.events.last
         expect(event.contexts.dig(:trace, :trace_id)).to eq(transaction.contexts.dig(:trace, :trace_id))
       end
+
+      context "with instrumenter :otel" do
+        before do
+          perform_basic_setup do |config|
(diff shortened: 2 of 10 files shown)


AbhiPrasad (approved):

st0012:
Should this be applied to `sentry-sidekiq`, `sentry-resque`, and `sentry-delayed_job` as well? 
If the answer is yes, can we encapsulate `Sentry.configuration.instrumenter == :sentry` with something like `Sentry.configuration.native_instrumenter?` or `Sentry.configuration.default_instrumenter?`?

sl0thentr0py:
@st0012 so the idea was to make this config as non-interfering as possible, since we don't know the future of this feature.

I don't want to add top-level apis we commit to for now, so the idea was to no-op the top-level hub `start_transaction`/`with_child_span` apis and also not to change every callsite. All the other gems/integrations should remain backwards compat with minimal change.

I added the rails check because of the span nil issue, which is fine in this case I think to avoid all the checks.

Potential users of `sentry-opentelemetry` will also ideally not add `sentry-rails|sidekiq|delayed_job|resque` since they will be using the otel sdk for instrumentation, so this is also another argument be made for leaving the other integrations untouched.

AbhiPrasad:
> Potential users of sentry-opentelemetry will also ideally not add sentry-rails|sidekiq|delayed_job|resque since they will be using the otel sdk for instrumentation, so this is also another argument be made for leaving the other integrations untouched.

Do we do error monitoring/add breadcrumbs via sentry-rails|sidekiq|delayed_job|resque? If yes, this will mean that they have to use these integrations anyway.

sl0thentr0py:
hmm yea, nvm you're right, ignore that second argument
the other argument about not changing the call-sites still holds.

st0012:
@sl0thentr0py Can you also add tests to those integrations to make sure they don't raise errors when the instrumenter is `otel`?

sl0thentr0py:
added config/otel specs to other gems

PR #1946, opened 2022-11-23
Expose `end_timestamp` in `Span#finish` and `Transaction#finish`

sl0thentr0py:


AbhiPrasad (approved):

sentry-ruby/lib/sentry/span.rb
```diff
@@ -90,11 +90,11 @@ def initialize(
 
     # Finishes the span by adding a timestamp.
     # @return [self]
-    def finish
+    def finish(end_timestamp: nil)
       # already finished
       return if @timestamp
```
st0012, line 95: This isn't mentioned in the spec so I want to double check it: we only allow the first override but not the second time or later?
If that's true, can we add a test case for it?

sentry-ruby/lib/sentry/span.rb
```diff
@@ -90,11 +90,11 @@ def initialize(
 
     # Finishes the span by adding a timestamp.
     # @return [self]
```
st0012, line 92: We forgot to mention that it could return `nil` too (early return)

st0012 (commented):
I think this can base on `master` instead.

sentry-ruby/lib/sentry/span.rb
sl0thentr0py: ah no, it should override either way, good catch!

sentry-ruby/lib/sentry/span.rb
sl0thentr0py: I think it should never return nil, will also fix

st0012 (approved):
I think we can merge this once the changelog entry is updated + the build is fixed by updating the base branch.

PR #1947, opened 2022-11-23
Add `Transaction#set_context` api

sl0thentr0py:
* allow arbitrary `contexts` to be set on the transaction directly
* also including two other changes which are not changelog worthy
  * add internal `Transaction#set_name` api to set name on transaction directly
  * extract `sentry-trace` regex logic into `Transaction.extract_sentry_trace` helper for reusability

### DIFF
--- a/sentry-ruby/lib/sentry/transaction.rb
+++ b/sentry-ruby/lib/sentry/transaction.rb
@@ -210,7 +210,7 @@ module Sentry
     # Finishes the transaction's recording and send it to Sentry.
     # @param hub [Hub] the hub that'll send this transaction. (Deprecated)
     # @return [TransactionEvent]
-    def finish(hub: nil)
+    def finish(hub: nil, end_timestamp: nil)
       if hub
         log_warn(
           <<~MSG
@@ -222,7 +222,7 @@ module Sentry
 
       hub ||= @hub
 
-      super() # Span#finish doesn't take arguments
+      super(end_timestamp: end_timestamp)
 
       if @name.nil?
         @name = UNLABELD_NAME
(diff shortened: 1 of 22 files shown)


sentry-ruby/lib/sentry/transaction.rb
```diff
@@ -244,6 +257,24 @@ def get_baggage
       @baggage
     end
 
+    # Set the transaction name directly.
+    # Considered internal api since it bypasses the usual scope logic.
+    # @param name [String]
+    # @param source [Symbol]
+    # @return [void]
+    def set_name(name, source: :custom)
```
st0012, line 265: It looks like we can also use this in the constructor to replace:

```rb
      @name = name
      @source = SOURCES.include?(source) ? source.to_sym : :custom
```

st0012 (commented):
This also feel like an independent change to me so can base on `master` directly.
Can we add test case to clarify the precedence of `Scope` and `Transaction`'s `contexts`, likely in `sentry_spec.rb` or `hub_spec.rb`?

sl0thentr0py:
@st0012 regarding the based on master, I'll merge them in order so they'll each get their updated base as master.
I'll have to solve changelog merge conflicts every time otherwise.

st0012:
IMO, all of these PRs deserve their own commits on master and independent changlog entries (not under the telemetry one) because 1) they're self-contained and 2) they also affect people who don't use `sentry-opentelemetry`.
I understand it's a bit tedious to resolve conflicts over and over, but it's worth doing to keep the history clear.

sl0thentr0py:
I will move the changelog entries, but the history will be clear, I don't want to re-create all these branches.
If they get merged in order, the base will be master every time, they are just chained right now.

st0012 (commented):
This seems to be mixed with other PRs' changes?

AbhiPrasad (approved):

PR #1948, opened 2022-11-23
Add new `sentry-opentelemetry` gem

sl0thentr0py:
### Summary
* new gem `sentry-opentelemetry` to be used as described [here](https://github.com/getsentry/sentry-ruby/blob/neel/otel-gem/sentry-opentelemetry/README.md#getting-started)
* all sentry instrumentation will be disabled by a new `config.instrumeneter` option so that sentry and otel do not step on each other's toes
* the 2 main things are
   * a [`SpanProcessor`](https://github.com/getsentry/sentry-ruby/blob/mjq/hackweek-otel/sentry-opentelemetry/lib/sentry/opentelemetry/span_processor.rb) that hooks into `on_start/on_finish` whenever OpenTelemetry starts/finishes a span. This in turn creates sentry transactions/spans as necessary and keeps track of them in an internal `span_map` hash.
   * a [`Propagator`](https://github.com/getsentry/sentry-ruby/blob/mjq/hackweek-otel/sentry-opentelemetry/lib/sentry/opentelemetry/propagator.rb) to make sure distributed tracing and dynamic sampling work by hooking into `inject/extract` for incoming/outgoing sentry-trace/baggage header management.

---
### Spec
https://develop.sentry.dev/sdk/performance/opentelemetry/

#skip-changelog

### DIFF
--- a/sentry-opentelemetry/README.md
+++ b/sentry-opentelemetry/README.md
@@ -37,9 +37,9 @@ gem "opentelemetry-instrumentation-all"
 
 ### Configuration
 
-First, make sure to initialize Sentry before OpenTelemetry by prefixing your initializers and set the `instrumenter` to `:otel`.
+Initialize Sentry for tracing and set the `instrumenter` to `:otel`.
 ```ruby
-# config/initializers/01_sentry.rb
+# config/initializers/sentry.rb
 
 Sentry.init do |config|
   config.dsn = "MY_DSN"
@@ -53,7 +53,7 @@ This will disable all Sentry instrumentation and rely on the chosen OpenTelemetr
 Next, configure OpenTelemetry as per your needs and hook in the Sentry span processor and propagator.
 
 ```ruby
-# config/initializers/02_otel.rb
+# config/initializers/otel.rb
 
 OpenTelemetry::SDK.configure do |c|
   c.use_all
--- a/sentry-opentelemetry/README.md
+++ b/sentry-opentelemetry/README.md
@@ -1,0 +1,65 @@
+<p align="center">
+  <a href="https://sentry.io" target="_blank" align="center">
+    <img src="https://sentry-brand.storage.googleapis.com/sentry-logo-black.png" width="280">
+  </a>
+  <br>
+</p>
+
+# sentry-opentelemetry, the OpenTelemetry integration for Sentry's Ruby client
+
+---
+
+
+[![Gem Version](https://img.shields.io/gem/v/sentry-opentelemetry.svg)](https://rubygems.org/gems/sentry-opentelemetry)
+![Build Status](https://github.com/getsentry/sentry-ruby/workflows/sentry-opentelemetry%20Test/badge.svg)
+[![Coverage Status](https://img.shields.io/codecov/c/github/getsentry/sentry-ruby/master?logo=codecov)](https://codecov.io/gh/getsentry/sentry-ruby/branch/master)
+[![Gem](https://img.shields.io/gem/dt/sentry-opentelemetry.svg)](https://rubygems.org/gems/sentry-opentelemetry/)
+[![SemVer](https://api.dependabot.com/badges/compatibility_score?dependency-name=sentry-opentelemetry&package-manager=bundler&version-scheme=semver)](https://dependabot.com/compatibility-score.html?dependency-name=sentry-opentelemetry&package-manager=bundler&version-scheme=semver)
+
+
+[Documentation](https://docs.sentry.io/platforms/ruby/guides/opentelemetry/) | [Bug Tracker](https://github.com/getsentry/sentry-ruby/issues) | [Forum](https://forum.sentry.io/) | IRC: irc.freenode.net, #sentry
+
+The official Ruby-language client and integration layer for the [Sentry](https://github.com/getsentry/sentry) error reporting API.
+
+
+## Getting Started
+
+### Install
+
+```ruby
+gem "sentry-ruby"
+gem "sentry-rails"
+gem "sentry-opentelemetry"
+
+gem "opentelemetry-sdk"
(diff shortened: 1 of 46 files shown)


sl0thentr0py:
@st0012 do you still want to review this separately or are you ok with us shipping this?

sentry-opentelemetry/README.md
```diff
@@ -0,0 +1,65 @@
...
+  config.instrumenter = :otel
+end
+```
+
+This will disable all Sentry instrumentation and rely on the chosen OpenTelemetry tracers for creating spans.
+
+Next, configure OpenTelemetry as per your needs and hook in the Sentry span processor and propagator.
+
+```ruby
+# config/initializers/otel.rb
```
st0012, line 56: Are these instructions tested? AFAIK, Rails initializers are run in alphabetical order so this will be ran BEFORE `sentry.rb`, which contradicts with what our document says:

```
The SDK is initialized before the OpenTelemetry SDK is initialized.
```

st0012:
@sl0thentr0py I want to review this separately because I was focusing on other changes in my first pass.
I'm leaving some comments now.

sentry-opentelemetry/lib/sentry-opentelemetry.rb
```diff
@@ -0,0 +1,6 @@
+require "sentry-ruby"
```
st0012, line 1: Can we add `# frozen_string_literal: true` to all of this gem's files?

sentry-opentelemetry/lib/sentry/opentelemetry/span_processor.rb
```diff
@@ -0,0 +1,159 @@
...
+
+      private
+
+      def from_sentry_sdk?(otel_span)
+        dsn = Sentry.configuration.dsn
+        return false unless dsn
+
+        if otel_span.name.start_with?("HTTP")
+          # only check client requests, connects are sometimes internal
+          return false unless %i(client internal).include?(otel_span.kind)
```
st0012, line 89: `%i(client internal)` could be a constant so we don't need to init a new array repeatedly.

sentry-opentelemetry/lib/sentry/opentelemetry/span_processor.rb
```diff
@@ -0,0 +1,159 @@
...
+      end
+
+      def on_start(otel_span, parent_context)
+        return unless Sentry.initialized? && Sentry.configuration.instrumenter == :otel
+        return unless otel_span.context.valid?
+        return if from_sentry_sdk?(otel_span)
+
+        trace_data = get_trace_data(otel_span, parent_context)
+
+        sentry_parent_span = @span_map[trace_data.parent_span_id] if trace_data.parent_span_id
```
st0012, line 27: Still a little concerning having this hash shared between multiple-threads. Could 2 or more otel spans share the same `parent_span_id`?

sentry-opentelemetry/lib/sentry/opentelemetry/span_processor.rb
```diff
@@ -0,0 +1,159 @@
...
+      include Singleton
+
+      SEMANTIC_CONVENTIONS = ::OpenTelemetry::SemanticConventions::Trace
+
+      # The mapping from otel span ids to sentry spans
+      # @return [Hash]
+      attr_reader :span_map
+
+      def initialize
+        @span_map = {}
```
st0012, line 17: I think we may store `configuration.logger` and `configuration.instrumenter` here so we don't need to perform configuration lookup repeatedly.

sentry-opentelemetry/README.md
sl0thentr0py: I told abhi to remove it, precedence order doesn't matter anymore.

sentry-opentelemetry/lib/sentry/opentelemetry/span_processor.rb
sl0thentr0py: we're just getting the sentry span corresponding to the otel span, multiple spans can have the same parent and that shouldn't cause any problems.

Regarding concurrency, we can eventually wrap stuff in mutexes _if someone complains_ but given that ruby hashes work with the GIL in cruby at least, this is fine for now. We might need to use `ConcurrentHashMap` in jruby or other implementations, but I would not worry about this now.

This is an MVP, we're getting it out and seeing if there's interest at all in such features. We will iterate once there is proven demand from the community.

sentry-opentelemetry/lib/sentry/opentelemetry/span_processor.rb
sl0thentr0py: * if we do this, the precedence order of initialization comes back because you need access to the instantiated `Sentry.configuration` object 
* eventually we might do all the otel config automatically ourselves and then we can fix all precedence problems in our setup, but we don't want to make assumptions on user setups right now

for now, I will just kill the logger statements since they were mostly for development and keep the other stuff as is.

sl0thentr0py:
@st0012 bump, is there something more that you want to discuss here?
I'm going on vacation next week, and we want to release this in the next few days so there's some time for bug reports etc before I leave.

Once again, this is a standalone gem and it does not have to be the perfect implementation, the main idea is 
* to start ingesting otel spans into sentry in the simplest and most low-friction way possible
* ensure distributed tracing works (via propagator)
* collect feedback on improving SDK/data/product experience and iterate on it later

We have thought through some of the trade-offs and are aware of some possible problems, your feedback is of course very welcome.

st0012 (approved):
Yeah I think we can test it first. Just one more thing: do you want to add it to the top-level` readme too?

sl0thentr0py:
docs PR merged, will be live soon
https://github.com/getsentry/sentry-docs/pull/5856

st0012 (approved):

PR #1963, opened 2022-12-11
Support redis-rb 5.0+

### LINKED ISSUE #1931: Redis instrumentation doesn't work with latest redis-rb library version
### Issue Description

I don't see any Redis traces in Sentry anymore. I think it's since redis-rb was updated a while back to v5.

I believe the issue is that you are patching the `logging` method in `Redis::Client` [here](https://github.com/getsentry/sentry-ruby/blob/master/sentry-ruby/lib/sentry/redis.rb#L74). But, that method has been removed as you can see in the latest `Redis::Client` code [here](https://github.com/redis/redis-rb/blob/master/lib/redis/client.rb).

### Reproduction Steps

Run code that calls Redis, using redis-rb version > 5

### Expected Behavior

There should be a redis trace span in Sentry

### Actual Behavior

There is no redis trace Span in Sentry. It just says "Missing Instrumentation".

### Ruby Version

3.1.2

### SDK Version

5.5.0

_No response_

### Sentry Config

_No response_

st0012:
This fixes #1931 

Changes:
- Use real Redis for testing because `fakeredis` doesn't support `redis-rb 5.0+`
- Add `REDIS_RB_VERSION` to control installed `redis-rb`'s version and test against bot `~> 4.0` and `~> 5.0` on CI
- Make `sentry-ruby` works with `redis-rb 5.0+` too.

### DIFF
--- a/.github/workflows/sentry_ruby_test.yml
+++ b/.github/workflows/sentry_ruby_test.yml
@@ -21,34 +21,71 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        ruby_version: [2.4, 2.5, 2.6, 2.7, '3.0', head, jruby]
+        ruby_version: [2.4, 2.5, 2.6, 2.7, "3.0", head, jruby]
         rack_version: [2.0, 3.0]
+        redis_rb_version: [4.0]
         os: [ubuntu-latest]
         include:
-          - { os: ubuntu-latest, ruby_version: 3.1, rack_version: 0 }
-          - { os: ubuntu-latest, ruby_version: 3.1, rack_version: 2.0 }
-          - { os: ubuntu-latest, ruby_version: 3.1, rack_version: 3.0 }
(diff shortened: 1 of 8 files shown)


st0012:
@sl0thentr0py Can you help me check why danger is failing?

```
Request failed [403]: https://api.github.com/user
Response: {
  "message": "Resource not accessible by integration",
  "documentation_url": "https://docs.github.com/rest/reference/users#get-the-authenticated-user"
}
Request failed [403]: https://api.github.com/user
Response: {
  "message": "Resource not accessible by integration",
  "documentation_url": "https://docs.github.com/rest/reference/users#get-the-authenticated-user"
}
```

sl0thentr0py:
@st0012 no clue, but those logs are always present. I'll report upstream..

Not liking the danger experience so far, I'll look into something better once I'm back at work in January.
For now, feel free to ignore it if it's not changelog related, we removed the required status on that check.

sl0thentr0py (approved):

PR #1966, opened 2022-12-13
Return value from 'perform_action' when initialized

ronocod:
We use `ActionCable::Channel::Base#perform_action` in our Action Cable testing and would like to use its return value as part of those tests. At the minute, when the SDK is not initialized then this works OK, as the return value of `block.call` is immediately returned: https://github.com/getsentry/sentry-ruby/blob/dce8ef069ddf2393497863e6c62a22221a4454ff/sentry-rails/lib/sentry/rails/action_cable.rb#L9

However, if we initialize the SDK, then the return value is swallowed. This change ensures that the result of `block.call` is returned in all situations, apart from when an error is raised in the block.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,6 @@
 ### Features
 
 - Allow [tags](https://docs.sentry.io/platforms/ruby/enriching-events/tags/) to be passed via the context hash when reporting errors using ActiveSupport::ErrorReporter and Sentry::Rails::ErrorSubscriber in `sentry-rails` [#1932](https://github.com/getsentry/sentry-ruby/pull/1932)
-- Return value from `perform_action` in ActionCable::Channel instances when initialized [#1966](https://github.com/getsentry/sentry-ruby/pull/1966)
 
 ### Bug Fixes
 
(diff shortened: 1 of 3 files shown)


sentry-rails/lib/sentry/rails/action_cable.rb
```diff
@@ -20,8 +20,9 @@ def capture(connection, transaction_name:, extra_context: nil, &block)
               scope.set_span(transaction) if transaction
 
               begin
-                block.call
+                result = block.call
                 finish_transaction(transaction, 200)
+                return result
```
st0012, line 25: This `return` is not needed:

```suggestion
                result
```

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### Features
 
 - Allow [tags](https://docs.sentry.io/platforms/ruby/enriching-events/tags/) to be passed via the context hash when reporting errors using ActiveSupport::ErrorReporter and Sentry::Rails::ErrorSubscriber in `sentry-rails` [#1932](https://github.com/getsentry/sentry-ruby/pull/1932)
+- Return value from `perform_action` in ActionCable::Channel instances when initialized [#1966](https://github.com/getsentry/sentry-ruby/pull/1966)
```
st0012, line 6: I think this categorises as bug fix though.

st0012 (approved):
Thanks for the fix, just 2 small comments 👍

st0012:
Rebasing `master` should resolve the CI issue.

PR #1968, opened 2022-12-19
Tag an ActiveRecord span as cached

kkumler:
When digging through a transaction, we could not readily confirm if a repeated ActiveRecord query was using the cache or not. The information is already sent along with the `sql.active_record` instrumentation since [Rails 5.1](https://github.com/rails/rails/blob/6cacbac68e25e1c98b838409498a70e9b4f3bf12/activerecord/lib/active_record/connection_adapters/abstract/query_cache.rb#L152) but only from the QueryCache origin.
The tag is added only if `cached: true` is in the payload.

### DIFF
--- a/sentry-rails/spec/sentry/rails/tracing/active_record_subscriber_spec.rb
+++ b/sentry-rails/spec/sentry/rails/tracing/active_record_subscriber_spec.rb
@@ -51,10 +51,10 @@ RSpec.describe Sentry::Rails::Tracing::ActiveRecordSubscriber, :subscriber do
       expect(transaction[:type]).to eq("transaction")
       expect(transaction[:spans].count).to eq(2)
 
-      span = transaction[:spans][1]
-      expect(span[:op]).to eq("db.sql.active_record")
-      expect(span[:description]).to eq("SELECT \"posts\".* FROM \"posts\"")
-      expect(span[:tags]).to include({cached: true})
+      cached_query_span = transaction[:spans][1]
+      expect(cached_query_span[:op]).to eq("db.sql.active_record")
+      expect(cached_query_span[:description]).to eq("SELECT \"posts\".* FROM \"posts\"")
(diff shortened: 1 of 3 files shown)


sentry-rails/spec/sentry/rails/tracing/active_record_subscriber_spec.rb
```diff
@@ -32,6 +32,29 @@
       expect(span[:description]).to eq("SELECT \"posts\".* FROM \"posts\"")
```
st0012, line 32: We should also verify that `cached` is not added by default.

```suggestion
      expect(span[:description]).to eq("SELECT \"posts\".* FROM \"posts\"")
      expect(span[:tags].key?(:cached)).to eq(false)
```

sentry-rails/spec/sentry/rails/tracing/active_record_subscriber_spec.rb
```diff
@@ -32,6 +32,29 @@
...
+      expect(transport.events.count).to eq(1)
+
+      transaction = transport.events.first.to_hash
+      expect(transaction[:type]).to eq("transaction")
+      expect(transaction[:spans].count).to eq(2)
+
+      span = transaction[:spans][1]
+      expect(span[:op]).to eq("db.sql.active_record")
+      expect(span[:description]).to eq("SELECT \"posts\".* FROM \"posts\"")
+      expect(span[:tags]).to include({cached: true})
```
st0012, line 56:
```suggestion
      cached_query_span = transaction[:spans][1]
      expect(cached_query_span[:op]).to eq("db.sql.active_record")
      expect(cached_query_span[:description]).to eq("SELECT \"posts\".* FROM \"posts\"")
      expect(cached_query_span[:tags]).to include({cached: true})
```

st0012 (approved):

st0012:
Thanks for this nice addition 👍

PR #1969, opened 2022-12-25
Refactor workflow matrixes

st0012:
Instead of putting newer Ruby/lib versions into the ever-growing `exclude` list, we should put old Ruby/lib version to the `include` list. This will allow us to just remove those entries when we decide to drop them in the future.

#skip-changelog


sl0thentr0py (approved):
thx for cleaning up!

PR #1972, opened 2022-12-30
Add Span#set_extra and Span#set_extras

aka-nez:
Hi! I'm implementing a feature that tracks performance for our heavy operations with many steps.

This feature allows adding extra payload to performance transactional events.

It can be beneficial to find out bottlenecks with transactions that don't relate to any users or requests, for example.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Features
 
 - Allow [tags](https://docs.sentry.io/platforms/ruby/enriching-events/tags/) to be passed via the context hash when reporting errors using ActiveSupport::ErrorReporter and Sentry::Rails::ErrorSubscriber in `sentry-rails` [#1932](https://github.com/getsentry/sentry-ruby/pull/1932)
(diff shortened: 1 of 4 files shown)


st0012:
Doesn't `set_data` work for this case? It exists exactly for such usages.

aka-nez:
It's a bit different things. Span#add_data puts data inside span info.
CleanShot 2022-12-31 at 09 25 55@2x

Span#add_extra puts data to additional info on whole transactional event and can be easly viewed
CleanShot 2022-12-31 at 09 25 15@2x

st0012:
If you want to add transaction-level data, you can use `Transaction#set_context` for it:

```rb
    transaction = Sentry.get_current_scope&.get_transaction

    if transaction
      transaction.set_context(:user, { name: "bar" })
    end
```

And it'd show up like this (see the `User` section), which IMO is better than using extra:

Screenshot 2022-12-31 at 11 56 26

aka-nez:
Oh. It is exactly what I looking for. Close PR.

PR #1973, opened 2022-12-31
Fix sentry-rails' controller span nesting

### LINKED ISSUE #1899: Span nesting for controller action incorrect
The span nesting is still not 100% correct since the db spans here should be under the controller.
This is because we just subscribe to the rails notifier signals and only get them once stuff finishes and the current span is not updated as code executes.

image

st0012:
As described in #1899, the controller span's nesting is not correct because we record it through ActiveSupport instrumentation.

So this PR I did a few things:

1. Make sure `Span#with_child_span` finishes the span properly when exceptions occurred, so users don't have to add boilerplate code in every call.
    - @sl0thentr0py I probably should make an isolated PR for this. Wdyt?
3. Use `Sentry.with_child_span` around Rails' controller action so spans generated during actions can be attached to the controller span.
4. Deprecate (not remove) the `ActionControllerSubscriber`. This is because tracing subscribers are public classes and removing them should be considered a breaking change.

Fixes #1899

### DIFF
--- a/sentry-rails/lib/sentry/rails/controller_transaction.rb
+++ b/sentry-rails/lib/sentry/rails/controller_transaction.rb
@@ -2,10 +2,34 @@ module Sentry
   module Rails
     module ControllerTransaction
       def self.included(base)
-        base.prepend_before_action do |controller|
-          if Sentry.initialized?
-            Sentry.get_current_scope.set_transaction_name("#{controller.class}##{controller.action_name}", source: :view)
+        base.prepend_around_action(:sentry_around_action)
+      end
+
+      private
+
+      def sentry_around_action
+        if Sentry.initialized?
+          transaction_name = "#{self.class}##{action_name}"
+          Sentry.get_current_scope.set_transaction_name(transaction_name, source: :view)
+          Sentry.with_child_span(op: "view.process_action.action_controller", description: transaction_name) do |child_span|
+            if child_span
+              begin
+                result = yield
+              ensure
+                child_span.set_http_status(response.status)
+                child_span.set_data(:format, request.format)
+                child_span.set_data(:method, request.method)
+                child_span.set_data(:path, request.path)
+                child_span.set_data(:params, request.params)
+              end
+
+              result
+            else
+              yield
+            end
           end
+        else
+          yield
         end
       end
     end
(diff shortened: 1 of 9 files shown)


sentry-rails/lib/sentry/rails/controller_transaction.rb
```diff
@@ -2,10 +2,32 @@ module Sentry
...
+            if child_span
+              yield.tap do
+                child_span.set_http_status(response.status)
+                child_span.set_data(
+                  :payload,
+                  {
+                    format: request.format,
+                    method: request.method,
+                    path: request.path
+                  }
```
sl0thentr0py, line 22: were we also sending `params` before?
https://guides.rubyonrails.org/active_support_instrumentation.html#process-action-action-controller

sl0thentr0py (commented):
just one question

sentry-rails/lib/sentry/rails/controller_transaction.rb
st0012: Added 👍

sentry-rails/lib/sentry/rails/controller_transaction.rb
```diff
@@ -2,10 +2,34 @@ module Sentry
...
+      def sentry_around_action
+        if Sentry.initialized?
+          transaction_name = "#{self.class}##{action_name}"
+          Sentry.get_current_scope.set_transaction_name(transaction_name, source: :view)
+          Sentry.with_child_span(op: "view.process_action.action_controller", description: transaction_name) do |child_span|
+            if child_span
+              begin
+                result = yield
+              ensure
+                child_span.set_http_status(response.status)
```
sl0thentr0py, line 19: should we guard for `response` being nil?

sl0thentr0py (approved):
looks good, see comment and ignore if not needed

sentry-rails/lib/sentry/rails/controller_transaction.rb
st0012: In Rails, the controller's request and response object are all [generated at the beginning of the request](https://github.com/rails/rails/blob/847cc9f8cbdaef59fc43efe78ef74eb4b769fa6c/actionpack/lib/action_controller/metal.rb#L272-L273). So in this case, we don't need to worry about it being `nil` 🙂

Issue #1975, opened 2023-01-04
Link errors and transactions in sentry-opentelemetry


sl0thentr0py:
Use a global event processor to add the correct `contexts.trace` in the opentelemetry `SpanProcessor`.
This is necessary since we have a scope-free implementation of the opentelemetry integration.

PR #1976, opened 2023-01-05
Add Sentry.add_global_event_processor API

### LINKED ISSUE #1974: Add global event processors
Other SDKs allow global event processors without configuring scope, add the same.

sl0thentr0py:
Other SDKs allow adding global event processors without configuring scope, so this adds the same functionality.

closes #1974 
(need it to implement #1975)


st0012 (commented):
Generally looks good 👍  But it'd be better if we don't use class variable for it.

st0012 (approved):

PR #1980, opened 2023-01-07
Avoid using class variable for Configuration's callbacks

st0012:
Class variable has many drawbacks and it's not recommended to use. Although it's not affecting us atm, we should still get rid of it sooner than later, especially given the changes required are few.

More discussions can be seen in https://bugs.ruby-lang.org/issues/18930

#skip-changelog


sl0thentr0py (approved):
thx for the discussion link, learnt something new :)

PR #1982, opened 2023-01-07
`Span#with_child_span` should finish the span even with exception raised

st0012:
This saves users from doing complicated exception handling logic inside their operation block to save the span data, such as

```rb
span.with_child_span do |child_span|
  begin
    do_something
  rescue
    child_span.finish
    raise
  end
end
```

Now users only need:

```rb
span.with_child_span do |child_span|
  do_something
end
```

I consider this a bug fix because I thought it's already working this way 😅


sl0thentr0py (approved):
yep, this matches python behaviour now since there it's a context manager which automatically does the `rescue`

PR #1988, opened 2023-01-21
Do not capture exception in Rails runner at_exit if the exit code is 0 / success

sco11morgan:
## Description
Do not capture exception in Rails runner at_exit if the exit code is 0 / success.

## Why
Recently a change went in to start capturing a Rails runner exceptions in `at_exit` https://github.com/getsentry/sentry-ruby/pull/1820

Unfortunately this captures a Sentry alert if a Rails script finishes with `exit 0`.  

Our application has existing scripts that exit with `exit 0` and we started seeing Sentry alerts when we upgraded.

### DIFF
--- a/sentry-rails/spec/sentry/rails_spec.rb
+++ b/sentry-rails/spec/sentry/rails_spec.rb
@@ -104,7 +104,6 @@ RSpec.describe Sentry::Rails, type: :request do
         pipe_in.close
 
         allow(Sentry::Rails).to receive(:capture_exception) do |event|
-          # raise "hello"
           pipe_out.puts event
         end
 
--- a/sentry-rails/spec/sentry/rails_spec.rb
+++ b/sentry-rails/spec/sentry/rails_spec.rb
@@ -91,6 +91,45 @@ RSpec.describe Sentry::Rails, type: :request do
     end
   end
 
+  context "at exit" do
+    before do
+      make_basic_app
+      Rails.application.load_runner
+    end
+
+    def capture_in_separate_process
+      pipe_in, pipe_out = IO.pipe
+
+      fork do
+        pipe_in.close
+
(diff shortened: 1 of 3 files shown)


sentry-rails/spec/sentry/rails_spec.rb
```diff
@@ -91,6 +91,52 @@
...
+        end
+
+        # silence process
+        $stderr.reopen('/dev/null', 'w')
+        $stdout.reopen('/dev/null', 'w')
+
+        exit exit_code
+      end
+
+      pipe_out.close
```
sl0thentr0py, line 118: @st0012 do you have opinions on this test, seems a bit like too much extra work

sco11morgan: Note this same test pattern was [copied from sentry-raven](https://github.com/getsentry/sentry-ruby/blob/94c28fe9a7fd1c15a10eeeb0e1c4fcac099d4166/sentry-raven/spec/raven/raven_spec.rb#L17-L41)

sentry-rails/spec/sentry/rails_spec.rb
```diff
@@ -91,6 +91,52 @@
...
+    end
+
+    def capture_in_separate_process(exit_code:)
+      pipe_in, pipe_out = IO.pipe
+
+      fork do
+        pipe_in.close
+
+        allow(Sentry::Rails).to receive(:capture_exception) do |event|
+          # raise "hello"
```
sco11morgan, line 107:
```suggestion
```

st0012 (approved):
Thanks for the fix 👍 
I think the test complexity is going to be regardless of the approach when testing `at_exit` logic end-to-end. So I'm fine leaving it like this.

sl0thentr0py (approved):

PR #1989, opened 2023-01-23
Add before_send_transaction hook

### LINKED ISSUE #1937: Add beforeSendTransaction
Problem Statement
Similar to beforeSend, we will add beforeSendTransaction to allow people to strip Pii or drop transactions before they are sent to Sentry.

Solution Brainstorm
https://github.com/getsentry/sentry-javascript/pull/6121

sl0thentr0py:
resolves #1937

### DIFF
--- a/sentry-ruby/spec/sentry/client/event_sending_spec.rb
+++ b/sentry-ruby/spec/sentry/client/event_sending_spec.rb
@@ -224,6 +224,15 @@ RSpec.describe Sentry::Client do
           expect(event["tags"]["called"]).to eq(true)
         end
       end
+
+      it "doesn't apply before_send_transaction to Event" do
+        dbl = double("before_send_transaction")
+        allow(dbl).to receive(:call)
+        configuration.before_send_transaction = dbl
+
+        expect(dbl).not_to receive(:call)
+        subject.send_event(event)
+      end
     end
 
     it_behaves_like "Event in send_event" do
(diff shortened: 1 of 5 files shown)


sentry-ruby/spec/sentry/client/event_sending_spec.rb
```diff
@@ -224,6 +224,15 @@
...
         end
       end
+
+      it "doesn't apply before_send_transaction to Event" do
+        dbl = double("before_send_transaction")
+        allow(dbl).to receive(:call)
+        configuration.before_send_transaction = dbl
+
+        expect(dbl).not_to receive(:call)
+        subject.send_event(event)
```
st0012, line 234: Can we avoid using double/mocking when it's not too hard to test it end-to-end? In my experience taking over the old SDK it makes things hard to maintain in the long-term.

sentry-ruby/spec/sentry/client/event_sending_spec.rb
```diff
@@ -246,6 +255,26 @@
 
         subject.send_event(event)
       end
+
+      it "applies before_send_transaction callback before sending the event" do
+        configuration.before_send_transaction = lambda do |event, _hint|
+          if event.is_a?(Sentry::Event)
```
st0012, line 261: I think we can be more specific to use `Sentry::TransactionEvent` here and below?

st0012 (commented):
Just small comments on test.

sentry-ruby/spec/sentry/client/event_sending_spec.rb
sl0thentr0py: I did this because of the new codecov enforcing (which was turned on org-wide on github), it was complaining about a test line not being run. I will generally not use doubles, but in this case it makes sense.

antonpirker (approved):
Now I know what virtual attributes and doubles are in Ruby

PR #1995, opened 2023-02-05
Add minimum VSCode setup for the project

st0012:
These settings are mostly for my convenience using VS Code, which should have no impact on non-VS Code developers. However, if you do use VS Code when working on this/other Ruby project, I think the 2 extensions are worth installing 😉

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,6 @@
+{
+    "recommendations": [
+        "shopify.ruby-lsp",
+        "koichisasada.vscode-rdbg"
+    ]
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,8 @@
+{
+  "[ruby]": {
+    "editor.tabSize": 2,
+    "editor.insertSpaces": true,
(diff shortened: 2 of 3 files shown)


st0012:
@sl0thentr0py I'm curious if you:

- Use VS Code too?
- Have preference on Ruby formatting tools and rules?
    - I want to have auto-formatting via editors. But we first need to format the current codebase with consistent rules first, or we'll introduce a lot of unrelated formatting changes in PRs.

sl0thentr0py:
no i just use neovim.
I'm fine with whichever tools you pick. :)
We can do a one time formatting PR to update the whole codebase and then enforce via CI/precommit.
Are you thinking of rubocop or something else?

sl0thentr0py (approved):

st0012:
I personally don't want to configure any rubocop rules. So I'd go with [`syntax_tree`](https://github.com/ruby-syntax-tree/syntax_tree), which only has one opinionated rule and is `ruby-lsp`'s dependency + default formatter.

sl0thentr0py:
oki, also fine with me!

PR #1997, opened 2023-02-06
Ignore redis key if not UTF8

### LINKED ISSUE #1992: Transaction sending failed: "\xD9" from ASCII-8BIT to UTF-8
### Issue Description

(Similar to #1911)

Hello,

I got the error `Transaction sending failed: "\xD9" from ASCII-8BIT to UTF-8` that's raised from this method:

https://github.com/getsentry/sentry-ruby/blob/193446b39dfa0d9d6b484631344fdc5aeaa75db8/sentry-ruby/lib/sentry/envelope.rb#L12-L15

Here is the log:

Screenshot 2023-02-02 at 22 40 55

We're using [MessagePack](https://msgpack.org/index.html) to serialize data before storing them into Redis. The encoding of the string returned by MessagePack is `ASCII_8BIT`. This string is captured in a Sentry span description, and the serialization into JSON fails due to the invalid encoding:

Screenshot 2023-02-02 at 22 54 31

We try to fix this issue using the `before_send` hook, but unfortunately we've not been able to find spans in the event object.

We finally fixed this issue by monkey patching the `set_description` method of the `Sentry::Span` class:

Screenshot 2023-02-02 at 23 14 30

Is this a known issue that should be fixed in Sentry SDK directly? If not, is there a way to deal with non UTF-8 string using the `before_send` hook?

### Reproduction Steps

In a Rails project with Sentry and MessagePack installed:

`Sentry.capture_message(MessagePack.pack('test message'))`

### Expected Behavior

Event is sent to Sentry.

### Actual Behavior

Event isn't sent to Sentry.

### Ruby Version

2.7.4

### SDK Version

5.7.0

rails (6.0.5.1), sentry-rails (5.7.0), sidekiq (5.2.2), sentry-sidekiq (5.7.0)

### Sentry Config

``` ruby
# frozen_string_literal: true

Sentry.init do |config|
  config.dsn = ENV['SENTRY_DSN']

  # Tracing Options
end
```

sl0thentr0py:
fixes #1992

### DIFF
--- a/sentry-ruby/lib/sentry/redis.rb
+++ b/sentry-ruby/lib/sentry/redis.rb
@@ -54,7 +54,8 @@ module Sentry
     def parsed_commands
       commands.map do |statement|
         command, key, *arguments = statement
-        command_set = { command: command.to_s.upcase, key: key }
+        command_set = { command: command.to_s.upcase }
+        command_set[:key] = key if Utils::EncodingHelper.valid_utf_8?(key)
 
         if Sentry.configuration.send_default_pii
           command_set[:arguments] = arguments
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,8 @@
 - Fix sentry-rails' controller span nesting [#1973](https://github.com/getsentry/sentry-ruby/pull/1973)
   - Fixes [#1899](https://github.com/getsentry/sentry-ruby/issues/1899)
 - Do not report exceptions when a Rails runner exits with `exit 0` [#1988](https://github.com/getsentry/sentry-ruby/pull/1988)
+- Ignore redis key if not UTF8 [#1997](https://github.com/getsentry/sentry-ruby/pull/1997)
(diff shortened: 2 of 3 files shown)


sentry-ruby/lib/sentry/redis.rb
```diff
@@ -54,7 +54,8 @@ def commands_description
     def parsed_commands
       commands.map do |statement|
         command, key, *arguments = statement
-        command_set = { command: command.to_s.upcase, key: key }
+        command_set = { command: command.to_s.upcase }
+        command_set[:key] = key if Utils::EncodingHelper.valid_utf_8?(key)
```
st0012, line 58: Since `key` is kinda important in commands, do you think it'd be better insert a placeholder like `"<invalid encoded key>"` instead?

sl0thentr0py: I think the type of redis command signifies that there's a key there, I think that's fine. 
This is edge-case behaviour anyway.

st0012 (approved):

PR #2001, opened 2023-02-09
Fix comment on Raven::Configuration#encoding

dbackeus:
The encoding= method accepts a `String`, not a `Symbol`: https://github.com/getsentry/sentry-ruby/blob/master/sentry-raven/lib/raven/configuration.rb#L312


sl0thentr0py:
thx @dbackeus but `sentry-raven` is deprecated and we no longer do releases for it.
Please upgrade to `sentry-ruby`, there's a [nice migration guide](https://docs.sentry.io/platforms/ruby/migration/#api-changes). :)

PR #2005, opened 2023-02-15
Add config.enable_tracing for easier performance tracing setup

### LINKED ISSUE #1996: Add enableTracing Option
A new simplified options for configuring to turn tracing & performance on should be added to all SDKs. For each SDK the default sample rate should be agreed upon.  

Docs do not need to be updated yet, nor the onboarding wizard. This can wait until the top level options are added to a number of SDKs and we decide if we want to configure the default onboarding experience, for all new users.

See [develop docs](https://develop.sentry.dev/sdk/performance/#enabletracing) for more information. This is a non-breaking change to be clear. Any of the three options should enable performance, alone or in combination

### Open point:
recommend default it 1.0 or 100%. Can discuss if it should be less.

sl0thentr0py:
The new optional boolean will set traces_sample_rate to 1.0
automatically if not set yet. If false, this will also disable tracing
even if a sample rate or sampler is set.

closes #1996

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## Unreleased
+
+### Features
+
+- Add new boolean option `config.enable_tracing` to simplify enabling performance tracing [#2005](https://github.com/getsentry/sentry-ruby/pull/2005)
+  - `config.enable_tracing = true` will set `traces_sample_rate` to `1.0` if not set already
+  - `config.enable_tracing = false` will turn off tracing even if `traces_sample_rate/traces_sampler` is set
+  - `config.enable_tracing = nil` (default) will keep the current behaviour
+
 ## 5.8.0
 
 ### Features
--- a/sentry-ruby/lib/sentry/configuration.rb
+++ b/sentry-ruby/lib/sentry/configuration.rb
@@ -227,6 +227,11 @@ module Sentry
     # @return [Proc]
     attr_accessor :traces_sampler
 
+    # Easier way to use performance tracing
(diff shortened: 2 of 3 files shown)


antonpirker (approved):
The only thing I do not get is what "sending not allowed" means, but otherwise it looks good.

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## Unreleased
+
+### Features
+
+- [Add](Add) new boolean option `config.enable_tracing` to simplify enabling performance tracing [#2005](https://github.com/getsentry/sentry-ruby/pull/2005)
```
st0012, line 5: Why does the `Add` need a link?

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
...
+  - `config.enable_tracing = true` will set `traces_sample_rate` to `1.0` if not set already
+  - `config.enable_tracing = false` will turn off tracing even if `traces_sample_rate/traces_sampler` is set
+  - `config.enable_tracing = nil` will keep the current behaviour
```
st0012, line 8: Should we indicate that this is the default?

st0012 (approved):
Just some suggestions on the changelog

sl0thentr0py:
master rails is broken :( I will fix later

st0012:
@sl0thentr0py After checking the error, it seems to be caused by Rails itself, see: https://github.com/rails/rails/issues/47377. And there is already a PR for it: https://github.com/rails/rails/pull/47409

So IMO we can wait a couple of days for the issue to be resolved upstream.

Issue #2006, opened 2023-02-16
Allow `capture_exception` to manually bypass configured `excluded_exceptions`


ty-porter:
## **Describe the idea**
Currently, sentry-ruby filters common controller exceptions out of the box via an array of default `excluded_exceptions` in the configuration. Most of the time, this does a great job cutting down noisy exceptions caused by common user errors like 404s.

However, there's no way to avoid filtering exceptions, and that means they are ALWAYS dropped, even when they are potentially useful information (specifically `ActiveRecord::RecordNotFound`, which I know comes from sentry-rails but the implementation for exclusion occurs in sentry-ruby).

Allow users to manually override the filter, perhaps as such:

```ruby
Sentry.capture_exception(exception, ignore_excluded: true)
```

## **Why do you think it's beneficial to most of the users**
As far as I can tell, other Sentry SDKS for other languages and frameworks do not implement exclusions as aggressively.

Sometimes, it's possible for the default exclusions to contain useful errors, perhaps in a complicated service where the failure might indicate a problem with application code and not necessarily user error.

```ruby
# Will throw ActiveRecord::RecordNotFound
# but this indicates a problem with application code NOT user error (404)!
def always_fails
  Thing.find(-1)
rescue StandardError => exception
  Sentry.capture_exception(exception)
end
```

## **Possible implementation**
#2007

st0012:
I think this is a reasonable request. @sl0thentr0py do other SDKs have mechanisms for this?

sl0thentr0py:
there's no official exposed argument in any other SDK and I wouldn't touch the function signature but we can just hack it into `hint` like we do with `background`? The use case is valid.

ty-porter:
For more context my PR is based around similar changes from #1550

I don't have strong feelings either way, I don't personally use the `hint` arg so if it makes more sense there I'm happy to change

st0012:
I think using `hint` is a good idea. @ty-porter would you mind changing your implementation to use it? Thx

st0012:
Implemented in #2007

PR #2007, opened 2023-02-16
Allow ignoring excluded_exceptions when manually capturing exceptions

ty-porter:
## Description
Addresses #2006 and allows manually bypassing `excluded_exceptions` array from Sentry config.


sentry-ruby/spec/sentry/hub_spec.rb
```diff
@@ -195,6 +195,11 @@
       end.to change { transport.events.count }.by(1)
     end
 
+    it "takes ignore_exclusions option" do
```
st0012, line 198: I prefer having an end-to-end test here.

st0012 (approved):
Sorry for the delay. LGTM 👍

PR #2014, opened 2023-03-04
Fix CI

st0012:
There are several issues causing the CI to fail:

### Fix sentry-rails' test failures caused by Zeitwerk exception

1. By default, Rails eager loads the app through Zeitwerk and registers
    future eager load calls with the `after_class_unload` hook. This hook
    makes sure the app would be eagerly reloaded after the app is unloaded.
    ([code](https://github.com/rails/rails/blob/64962266e38817e025de53234accc1e54ed2b32e/railties/lib/rails/application/finisher.rb#L77-L81))

2. Since version 2.6.5, Zeitwerk raises an error when the app is being
    eager loaded without a proper setup.
3. Our test app is not properly setup for being reloaded. So when the app
    is unloaded, it's not ready for the next eager load call, which means
    it'd trigger the error mentioned in 2.

So this PR simply marks the app as not reloadable, which means it doesn't register the callback mentioned in 1 and avoids the later errors.

### Fix sentry-resque's test failure caused by `mono_logger`

It's caused by `mono_logger` (a resque dependency)'s conflict with Ruby 3.3's `Logger`. I've opened an [issue](https://github.com/steveklabnik/mono_logger/issues/13) and worked around it by changing Resque's logger in our tests.

### Fix sentry-ruby's test failure caused by error message change

#skip-changelog

### DIFF
--- a/.github/workflows/sentry_rails_test.yml
+++ b/.github/workflows/sentry_rails_test.yml
@@ -20,6 +20,7 @@ jobs:
     name: Ruby ${{ matrix.ruby_version }} & Rails ${{ matrix.rails_version }}, options - ${{ toJson(matrix.options) }}
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         rails_version: [6.1.0, 7.0.0, 7.1.0]
         ruby_version: [2.7, '3.0', '3.1', '3.2', head]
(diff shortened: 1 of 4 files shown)


st0012:
@sl0thentr0py There still are some test failures to address. I'll try to fix everything this week.

sl0thentr0py (approved):

PR #2023, opened 2023-03-20
Tweak #2007 and add changelog for it

st0012:
When writing a changelog entry for #2007, I found that the `ignore_exclusions` was added as a top-level option when using `Sentry.capture_exception`, which is not what was advised in #2006. I believe this is due to minor misunderstanding so I corrected it here.


st0012:
@sl0thentr0py Do you know how we should present such PRs in changelog 🤔

sl0thentr0py:
ah we can just ignore danger for this, it's fine

sl0thentr0py (approved):

PR #2024, opened 2023-03-21
Add stackprof based profiler

### LINKED ISSUE #2013: Profiling for Ruby
We want to look into extending our Profiling product with support for Ruby.

On other platforms, we used an existing profiler as a starting point or completely relied on external packages.

When a txn starts, a new `profiling_sample_rate`, which works in relation to `traces_sample_rate`, determines if this txn should be profiled. Our default sample rate on other platforms is 101Hz.

On finish, we convert the raw output to our own profiling format and attach it as a profile envelope item to the transaction.

https://github.com/rbspy/rbspy might be worth looking into, as it's a sampling profiler that also supports outputting a speedscope compatible format, something our own format is derived from.

sl0thentr0py:
resolves #2013 

### Usage

* Add `stackprof` to `Gemfile` as a peer dependency

```ruby
Sentry.init do |config|
  config.profiles_sample_rate = 1.0
  config.release = "release"
end
```


sl0thentr0py:
hey @st0012 @Zylphrex I believe this is shippable now, I'll write up a PR description and documentation etc next week, feel free to take a look and start reviewing!

sentry-ruby/lib/sentry/hub.rb
```diff
@@ -90,6 +90,12 @@ def start_transaction(transaction: nil, custom_sampling_context: {}, instrumente
       sampling_context.merge!(custom_sampling_context)
 
       transaction.set_initial_sample_decision(sampling_context: sampling_context)
+
+      if transaction.profiler
+        transaction.profiler.set_initial_sample_decision(transaction.sampled)
+        transaction.profiler.start
+      end
```
st0012, line 97: I feel like this could be a `Transaction#start_profiler!` method?

sentry-ruby/lib/sentry/transaction_event.rb
```diff
@@ -49,5 +54,33 @@ def to_hash
...
+
+    def populate_profile(transaction)
+      return nil unless transaction.profiler
+
+      profile_hash = transaction.profiler.to_hash
+      return nil unless profile_hash
+
+      profile_hash.merge!(
+        environment: environment,
+        release: release,
```
st0012, line 68: IIRC, information like environment or release are also in `context`. Can't Sentry get them from the associated transaction's context?

sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -395,6 +401,18 @@ def enable_tracing=(enable_tracing)
       @traces_sample_rate ||= 1.0 if enable_tracing
     end
 
+    def profiles_sample_rate=(profiles_sample_rate)
+      log_info("Thank you for trying out Profiling on Sentry. Please note that this is currently an experimental feature.")
```
st0012, line 405: I feel that noting this is experimental in docs and changelog should be enough. If I were an user that opt in to this voluntarily, these information wouldn't be helpful.

We should keep the `stackprof` inclusion warning though.

sentry-ruby/lib/sentry/profiler.rb
```diff
@@ -0,0 +1,232 @@
...
+
+      idx = 0
+      elapsed_since_start_ns = 0
+      samples = []
+
+      num_seen.each_with_index do |n, i|
+        n.times do
+          # stackprof deltas are in microseconds
+          delta = results[:raw_timestamp_deltas][idx]
+          elapsed_since_start_ns += (delta * 1e3).to_i
```
st0012, line 137: I feel this `1e3` could also be a documented constant?

sentry-ruby/lib/sentry/profiler.rb
```diff
@@ -0,0 +1,232 @@
...
+        return
+      else
+        @sampled = Random.rand < profiles_sample_rate
+      end
+
+      log('Discarding profile due to sampling decision') unless @sampled
+    end
+
+    def to_hash
+      return nil unless Sentry.initialized?
```
st0012, line 78: In other parts of the SDK, `to_hash` always return `Hash`. Can we return empty hash here instead? I'd be surprising to see `to_hash` return `nil` IMO.

sentry-ruby/lib/sentry/profiler.rb
```diff
@@ -0,0 +1,232 @@
...
+
+      frame_map = {}
+
+      frames = results[:frames].to_enum.with_index.map do |frame, idx|
+        frame_id, frame_data = frame
+
+        # need to map over stackprof frame ids to ours
+        frame_map[frame_id] = idx
+
+        in_app = in_app?(frame_data[:file])
```
st0012, line 96: Can we declare `frame_data[:file]` as a local variable? Likely `file_path`?

sentry-ruby/lib/sentry/profiler.rb
```diff
@@ -0,0 +1,232 @@
...
+        function, mod = split_module(frame_data[:name])
+
+        {
+          abs_path: frame_data[:file],
+          function: function,
+          module: mod,
+          lineno: frame_data[:line],
+          filename: filename,
+          in_app: in_app
+        }.compact
```
st0012, line 107: TIL `Hash#compact` 👍 
But if looks like the only value could be `nil` here is `frame_data[:line]`? How often does that happen? If it's not a common case, we should probably change this to:


```rb
        hash = {
          abs_path: frame_data[:file],
          function: function,
          module: mod,
          filename: filename,
          in_app: in_app
        }
        hash[:lineno] = frame_data[:line] if frame_data[:line]
        hash
```

Otherwise, `Hash#compact` always allocate a new hash

```
irb(main):006:0> h = { a: 1 }; puts h.object_id; puts h.compact.object_id
291420
291440                                                                                                                      
=> nil    
```

which is probably not much, but IMO could be avoided here.

sentry-ruby/lib/sentry/profiler.rb
```diff
@@ -0,0 +1,232 @@
...
+    end
+
+    private
+
+    def log(message)
+      Sentry.logger.debug(LOGGER_PROGNAME) { "[Profiler] #{message}" }
+    end
+
+    def project_root
+      @project_root ||= Sentry.configuration.project_root
```
st0012, line 185: We should be able to define this in the constructor?

sentry-ruby/lib/sentry/profiler.rb
```diff
@@ -0,0 +1,232 @@
...
+    end
+
+    def in_app_pattern
+      @in_app_pattern ||= begin
```
st0012, line 189: Same as the above, we should be able to define this in the constructor?

sentry-ruby/lib/sentry/profiler.rb
sl0thentr0py: this is safeguarded by the `Sentry.initialized?` check above otherwise `Sentry.configuration` could be nil. I don't know of a clean way to do this, in `Transaction` for instance, we pass in `hub` manually but that's marked as deprecated and I also don't want to pass in `hub` here.

st0012: If the SDK is not initialised, it probably won't even create any Transaction (otherwise [these would error](https://github.com/getsentry/sentry-ruby/blob/master/sentry-ruby/lib/sentry/transaction.rb#L76-L82)), which means no `Profiler.new` will be called anyway?

sl0thentr0py: yea ok true, then I'll move them up, thanks!

sentry-ruby/lib/sentry/profiler.rb
sl0thentr0py: not sure I follow, this is for each `frame`, what would the local variable be?

sentry-ruby/lib/sentry/profiler.rb
sl0thentr0py: good point, `module` can also be `nil`, but I will add both optionally.

sentry-ruby/lib/sentry/profiler.rb
sl0thentr0py: this causes a bunch of tests to fail because we don't always initialized sentry in tests, for instance in `hub_spec`...

sl0thentr0py: so either we pass in `hub` to `Profiler.new` like with `Transaction` or we need to guard with `Sentry.initialized?` checks like I did. This is more a testing problem but I don't want to change the way we test these things for this PR.

Which do you prefer?

sentry-ruby/lib/sentry/profiler.rb
```diff
@@ -0,0 +1,228 @@
...
+
+        # our call graph is reversed
+        stack = results[:raw].slice(idx, len).map { |id| frame_map[id] }.compact.reverse
+        stacks << stack
+
+        num_seen << results[:raw][idx + len]
+        idx += len + 1
+
+        log('Unknown frame in stack') if stack.size != len
+      end
```
Zylphrex, line 134: So if I'm understanding this correctly if the profile looks like

```
at t = 0 stack is a -> b
at t = 1 stack is a -> b -> c
at t = 2 stack is a -> b
```

the `stacks` lists will contain `a -> b` twice since it only merges identical stacks that are adjacent to each other?

This could potentially result in a massive payload from the duplicate stacks for certain types of profiles, but not a blocker for this initial release.

sentry-ruby/lib/sentry/transaction_event.rb
st0012: I see, thanks for the context 👍

sentry-ruby/lib/sentry/profiler.rb
st0012: I actually prefer not calling `Sentry.configuration` at all from the SDK's key components due to this reason. For example, if we call it inside `Transaction`, then we'll need to initialise the entire SDK to test it, which makes the test hard to maintain and make tests couple.

Since `Transaction` current holds `@configuration`, which was planned to be removed. How about we don't remove it anymore and pass it to `Profiler.new`? This will make testing `Profiler` easier as well.

sentry-ruby/lib/sentry/profiler.rb
```diff
@@ -0,0 +1,228 @@
...
+
+    def start
+      return unless @sampled
+
+      @started = StackProf.start(interval: DEFAULT_INTERVAL,
+                                 mode: :wall,
+                                 raw: true,
+                                 aggregate: false)
+
+      @started ? log('Started') : log('Not started since running elsewhere')
```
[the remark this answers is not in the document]
sl0thentr0py: yep! if `StackProf` is already running, the `start` method will return false.

st0012 (commented):
If we only look at the code, I think there are many questions to ask and keep in mind when we check if a profiler would actually be sent:

1. Is `Profiler` defined?
2. Does `Transaction#profiler` return `nil`?
3. Does `Profiler#start` start?
4. Does `Profiler#to_hash` return `nil` or a Hash?

I feel we can improve it by:
- Always define `Profiler` and initialises one in `Transaction#initialize`. So `transaction.profiler` is never `nil`.
- Make Profiler methods no-op if `StackProf` is not defined. In another words, let's hide all the `StackProf` complexity all inside `Profiler`.
- Always return a Hash in `Profiler#to_hash` (empty one if it's not sampled)

WDYT?

sentry-ruby/lib/sentry/profiler.rb
```diff
@@ -0,0 +1,228 @@
...
+            # https://bugs.ruby-lang.org/issues/10602
+            thread_id: '0',
+            elapsed_since_start_ns: elapsed_since_start_ns.to_s
+          }
+        end
+      end
+
+      log('Some samples thrown away') if samples.size != results[:samples]
+
+      if samples.size <= 2
```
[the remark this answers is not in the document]
sl0thentr0py: I'm throwing away all the small deltas above
https://github.com/getsentry/sentry-ruby/blob/b0e92692c9c451da6fc7da6810f5fd2a0f2a74b5/sentry-ruby/lib/sentry/profiler.rb#L150

sentry-ruby/lib/sentry/profiler.rb
st0012: I mean there are 3 occurrences of `frame_data[:file]`, maybe we can extract it to a local variable?

https://github.com/getsentry/sentry-ruby/blob/b0e92692c9c451da6fc7da6810f5fd2a0f2a74b5/sentry-ruby/lib/sentry/profiler.rb#L100-L105

Sorry if it feels like a nitpick 😅

sentry-ruby/lib/sentry/profiler.rb
sl0thentr0py: yes, this only optimizes adjacent samples, logic is below
https://github.com/tmm1/stackprof/blob/52d1df6502eaede5061a2b5d28026e982bcf262a/ext/stackprof/stackprof.c#L513-L528

If we write our own extension later, I'll keep this in mind to also optimize.

sentry-ruby/lib/sentry/profiler.rb
```diff
@@ -0,0 +1,228 @@
...
+
+      results = StackProf.results
+      return nil unless results
+      return nil if results.empty?
+      return nil if results[:samples] == 0
+      return nil unless results[:raw]
+
+      frame_map = {}
+
+      frames = results[:frames].to_enum.with_index.map do |frame, idx|
```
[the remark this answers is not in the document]
sl0thentr0py: afaict yes, there's some hashing logic here and the `results[:frames]` object is a hash from some `frame_ids` to the actual `frame_data`.

sentry-ruby/lib/sentry/profiler.rb
sl0thentr0py: ah sorry local var, I read instance variable in my head sorry. Sometimes I can't read properly >.<

sl0thentr0py:
thanks @st0012 that actually makes sense, this was just the state I ended up in with all my hacking around, but what you describe is indeed cleaner!

sl0thentr0py:
alright I believe I covered all the comments, @st0012

st0012 (commented):
IMO using non-object pattern here is an overkill. 

Give that:

1. All public profiler APIs no-op if not `@sampled`
2. `@sampled` will never be `true` unless `@profiling_enabled`

Can we do this instead?

1. Define a single `Profiler` class
2. In `Profiler`'s constructor, we use:

```rb
@profiling_enabled = defined?(StackProf) && configuration.profiling_enabled?
```

Alternatively, I think just early returning the `initialize` method may also work.

sl0thentr0py:
ok I did one more round of sanity check testing, looks ok. I'm merging!

PR #2026, opened 2023-03-26
Support capturing low-level errors propagated to Puma

### LINKED ISSUE #1511: Capture exceptions that happen outside of Rails middleware
**Describe the idea**

Suppose you have a Rails middleware that does this:

```ruby
def call(env)
  return [200, { 'Content-Type' => 'text/plain' }, 'hi']
end
```

Since `'hi'` does not respond to `each` as specified in the Rack spec (https://github.com/rack/rack/blob/master/SPEC.rdoc#label-The+Body), this fails quietly in the Web server with this message:

```
#<NoMethodError: undefined method `each' for "hi":String>
```

Other users have reported this in https://github.com/puma/puma/issues/1782.

While this problem is clearly a bug in the middleware, we've come across it enough times that sometimes it takes a long to troubleshoot. The backtrace is not always available, and the log messages are hidden away in an unstructured `stderr` output. It's surprising that this exception isn't reported via Sentry. 

**Possible implementation**

Capturing this exception might depend on integration with the Web server in question. For example, if you use Thin, the exception is encountered here: https://github.com/macournoyer/thin/blob/0712d603a31d97b9fa8a0260da400da2e4217d60/lib/thin/connection.rb#L114-L123

For Puma, there is this low-level error handler: https://github.com/puma/puma#error-handling. Perhaps it makes sense to hook Sentry up to the Puma `ErrorLogger`? https://github.com/puma/puma/blob/v5.3.2/lib/puma/error_logger.rb

Note that the exception is:

1. Hit on this line: https://github.com/puma/puma/blob/bda19f8225a36f83bb2671fc4baef870df9f76b4/lib/puma/request.rb#L149. 
2. Rescued here: https://github.com/puma/puma/blob/f4766ce46976ec623f163a0428515d157f53e420/lib/puma/server.rb#L476
3. Reported here: https://github.com/puma/puma/blob/f4766ce46976ec623f163a0428515d157f53e420/lib/puma/server.rb#L517
4. Logged here: https://github.com/puma/puma/blob/f4766ce46976ec623f163a0428515d157f53e420/lib/puma/events.rb#L123

The backtrace shows:

```
lib/ruby/gems/2.7.0/gems/rack-2.2.3/lib/rack/body_proxy.rb:41:in `method_missing'
lib/ruby/gems/2.7.0/gems/rack-2.2.3/lib/rack/body_proxy.rb:41:in `method_missing'
lib/ruby/gems/2.7.0/gems/rack-2.2.3/lib/rack/body_proxy.rb:41:in `method_missing'
lib/ruby/gems/2.7.0/gems/rack-2.2.3/lib/rack/body_proxy.rb:41:in `method_missing'
lib/ruby/gems/2.7.0/gems/rack-2.2.3/lib/rack/body_proxy.rb:41:in `method_missing'
lib/ruby/gems/2.7.0/gems/puma-5.3.2/lib/puma/request.rb:149:in `handle_request'
lib/ruby/gems/2.7.0/gems/puma-5.3.2/lib/puma/server.rb:438:in `process_client'
lib/ruby/gems/2.7.0/gems/puma-5.3.2/lib/puma/thread_pool.rb:145:in `block in spawn_thread'
```

Similar issues:

1. https://github.com/getsentry/sentry-ruby/issues/917.
1. https://forum.sentry.io/t/puma-exception-handling/979.

st0012:
This PR implements Puma's error reporting by patching `Puma::Server#lowlevel_error`.

I thought about using the `lowlevel_error_handler` API to add a Sentry error handler, but it's not possible to do that after `puma.rb` is loaded (which is extremely early in a normal app setup, way before the SDK is loaded). There is also no good ways to fetch and update the server's `lowlevel_error_handler` option. So the only option left is to patch the method.

Notes:

- The `lowlevel_error` patch is compatible with [Puma 3.0+](https://github.com/puma/puma/blob/v3.0.0/lib/puma/server.rb#L808) (current latest is 6.1). So it's pretty stable and I'm not sure if it'd be worth introducing another test matrix for Puma versions.
- I think for patches like this it's better to NOT require the related library in `spec_helper` as the nonappearance of the library should also be tested. This is why I added the spec under `isolated` folder.
    - I plan to apply this test pattern to the redis patch in another PR.
- ~~I also want to improve the `Sentry.register_patch` API as we now have 3 patches using it and I've spotted some common boilerplate code.~~

Closes #1511

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
   Sentry.capture_exception(ignored_exception) # won't be sent to Sentry
   Sentry.capture_exception(ignored_exception, hint: { ignore_exclusions: true }) # will be sent to Sentry
   ```
+- Support capturing low-level errors propagated to Puma [#2026](https://github.com/getsentry/sentry-ruby/pull/2026)
 
 - Add `spec` to `Backtrace::APP_DIRS_PATTERN` [#2029](https://github.com/getsentry/sentry-ruby/pull/2029)
 - Forward all `baggage` header items that are prefixed with `sentry-` [#2025](https://github.com/getsentry/sentry-ruby/pull/2025)
--- a/sentry-ruby/Gemfile
+++ b/sentry-ruby/Gemfile
@@ -10,6 +10,8 @@ gem "rack", "~> #{Gem::Version.new(rack_version)}" unless rack_version == "0"
 redis_rb_version = ENV.fetch("REDIS_RB_VERSION", "5.0")
 gem "redis", "~> #{redis_rb_version}"
 
+gem "puma"
+
 gem "rake", "~> 12.0"
 gem "rspec", "~> 3.0"
 gem "rspec-retry"
--- a/sentry-ruby/Rakefile
+++ b/sentry-ruby/Rakefile
@@ -8,6 +8,13 @@ require "rspec/core/rake_task"
 
 RSpec::Core::RakeTask.new(:spec).tap do |task|
   task.rspec_opts = "--order rand"
+  task.exclude_pattern = "spec/isolated/**/*_spec.rb"
 end
 
-task :default => :spec
+task :isolated_specs do
+  Dir["spec/isolated/*"].each do |file|
+    sh "bundle exec rspec #{file}"
+  end
+end
+
+task :default => [:spec, :isolated_specs]
--- a/sentry-ruby/lib/sentry-ruby.rb
+++ b/sentry-ruby/lib/sentry-ruby.rb
@@ -519,3 +519,4 @@ end
 # patches
 require "sentry/net/http"
 require "sentry/redis"
+require "sentry/puma"
--- a/sentry-ruby/lib/sentry/puma.rb
+++ b/sentry-ruby/lib/sentry/puma.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module Sentry
+  module Puma
+    module Server
+      def lowlevel_error(e, env, status=500)
+        result = super
+
+        begin
+          Sentry.capture_exception(e) do |scope|
+            scope.set_rack_env(env)
+          end
+        rescue
+          # if anything happens, we don't want to break the app
+        end
+
+        result
+      end
+    end
+  end
+end
+
+if defined?(Puma::Server)
+  Sentry.register_patch(Sentry::Puma::Server, Puma::Server)
+end
--- a/sentry-ruby/spec/isolated/puma_spec.rb
+++ b/sentry-ruby/spec/isolated/puma_spec.rb
@@ -1,0 +1,91 @@
+require "puma"
+require_relative "../spec_helper"
+
+# Because puma doesn't have any dependency, if Rack is not installed the entire test won't work
+return if ENV["RACK_VERSION"] == "0"
+
+RSpec.describe Puma::Server do
+  class TestServer
+    def initialize(app, options)
+      @host = "127.0.0.1"
+      @ios = []
+      @server = Puma::Server.new(app, nil, options)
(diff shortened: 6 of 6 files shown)


sl0thentr0py (approved):

dentarg:
@st0012 @sl0thentr0py Is it possible to opt-out from this?

This makes Sentry by default catch the generally non-actionable `Puma::HttpParserError` exception :/

dentarg:
How does this behave if you already use Sentry in the `lowlevel_error_handler`?

st0012:
> How does this behave if you already use Sentry in the lowlevel_error_handler?

I think it'll send 2 events in this case.

> This makes Sentry by default catch the generally non-actionable Puma::HttpParserError exception

Sorry that I wasn't aware such exceptions would be captured too. Does adding it to `config.excluded_exceptions` help? If it does, I will add it to the default exclusion list.

dentarg:
> So I'd advise removing the custom lowlevel_error_handler

that's is not a great solution, that handler is also used to define the HTTP response you want you application to have 

> Does adding it to config.excluded_exceptions help?

I assume so (we have done that now) but I haven't verified it

dentarg:
Can you consider raising a feature request on the Puma repository to facilitate doing this in a less intrusive way? (You could contribute the feature if we reach some good idea how to do it)

st0012:
> that's is not a great solution, that handler is also used to define the HTTP response you want you application to have

Sorry I'm not saying not using that handler, just not sending events to Sentry from it now that we have this integration.

shouichi:
We started to see `Puma::HttpParserError` after puma v6.2x. Can you consider a proper solution without monkey patching?

st0012:
Apologies for the delayed response.

I believe the core issue here is that I failed to implement proper exception filtering along with the feature. I will address this in a subsequent patch.

> facilitate doing this in a less intrusive way?

I'm open to creating a feature proposal for this, but I'm unclear about how a less intrusive API would function. Based on the comments in this thread, the problem isn't that the patch causes an infinite loop or interferes with other projects' patches, is it?

As for exception captures, wouldn't `Puma::HttpParserError`, `HttpParserError501`, and `MiniSSL::SSLError` show up in the users' `lowlevel_error_handler` proc if they configure one? In other words, if a user also invokes `Sentry.capture_exception` within their `lowlevel_error_handler`, won't they encounter a similar issue until they manually filter out those exceptions? According to the README, that seems to be a primary use case for the configuration?

dentarg:
> I'm open to creating a feature proposal for this, but I'm unclear about how a less intrusive API would function. Based on the comments in this thread, the problem isn't that the patch causes an infinite loop or interferes with other projects' patches, is it?

It is not easy to opt-out of the monkey patch (even impossible perhaps?)

> As for exception captures, wouldn't `Puma::HttpParserError`, `HttpParserError501`, and `MiniSSL::SSLError` show up in the users' `lowlevel_error_handler` proc if they configure one? In other words, if a user also invokes `Sentry.capture_exception` within their `lowlevel_error_handler`, won't they encounter a similar issue until they manually filter out those exceptions? According to the README, that seems to be a primary use case for the configuration?

Yes, but if the user have a low level handler invoking `Sentry.capture_exception` they have probably already added the exceptions they want to exclude, to the exclude list.

Yeah, the problem was rolling this out as opt-in by default without any exclusions. You could have made it opt-in but I guess you thought the adoption of the feature would have been slow then.

st0012:
Yeah so if the problem is exception filtering, I'm sorry for not including it in the first place. 

However, given that
- Introducing new APIs in puma won't replace the ignoring list in #2055 
- We'll still need to keep the monkey-patch for older puma versions

I don't plan to initiate a new API for this feature. But if you do plan to propose one, I'd love to participate the discussion and become early adopters 😄

PR #2029, opened 2023-04-06
Add spec to Backtrace::APP_DIRS_PATTERN

sl0thentr0py:
while working on the profiler, I noticed `spec` was missing here.


HazAT (approved):

st0012 (approved):
Not a blocker, but we probably should find another non-in-app trace to verify that check.

sl0thentr0py:
added :)

st0012 (approved):

Issue #2035, opened 2023-04-19
Support reading SENTRY_TRACES_SAMPLE_RATE from environment


nbr:
**Describe the idea**

Both the Python and JavaScript SDKs support setting `$SENTRY_TRACES_SAMPLE_RATE` in the environment, whereas the Ruby SDK does not (the currently supported environment variables seem to be listed [here](https://github.com/getsentry/sentry-ruby/blob/a466ec8775a8310005879127cf99c62073b19b63/sentry-rails/spec/spec_helper.rb#L48-L51)).

**Why do you think it's beneficial to most of the users**

Uniform configuration across SDKs would help make the developer experience more intuitive.

**Possible implementation**

Fall back to environment variable if `traces_sample_rate` not set in `Sentry.init()`

PR #2036, opened 2023-04-25
handle sample rate type incompatibilities

nbr:
## Description

The suggested workaround in https://github.com/getsentry/sentry-ruby/issues/2035 results in an `ArgumentError`, since we aren't casting to `Float`. This PR will allow us to fail earlier and with a more specific error message.

### DIFF
--- a/sentry-ruby/lib/sentry/configuration.rb
+++ b/sentry-ruby/lib/sentry/configuration.rb
@@ -216,8 +216,8 @@ module Sentry
     attr_reader :transport
 
     # Take a float between 0.0 and 1.0 as the sample rate for tracing events (transactions).
-    # @return [Float]
-    attr_accessor :traces_sample_rate
+    # @return [Float, nil]
+    attr_reader :traces_sample_rate
 
     # Take a Proc that controls the sample rate for every tracing event, e.g.
     # @example
@@ -327,7 +327,6 @@ module Sentry
       self.before_send = nil
       self.before_send_transaction = nil
       self.rack_env_whitelist = RACK_ENV_WHITELIST_DEFAULT
-      self.traces_sample_rate = nil
       self.traces_sampler = nil
       self.enable_tracing = nil
 
@@ -409,9 +408,13 @@ module Sentry
       @traces_sample_rate ||= 1.0 if enable_tracing
     end
 
+    def traces_sample_rate=(traces_sample_rate)
+      @traces_sample_rate = traces_sample_rate&.to_f
+    end
+
     def profiles_sample_rate=(profiles_sample_rate)
       log_info("Please make sure to include the 'stackprof' gem in your Gemfile to use Profiling with Sentry.") unless defined?(StackProf)
-      @profiles_sample_rate = profiles_sample_rate
+      @profiles_sample_rate = profiles_sample_rate&.to_f
     end
 
     def sending_allowed?
--- a/sentry-ruby/lib/sentry/configuration.rb
+++ b/sentry-ruby/lib/sentry/configuration.rb
@@ -446,19 +446,22 @@ module Sentry
       enabled_environments.empty? || enabled_environments.include?(environment)
     end
 
+    def valid_sample_rate?(sample_rate)
+      sample_rate &&
+      sample_rate.respond_to?(:>=) &&
+      sample_rate >= 0.0 &&
+      sample_rate.respond_to?(:<=) &&
+      sample_rate <= 1.0
+    end
+
     def tracing_enabled?
-      valid_sampler = !!((@traces_sample_rate &&
-                          @traces_sample_rate >= 0.0 &&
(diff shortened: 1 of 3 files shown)


sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -435,19 +438,22 @@ def enabled_in_current_env?
       enabled_environments.empty? || enabled_environments.include?(environment)
     end
 
+    def valid_sample_rate?(sample_rate)
+      sample_rate &&
+      sample_rate.respond_to?(:>=) &&
+      sample_rate >= 0.0 &&
+      sample_rate.respond_to?(:<=) &&
+      sample_rate <= 1.0
```
st0012, line 446: I think we can check the sample rate's type instead of the methods it supports. We don't need to be that flexible here.

```suggestion
      return false unless sample_rate.is_a?(Numeric)
      sample_rate >= 0.0 && sample_rate <= 1.0
```

sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -401,9 +400,13 @@ def enable_tracing=(enable_tracing)
       @traces_sample_rate ||= 1.0 if enable_tracing
     end
 
+    def traces_sample_rate=(traces_sample_rate)
```
st0012, line 403: Let's raise `ArgumentError` here and in `profiles_sample_rate=` if the argument isn't a `Numeric`.

We already raise `ArgumentError` when users pass non-callable values to callable configs. So this aligns with the current practice.

st0012 (commented):
Thanks for raising the issue. I think in the case of configurations, we should raise errors to explicitly warn users about unexpected values.

sentry-ruby/lib/sentry/configuration.rb
nbr: @st0012 I've implemented this, but it was also necessary to allow `nil` since that is the default value for both sample rates. Otherwise, this causes many tests to fail.

sentry-ruby/spec/sentry/configuration_spec.rb
```diff
@@ -57,6 +57,25 @@
...
+        expect { subject.traces_sample_rate = "foobar" }.to raise_error(ArgumentError)
+      end
+    end
+
+    context "when enable_tracing is true" do
+      it "is non-Numeric and results in an ArgumentError" do
+        subject.enable_tracing = true
+        expect { subject.traces_sample_rate = "foobar" }.to raise_error(ArgumentError)
+      end
+    end
```
st0012, line 76: Given `traces_sample_rate=` doesn't check the value of `enable_tracing`, I don't think having those context blocks are necessary?

I also think we should cover more cases wrt the value it accepts and returns.

```suggestion
    it "returns nil by default" do
      expect(subject.traces_sample_rate).to eq(nil)
    end

    it "accepts Numeric values" do
      subject.traces_sample_rate = 1
      expect(subject.traces_sample_rate).to eq(1)
      subject.traces_sample_rate = 1.0
      expect(subject.traces_sample_rate).to eq(1.0)
    end

    it "accepts nil value" do
      subject.traces_sample_rate = 1
      subject.traces_sample_rate = nil
      expect(subject.traces_sample_rate).to eq(nil)
    end

    it "raises ArgumentError when the value is not Numeric nor nil" do
      expect { subject.traces_sample_rate = "foobar" }.to raise_error(ArgumentError)
    end
```

sentry-ruby/spec/sentry/configuration_spec.rb
```diff
@@ -154,6 +173,16 @@
     end
   end
 
+  describe "#profiles_sample_rate" do
+    it "returns nil by default" do
```
st0012, line 177: Let's replicate `traces_sample_rate`'s tests here if you agree with my suggestions.

sentry-ruby/lib/sentry/configuration.rb
st0012: Yeah we need to take `nil` too.

st0012 (commented):
> This will result in Configuration#trace_sample_rate == 0.0.

I was a bit confused by this statement. Is that still true after your change?

nbr:
> > This will result in Configuration#trace_sample_rate == 0.0.
> 
> I was a bit confused by this statement. Is that still true after your change?

It is no longer true, I've now updated the description.

st0012:
This looks good now. Can you add a changelog entry for it? Thx

nbr:
@st0012 Changelog entry added here: https://github.com/getsentry/sentry-ruby/pull/2036/commits/0d99b2fdc1cfcd1bb450addb66204c4c40913a63. Thanks for your time reviewing this.

st0012 (approved):
Thank you so much!

PR #2039, opened 2023-05-12
Move http query to span data

sl0thentr0py:
Align with the [query/fragment http spec](https://develop.sentry.dev/sdk/data-handling/#spans).

Notes:
* username/password in the url is not an issue since we already don't expose those
* fragment is currently not extractable from `req` in our net/http patch, so just ignore for now

fixes https://github.com/getsentry/sentry-ruby/issues/1955

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Features
+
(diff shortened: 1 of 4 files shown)


st0012 (approved):
👍

PR #2040, opened 2023-05-12
Validate release is a String during configuration

### LINKED ISSUE #2004: Throw warning when release is not a string
avoid issues like https://github.com/getsentry/sentry-ruby/pull/2000

sl0thentr0py:
closes #2004

### DIFF
--- a/sentry-ruby/lib/sentry/configuration.rb
+++ b/sentry-ruby/lib/sentry/configuration.rb
@@ -14,6 +14,8 @@ module Sentry
   class Configuration
     include CustomInspection
     include LoggingHelper
+    include ArgumentCheckingHelper
+
     # Directories to be recognized as part of your app. e.g. if you
     # have an `engines` dir at the root of your project, you may want
     # to set this to something like /(app|config|engines|lib)/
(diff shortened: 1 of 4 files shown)


sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -500,6 +506,12 @@ def check_callable!(name, value)
       end
     end
 
+    def check_instance!(name, value, klass)
```
st0012, line 509: We already have a [helper](https://github.com/getsentry/sentry-ruby/blob/master/sentry-ruby/lib/sentry/utils/argument_checking_helper.rb) to check argument types. Maybe we can update that method so it accepts 2 expected types?

sl0thentr0py: hmm I don't even think we need the `nil` check because no one should be setting release explicity to nil, I'll change the `release_detector` assignment above to an instance variable.

st0012: It may be easier for users because they can just set this and don't need to worry about if the env is present.

```rb
config.release = ENV["MY_RELEASE"]
```

If we raise in that case, users need to write 

```rb
config.release = ENV["MY_RELEASE"] if ENV["MY_RELEASE"] 
```

AFTER seeing at least one error.

st0012 (approved):
Looks good. I will try a small refactor after this is merged.
BTW, is it possible that we can work with codecov developers to improve the coverage accuracy in this repo? These false warnings are pretty annoying 😕

Issue #2043, opened 2023-05-17
JRuby: unable to capture Java exceptions


richard-kramer:
### Issue Description

Using sentry in JRuby, it is not possible to directly capture exceptions, which originated in java, as they do not inherit from `Exception`.

Currently, the only ways to log these exceptions, would be to wrap the Java exception in a custom Ruby exception an pass that to Sentry or to patch the `is_a?` method on the Java exception to respond with `true` if `Exception` is passed. Both solutions don't really seem that clean.

I also submitted an [issue over at JRuby](https://github.com/jruby/jruby/issues/7792), but it seems that behavior is intended. Java exceptions inherit from `Java::JavaLang::Error` and `Java::JavaLang::Throwable`, so maybe one of those (probably `Java::JavaLang::Throwable`) could be added to [the type check](https://github.com/getsentry/sentry-ruby/blob/c0dd6dfa977b61df69857ee043202e629029233c/sentry-ruby/lib/sentry/hub.rb#L119) in JRuby?

### Reproduction Steps

The following snippet enters the `rescue` block, but the exception cannot be captured:

```ruby
begin
  raise java.lang.OutOfMemoryError.new("A Java error")
rescue Exception => e
  e.is_a?(Exception) # => false
  Sentry.capture_exception(e) # => expect the argument to be a Exception, got Java::JavaLang::OutOfMemoryError (java.lang.OutOfMemoryError: A Java error)
end
```

### Expected Behavior

The exception can be captured by Sentry as it should respond to the expected interface of an exception.

### Actual Behavior

Capturing the exception results in an error thrown by sentry because of [this type check](https://github.com/getsentry/sentry-ruby/blob/c0dd6dfa977b61df69857ee043202e629029233c/sentry-ruby/lib/sentry/hub.rb#L119).

### Ruby Version

jruby-9.2.20.0

### SDK Version

5.5.0

### Integration and Its Version

sentry-rails (5.5.0)

sl0thentr0py:
thanks @richard-kramer we should definitely support this if possible. PRs are always welcome, otherwise I'll take a look at the possibilities when I have some time.

richard-kramer:
I created #2044 as a simple fix. Hopefully, this is already enough.

st0012:
@richard-kramer I think this can be closed?

PR #2044, opened 2023-05-17
Allow java exceptions as capture type in JRuby (#2043)

richard-kramer:
This should fix the problems described in #2043. It should now be possible to directly capture exceptions originating in Java when running in JRuby.

I'm not entirely sure, if the implementation is clean enough, and am open for feedback. I based it on [something similar in sentry/raven](https://github.com/getsentry/sentry-ruby/blob/fea18fee3dabd9c7af47a7811bc432df96a58799/sentry-raven/spec/raven/event_spec.rb#L721).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Move `http.query` to span data in net/http integration [#2039](https://github.com/getsentry/sentry-ruby/pull/2039)
 - Validate `release` is a `String` during configuration [#2040](https://github.com/getsentry/sentry-ruby/pull/2040)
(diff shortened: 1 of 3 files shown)


st0012:
Sorry for the delay. Yeah I think the implementation is reasonable.
Just one small request: can you make changelog's changes limit to the new entry?

richard-kramer:
Oh, I'm sorry. Seems, like I dodn't notice my formatter formatting the whole file :see_no_evil:. Should be fixed now.

st0012 (approved):
Thank you 👍

PR #2049, opened 2023-05-27
Support Rails 7.1's show exception check

st0012:
The `request.show_exceptions?` method will be removed in Raiils 7.1 (https://github.com/rails/rails/pull/45867)

So this PR adopts the new API to check whether to show exception if the Rails version is 7.1 or later.


st0012:
👋  @sl0thentr0py we need this to fix CI

Issue #2071, opened 2023-07-16
Expose `:value` in `SingleExceptionInterface`


jeffwidman:
👋 Hi from the :dependabot:  team.

I'm attempting to migrate from `sentry-raven` to `sentry-ruby`:
* https://github.com/dependabot/dependabot-core/pull/7553

With Raven we performed some basic client-side sanitization of the exception message:
https://github.com/dependabot/dependabot-core/blob/65f1c2d24e88ac57ed8df980871446deb5ea2d94/updater/lib/dependabot/sentry.rb#L5-L21

The unit tests illustrate how it works:
https://github.com/dependabot/dependabot-core/blob/65f1c2d24e88ac57ed8df980871446deb5ea2d94/updater/spec/dependabot/sentry_spec.rb#L24-L91

Migrating this from Raven to `sentry-ruby`'s `before_send` is straightforward. However, it looks like `sentry-ruby` doesn't expose `:value` on `SingleExceptionInterface`:
https://github.com/getsentry/sentry-ruby/blob/7764384e07bf7927c5a9a9d4c19688e9c1b11f56/sentry-ruby/lib/sentry/interfaces/single_exception.rb#L14

I see a recent change was made to expose `:values` on `ExceptionInterface`:
* https://github.com/getsentry/sentry-ruby/pull/1843

So I'm guessing there's no technical reason this can't be made publicly writable? Just no one has raised a feature request for it yet?

jeffwidman:
For now I've hacked around it using `instance_variable_set` but that feels a bit dirty:
https://github.com/dependabot/dependabot-core/pull/7553/files#diff-9ee3e6391d18c069de715a1740bb6239aa5f5c4734b3c03acd3455ba8bba3702R22-R23

sl0thentr0py:
yea I can expose it, PR obviously welcome to speed it up

jeffwidman:
Thanks, I put up a PR to expose it two weeks ago:
* https://github.com/getsentry/sentry-ruby/pull/2072

sl0thentr0py:
sorry didn't see that, need some time to catch up with this repo

PR #2072, opened 2023-07-16
Expose `:value` in `SingleExceptionInterface`

jeffwidman:
Expose `:value` in `SingleExceptionInterface` for users wishing to do basic sanitization of the exception message string in `before_send()`.

For example, the :dependabot: project does basic sanitization of `:value` using `sentry-raven`:
https://github.com/dependabot/dependabot-core/blob/65f1c2d24e88ac57ed8df980871446deb5ea2d94/updater/lib/dependabot/sentry.rb#L21

So exposing this makes the migration path easier for those users. And makes it possible for anyone else to do basic redaction of potential secrets, etc.

Fix https://github.com/getsentry/sentry-ruby/issues/2071

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+
+## 5.11.0
+
+### Features
+
+- Make `:value` in `SingleExceptionInterface` writable, so that it can be modified in `before_send` under `event.exception.values[n].value`
+
 ## 5.10.0
 
 ### Features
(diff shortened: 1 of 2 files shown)


jeffwidman:
I should caveat this PR with a note that I'm relatively new to Ruby, so if the goal is reasonable but the implementation is lacking, please point me in the right direction.

sl0thentr0py (approved):

jeffwidman:
Awesome thanks for merging!

Issue #1279, opened 2021-02-10
[Plan] Version 6


st0012:
- Drop support of Rails 5.0 & 5.1
- Drop support of Ruby 2.4
- Drop some configuration options
  - `async` - #1522
  - `linecache` - I guess no one is using this. if you do, please let me know your use case
  - `exclude_loggers` - this only works with `sentry_logger` breadcrumb logger. given that probably not many people use the logger, there should be even fewer (or no) people use this option.
- `Event#to_hash` only returns a string key hash.
  - Currently, Hash-form events have string keys if they're generated from `#to_json_compatible`. But `#to_hash` returns a  symbol key hash. So it can be confusing to users when dealing with events in callbacks.
- Remove `hub` argument from `Transaction#finish`
- Advanced Sidekiq reporting control
  - We may support more fine-grained control on retried jobs reporting. For example,
    - `attempt_threshold` - Only start reporting after the `nth` retry failed - https://github.com/getsentry/sentry-ruby/issues/1632
    - `reportable_retries` -  Only report a number of times for retry failures - https://github.com/getsentry/sentry-ruby/issues/1570
- Type signature (maybe)
- Fix stacktrace trimming https://github.com/getsentry/sentry-ruby/issues/2018

（feel free to add your feedback/request)

dlibanori:
Please, drop faraday. As utility library you shouldn't set any gem your client may be using. Right now I have to "upgrade" my farady to migrate from `sentry-raven` to `sentry-ruby` :(

st0012:
@dlibanori the version requirement on `faraday` is to make sure it'll always work with the SDK.  you can use `sentry-ruby-core` to bypass the limitation. but depending on the `faraday` version you have, it may or may not work.

AlexWayfer:
Update Faraday to version 2?

st0012:
@AlexWayfer Faraday is no longer the SDK's dependency after version 5.

AlexWayfer:
> @AlexWayfer Faraday is no longer the SDK's dependency after version 5.

Sorry, my bad, there left other dependencies after 4→5 update.

adis-io:
Hi, any dates on when `sentry-raven` stops working

jeffwidman:
Looks like this was scheduled for removal in `v5` but accidentally left in, so should probably ship with `v6`?
* https://github.com/getsentry/sentry-ruby/pull/2073

sl0thentr0py:
[...]

PR #2073, opened 2023-07-16
Delete deprecated `configuration` method

jeffwidman:
Use `Sentry.configuration` instead.

This was slated for removal in `v5.0.0` but was accidentally left in, so removing as part of `v6.0.0`.

Related:
* https://github.com/getsentry/sentry-ruby/issues/1279

### DIFF
--- a/sentry-ruby/lib/sentry/event.rb
+++ b/sentry-ruby/lib/sentry/event.rb
@@ -98,12 +98,6 @@ module Sentry
       end
     end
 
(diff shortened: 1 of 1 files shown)


sl0thentr0py:
hey @jeffwidman thanks for this but at some point I'll make a separate PR going through all the deprecations systematically when we do the major. Closing now, sorry!

jeffwidman:
How are you tracking not forgetting them? I opened the PR because it looked like it'd been overlooked on the last major version bump. 😃

Issue #2078, opened 2023-07-27
Add trace_propagation_targets


sl0thentr0py:
needed for https://github.com/getsentry/sentry-ruby/issues/2056
spec - https://develop.sentry.dev/sdk/performance/#tracepropagationtargets

Issue #2056, opened 2023-06-19
Tracing without Performance


stephanie-anderson:
# Description

Internal Notion Page:
https://www.notion.so/sentry/Tracing-without-performance-efab307eb7f64e71a04f09dc72722530

To always have access to a trace and span ID, add a new internal `PropagationContext` property to the scope, an object holding a `traceId` and `spanId` and an optional `dynamicSamplingContext`.

Populate the `traceId` and `spanId` with valid, random IDs on construction.

You may update the `traceId` and `dynamicSamplingContext` from the request headers during an incoming request or if the process was exposed to a `SENTRY_TRACE` and/or `SENTRY_BAGGAGE` environment variable if performance is disabled.

The Dynamic Sampling Context should be lazily constructed once needed, using a new `fromOptions` function that constructs the DSC based on the client options if performance is disabled.

When `captureEvent` is called, you may use these values on the scope as a fallback to construct the `trace` context and envelope header item if no transaction exists in the SDK.

For outgoing HTTP requests, use the values in the scope as a fallback if no transaction is present. The `sentry-trace` will contain a trailing `-0` indicating an unsampled transaction.

```[tasklist]
### Task List for each SDK
- [x] All events are always sent through the `/envelope` endpoint if performance is enabled. (If possible just always use `/envelope`)
- [ ] https://github.com/getsentry/sentry-ruby/issues/2099
- [x] All **error**, **transaction** and **check-in** events sent to Sentry always contain a Dynamic Sampling Context (trace) envelope header item.
- [x] All **error, transaction** and **check-in** sent to Sentry always contain [a `trace` context](https://develop.sentry.dev/sdk/event-payloads/contexts/#trace-context).
- [x] All outgoing **HTTP requests** always contain a `sentry-trace` and `baggage` header if the destination is allow-listed in [[tracePropagationTargets](https://develop.sentry.dev/sdk/performance/#tracepropagationtargets)](https://develop.sentry.dev/sdk/performance/#tracepropagationtargets). https://github.com/getsentry/sentry-ruby/issues/2078
- [ ] Create detailed docs about how to setup distributed tracing
- [x] Set the `replay` context (Python & PHP only) https://develop.sentry.dev/sdk/event-payloads/contexts/#replay-context
```

sl0thentr0py:
tracking env variables separately, closing this

PR #2079, opened 2023-07-27
Add new trace_propagation_targets configuration 

sl0thentr0py:
to set outgoing targets for trace header propagation according to https://develop.sentry.dev/sdk/performance/#tracepropagationtargets
closes https://github.com/getsentry/sentry-ruby/issues/2078

### DIFF
--- a/sentry-ruby/lib/sentry/net/http.rb
+++ b/sentry-ruby/lib/sentry/net/http.rb
@@ -34,7 +34,7 @@ module Sentry
           set_sentry_trace_header(req, sentry_span, request_info)
 
           super.tap do |res|
-            record_sentry_breadcrumb(req, res, request_info)
+            record_sentry_breadcrumb(request_info, res)
 
             if sentry_span
               sentry_span.set_description("#{request_info[:method]} #{request_info[:url]}")
@@ -53,7 +53,6 @@ module Sentry
         return unless sentry_span
 
         client = Sentry.get_current_client
-        return unless client
         return unless propagate_trace?(request_info[:url], client.configuration.trace_propagation_targets)
 
         trace = client.generate_sentry_trace(sentry_span)
@@ -63,7 +62,7 @@ module Sentry
         req[BAGGAGE_HEADER_NAME] = baggage if baggage && !baggage.empty?
       end
 
-      def record_sentry_breadcrumb(req, res, request_info)
+      def record_sentry_breadcrumb(request_info, res)
         return unless Sentry.initialized? && Sentry.configuration.breadcrumbs_logger.include?(:http_logger)
 
         crumb = Sentry::Breadcrumb.new(
--- a/sentry-ruby/lib/sentry/net/http.rb
+++ b/sentry-ruby/lib/sentry/net/http.rb
@@ -30,13 +30,13 @@ module Sentry
         return super if from_sentry_sdk?
 
         Sentry.with_child_span(op: OP_NAME, start_timestamp: Sentry.utc_now.to_f) do |sentry_span|
-          set_sentry_trace_header(req, sentry_span)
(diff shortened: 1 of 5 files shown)


sentry-ruby/lib/sentry/net/http.rb
```diff
@@ -61,11 +63,9 @@ def set_sentry_trace_header(req, sentry_span)
         req[BAGGAGE_HEADER_NAME] = baggage if baggage && !baggage.empty?
       end
 
-      def record_sentry_breadcrumb(req, res)
+      def record_sentry_breadcrumb(req, res, request_info)
```
st0012, line 66: It looks like `req` here is not needed anymore?

sentry-ruby/lib/sentry/net/http.rb
```diff
@@ -49,10 +49,12 @@ def request(req, body = nil, &block)
 
       private
 
-      def set_sentry_trace_header(req, sentry_span)
+      def set_sentry_trace_header(req, sentry_span, request_info)
         return unless sentry_span
 
         client = Sentry.get_current_client
+        return unless client
```
st0012, line 56: Is there a case where Span is initialised (which means the SDK is initialised too?) and but client is `nil`?

sentry-ruby/lib/sentry/net/http.rb
```diff
@@ -30,13 +30,13 @@ def request(req, body = nil, &block)
         return super if from_sentry_sdk?
 
         Sentry.with_child_span(op: OP_NAME, start_timestamp: Sentry.utc_now.to_f) do |sentry_span|
-          set_sentry_trace_header(req, sentry_span)
+          request_info = extract_request_info(req)
+          set_sentry_trace_header(req, sentry_span, request_info)
```
st0012, line 34: I feel we can perform the `propagate_trace?` check here without changing `set_sentry_trace_header`.
And we can use `Sentry.configuration` too.

sl0thentr0py: this'll change a bit again as part of https://github.com/getsentry/sentry-ruby/issues/2056 so I'll clean the whole logic there if it's fine?

sentry-ruby/lib/sentry/net/http.rb
sl0thentr0py: only if there's some weird multi-threaded race, since we had a couple of bugs like that before in redis, just wanted to be safe

sentry-ruby/lib/sentry/net/http.rb
st0012: Will it be implemented soon? I think it'll be fine if we ship them in the sam release.

sentry-ruby/lib/sentry/net/http.rb
st0012: If you mean #2057, it's addressed with SDK initialisation check.

And in our case here, `set_sentry_trace_header` is only called in `HTTP#request`, which already checked `Sentry.initialized?`. `Sentry.with_child_span` also yields `nil` if the SDK is not initialised, which we also handled at the beginning of this method.

Given we already have 2 checks on the same thing, I feel we're covered from the issue you mentioned. And to me this actually feels a bit misleading instead.

sentry-ruby/lib/sentry/net/http.rb
sl0thentr0py: yea like this/next week

sentry-ruby/lib/sentry/net/http.rb
sl0thentr0py: removed

st0012 (approved):

PR #2084, opened 2023-08-02
Tracing without performance

### LINKED ISSUE #2056: Tracing without Performance

sl0thentr0py:
This PR is 1/2 to enable **Tracing without Performance**, i.e. make sure all our events are connected even if they are not Transactions.

This enables use cases such as Errors / Transactions / Replays etc all being connected across services and not just Transactions.

### Summary of changes

* new `PropagationContext` class that generates trace/span ids and baggage irrespective of whether there are transactions/spans active or not
* this lives on the `Scope`
* three new top level methods that first check the span and fallback on the scope's propagation context
  * `Sentry.get_traceparent`
  * `Sentry.get_baggage`
  * `Sentry.get_trace_propagation_headers`
* move `dynamic_sampling_context` to `Event` from `TransactionEvent` since all events will now have this info
* use the new top level helpers in `net/http` patch to set propagation headers

closes #2056 
also see #2089

### DIFF
--- a/sentry-ruby/lib/sentry/event.rb
+++ b/sentry-ruby/lib/sentry/event.rb
@@ -37,6 +37,11 @@ module Sentry
     # @return [RequestInterface]
     attr_reader :request
 
+    # Dynamic Sampling Context (DSC) that gets attached
+    # as the trace envelope header in the transport.
+    # @return [Hash, nil]
+    attr_accessor :dynamic_sampling_context
+
     # @param configuration [Configuration]
     # @param integration_meta [Hash, nil]
     # @param message [String, nil]
--- a/sentry-ruby/lib/sentry/event.rb
+++ b/sentry-ruby/lib/sentry/event.rb
@@ -59,6 +59,7 @@ module Sentry
       @tags          = {}
 
       @fingerprint = []
+      @dynamic_sampling_context = nil
 
       # configuration data that's directly used by events
       @server_name = configuration.server_name
--- a/sentry-ruby/lib/sentry/hub.rb
+++ b/sentry-ruby/lib/sentry/hub.rb
@@ -246,10 +246,15 @@ module Sentry
     def get_trace_propagation_headers
       return nil unless configuration.propagate_traces
 
-      {
-        SENTRY_TRACE_HEADER_NAME => get_traceparent,
-        BAGGAGE_HEADER_NAME => get_baggage
-      }.compact
+      headers = {}
+
+      traceparent = get_traceparent
+      headers[SENTRY_TRACE_HEADER_NAME] = traceparent if traceparent
+
+      baggage = get_baggage
+      headers[BAGGAGE_HEADER_NAME] = baggage if baggage && !baggage.empty?
+
+      headers
     end
 
     private
--- a/sentry-ruby/lib/sentry/hub.rb
+++ b/sentry-ruby/lib/sentry/hub.rb
@@ -243,6 +243,15 @@ module Sentry
         current_scope.propagation_context&.get_baggage&.serialize
     end
 
+    def get_trace_propagation_headers
+      return nil unless configuration.propagate_traces
+
+      {
+        SENTRY_TRACE_HEADER_NAME => get_traceparent,
+        BAGGAGE_HEADER_NAME => get_baggage
+      }.compact
+    end
+
     private
 
     def current_layer
--- a/sentry-ruby/lib/sentry/hub.rb
+++ b/sentry-ruby/lib/sentry/hub.rb
@@ -229,6 +229,20 @@ module Sentry
       end_session
     end
 
+    def get_traceparent
+      return nil unless current_scope
+
+      current_scope.get_span&.to_sentry_trace ||
+        current_scope.propagation_context&.get_traceparent
+    end
+
+    def get_baggage
+      return nil unless current_scope
+
+      current_scope.get_span&.to_baggage ||
+        current_scope.propagation_context&.get_baggage&.serialize
+    end
+
     private
 
     def current_layer
--- a/sentry-ruby/lib/sentry/hub.rb
+++ b/sentry-ruby/lib/sentry/hub.rb
@@ -233,14 +233,14 @@ module Sentry
       return nil unless current_scope
 
       current_scope.get_span&.to_sentry_trace ||
-        current_scope.propagation_context&.get_traceparent
+        current_scope.propagation_context.get_traceparent
     end
 
     def get_baggage
       return nil unless current_scope
 
       current_scope.get_span&.to_baggage ||
-        current_scope.propagation_context&.get_baggage&.serialize
+        current_scope.propagation_context.get_baggage&.serialize
     end
 
     def get_trace_propagation_headers
--- a/sentry-ruby/lib/sentry/hub.rb
+++ b/sentry-ruby/lib/sentry/hub.rb
@@ -255,6 +255,24 @@ module Sentry
       headers
     end
 
+    def continue_trace(env, **options)
+      configure_scope { |s| s.generate_propagation_context(env) }
+
+      return nil unless configuration.tracing_enabled?
+
+      propagation_context = current_scope.propagation_context
+      return nil unless propagation_context.incoming_trace
+
+      Transaction.new(
+        hub: self,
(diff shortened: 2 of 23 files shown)


sl0thentr0py:
@st0012 still gonna write some specs and test but feel free to start reviewing if you have time and let me know if you have questions on why we are doing this!

sentry-ruby/lib/sentry/propagation_context.rb
```diff
@@ -0,0 +1,78 @@
...
+
+      items = {
+        "trace_id" => trace_id,
+        "sample_rate" => configuration.traces_sample_rate&.to_s,
+        "environment" => configuration.environment,
+        "release" => configuration.release,
+        "public_key" => configuration.dsn&.public_key
+      }
+
+      user = @scope&.user
```
st0012, line 71: I don't think `@scope` could be `nil`?

sentry-ruby/lib/sentry/propagation_context.rb
```diff
@@ -0,0 +1,78 @@
...
+      items["user_segment"] = user["segment"] if user && user["segment"]
```
st0012, line 72: Since we'll call `items.compact!`, we can define `user_segment` in the items hash too.

```rb
      items = {
        "trace_id" => trace_id,
        "sample_rate" => configuration.traces_sample_rate&.to_s,
        "environment" => configuration.environment,
        "release" => configuration.release,
        "public_key" => configuration.dsn&.public_key,
        "user_segment" => user && user["segment"]
      }
```

sentry-ruby/lib/sentry/event.rb
```diff
@@ -37,6 +37,11 @@ class Event
     # @return [RequestInterface]
     attr_reader :request
 
+    # Dynamic Sampling Context (DSC) that gets attached
+    # as the trace envelope header in the transport.
+    # @return [Hash, nil]
+    attr_accessor :dynamic_sampling_context
```
st0012, line 43: Nit: could we add `@dynamic_sampling_context = nil` to the constructor?

sentry-ruby/lib/sentry/propagation_context.rb
```diff
@@ -0,0 +1,78 @@
...
+    # Returns the Dynamic Sampling Context from the baggage.
+    # @return [String, nil]
+    def get_dynamic_sampling_context
+      get_baggage&.dynamic_sampling_context
+    end
+
+    private
+
+    def populate_head_baggage
+      return unless Sentry.initialized?
```
st0012, line 59: Since `PropagationContext` could only be initialised by an active `Scope`, which means the SDK is initialised, I think this is not necessary?

sentry-ruby/lib/sentry/hub.rb
```diff
@@ -229,6 +229,34 @@ def with_session_tracking(&block)
...
+
+      current_scope.get_span&.to_sentry_trace ||
+        current_scope.propagation_context&.get_traceparent
+    end
+
+    def get_baggage
+      return nil unless current_scope
+
+      current_scope.get_span&.to_baggage ||
+        current_scope.propagation_context&.get_baggage&.serialize
```
st0012, line 243: I think a scope's `propagation_context` should not be `nil`?

Also, `PropagationContext#get_baggage` should also not be `nil` as the early return in `populate_head_baggage` is not supposed to happen (explained in its comment).

sentry-ruby/lib/sentry/propagation_context.rb
sl0thentr0py: this is again that multi-threaded case where one thread calls `Sentry.close` and another thread is still running this code. If we don't want this check, we will have to pass `configuration` into the constructor and not access it with `Sentry.configuration`. 

I'm ok with having these checks and using the global methods, this is also how python works.

st0012: Since we have specific case in mind for the check, could we have a comment for it and perhaps also a test? Also, doesn't that mean anything that calls `Sentry.configuration` would need such check too, like [here](https://github.com/getsentry/sentry-ruby/blob/8abb2d213fb96ed45130956a3452e9cc7a097de0/sentry-ruby/lib/sentry/breadcrumb/sentry_logger.rb#L80)?

And perhaps we should implement a better version of `close` and `Sentry.initialize?` 🤔 

Specifically:

- We don't erase hub variables, `@background_worker` and `@session_flusher`
- `Sentry.initialize?` is backed by a simple boolean state that's changed by `Sentry.ini` and `Sentry.close`.

We would still need `Sentry.initialize?` in some places to stop some actions after the SDK is closed, but we don't need it for preventing`nil` value anymore, which is getting out of hand IMO.

sentry-ruby/lib/sentry/hub.rb
```diff
@@ -229,6 +229,34 @@ def with_session_tracking(&block)
       end_session
     end
 
+    def get_traceparent
+      return nil unless current_scope
+
+      current_scope.get_span&.to_sentry_trace ||
+        current_scope.propagation_context.get_traceparent
```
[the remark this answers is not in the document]
sl0thentr0py: i like early returns

sentry-ruby/lib/sentry/propagation_context.rb
sl0thentr0py: yea I agree we should fix the close/initialize behavior but that's out of scope here, I'll make a separate issue.
But for the record, other SDKs also work like this, python has `hub/client` checks at the beginning of most functions.

sl0thentr0py: https://github.com/getsentry/sentry-ruby/issues/2096

sentry-ruby/lib/sentry/propagation_context.rb
```diff
@@ -0,0 +1,76 @@
...
+
+      configuration = Sentry.configuration
+
+      items = {
+        "trace_id" => trace_id,
+        "sample_rate" => configuration.traces_sample_rate&.to_s,
```
cleptric, line 65: If this is only used if perf is disabled, I would not add the `sample_rate` to the headers.

cleptric (approved):

PR #2087, opened 2023-08-07
Add option to notify Sentry only after the last retry on resque

GlauberrBatista:
`sentry-sidekiq` and `sentry-delayed_job` have the option to report the failure only after the retry process has finished.

With `sentry-resque` we can do the same when using the `resque-retry` gem. The implementation is not as clean as `Sidekiq` and `DelayedJob`, which expose the attempts in a better way, but still, I was able to make it work.
I noticed the configuration option was there but not implemented. I added a conditional inside the exception block to check if the job is an instance of `Resque::Plugins::Retry` and if so, it checks the retry key.

An odd behavior: `resque-retry` cleans the key on the last try, before being caught on the exception block. Thus, when the key no longer exists on Redis, it means the retries were exhausted.

I understand this change implies using the `resque-retry` gem, but I'm unaware of a similar gem widely used as `resque-retry`.

### DIFF
--- a/sentry-resque/lib/sentry/resque.rb
+++ b/sentry-resque/lib/sentry/resque.rb
@@ -32,6 +32,15 @@ module Sentry
 
               finish_transaction(transaction, 200)
             rescue Exception => exception
+              klass = payload['class'].constantize
+
+              if Sentry.configuration.resque.report_after_job_retries && defined?(::Resque::Plugins::Retry) == 'constant' && klass.is_a?(::Resque::Plugins::Retry)
+                retry_key = klass.redis_retry_key(payload['args'])
+                retry_attempt = ::Resque.redis.get(retry_key)
+
+                # If retry_attempt is nil, it means the resque-retry has already called the #clean_retry_key no retries are left
+                raise unless retry_attempt.nil?
+              end
               ::Sentry::Resque.capture_exception(exception, hint: { background: false })
               finish_transaction(transaction, 500)
               raise
--- a/sentry-resque/lib/sentry/resque.rb
+++ b/sentry-resque/lib/sentry/resque.rb
@@ -34,13 +34,11 @@ module Sentry
             rescue Exception => exception
               klass = payload['class'].constantize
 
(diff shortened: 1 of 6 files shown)


GlauberrBatista:
Hello, any update on this?

sentry-resque/lib/sentry/resque.rb
```diff
@@ -32,6 +32,15 @@ def record(queue, worker, payload, &block)
 
               finish_transaction(transaction, 200)
             rescue Exception => exception
+              klass = payload['class'].constantize
+
+              if Sentry.configuration.resque.report_after_job_retries && defined?(::Resque::Plugins::Retry) == 'constant' && klass.is_a?(::Resque::Plugins::Retry)
+                retry_key = klass.redis_retry_key(payload['args'])
+                retry_attempt = ::Resque.redis.get(retry_key)
```
sl0thentr0py, line 39: the SDK should never do its own database calls by any chance, that's a side effect that breaks all sorts of contracts.
Is there no way of knowing the retry status from the stuff available to this function?

sl0thentr0py (changes requested):
hey sorry for the delay on this but I don't think this solution is something we can put in the SDK (see comment)

If there's another possibility to fix the logic, I can take that otherwise I'll dig into resque later.

sentry-resque/lib/sentry/resque.rb
GlauberrBatista: @sl0thentr0py  In fact, there is. I didn't realize it, but in fact, we are catching an exception here (so the processed job context is available) and when we use constantize against the class name, resque-retry will capture the job context, so methods `retry_limit_reached?` and `retry_limit_attempt` will expose the current "status" of the job.

I just pushed a new commit that fixes it.

sl0thentr0py (approved):
sorry for the delay again and ty @GlauberrBatista !

GlauberrBatista:
@sl0thentr0py not a problem and thank you 🙌

PR #2089, opened 2023-08-10
Tracing without performance: new continue_trace api

sl0thentr0py:
This PR is 2/2 to enable Tracing without Performance, i.e. make sure all our events are connected even if they are not Transactions.

### Summary of changes
* Implement new top level `Sentry.continue_trace(env, **options)` API that standardizes continuing an incoming trace from a rack env like hash.
* Use this new API in rack/rails/sidekiq

part of #2056 
linked to #2084

### DIFF
--- a/sentry-ruby/lib/sentry/hub.rb
+++ b/sentry-ruby/lib/sentry/hub.rb
@@ -255,6 +255,24 @@ module Sentry
       headers
     end
 
+    def continue_trace(env, **options)
+      configure_scope { |s| s.generate_propagation_context(env) }
+
+      return nil unless configuration.tracing_enabled?
+
+      propagation_context = current_scope.propagation_context
+      return nil unless propagation_context.incoming_trace
+
+      Transaction.new(
+        hub: self,
+        trace_id: propagation_context.trace_id,
+        parent_span_id: propagation_context.parent_span_id,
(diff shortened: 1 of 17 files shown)


sentry-ruby/lib/sentry/hub.rb
```diff
@@ -229,6 +229,52 @@ def with_session_tracking(&block)
       end_session
     end
 
+    def get_traceparent
+      return nil unless current_scope
+
+      current_scope.get_span&.to_sentry_trace ||
+        current_scope.propagation_context&.get_traceparent
```
szokeasaurusrex, line 236:
```suggestion
      current_scope&.get_span&.to_sentry_trace ||
        current_scope&.propagation_context&.get_traceparent
```
I believe this suggestion produces the same result more compactly

sentry-ruby/lib/sentry/hub.rb
```diff
@@ -229,6 +229,52 @@ def with_session_tracking(&block)
...
+    end
+
+    def get_baggage
+      return nil unless current_scope
```
szokeasaurusrex, line 240: Same comment as above

cleptric (approved):

sl0thentr0py:
@st0012 we're going forward with merging these PRs, let me know later if you some significant problem

PR #1923, opened 2022-10-23
Update Tracing Span's op names

st0012:
Closes #1903

PR #2093, opened 2023-08-17
Consolidate HTTP span data conventions with OpenTelemetry

sl0thentr0py:
in preparation for some of our new performance UIs we're consolidating span data keys with OpenTelemetry with a new `Span::DataConventions` module. 

This PR handles the HTTP spans.

see https://github.com/getsentry/team-sdks/issues/20


antonpirker (approved):
lgtm

st0012 (commented):
When we did a span name change in #1923, some users reported that it "broke" their filtering logic and resulted in higher-than-expected usages. Since status code could be a common method for filtering, I'm a bit concerned that similar issue could happen again. Should we still set `status_code` so users don't get breaking change in a minor version upgrade?

szokeasaurusrex (approved):
Looks good!

sl0thentr0py:
@st0012 good point, but this is `data` whereas those were ops, I think it's fine to clean this up without consequences. The span status is still the same and that's what I'd recommend for filtering just in case.

st0012 (approved):

PR #2104, opened 2023-09-08
Drop IRB's patch

st0012:
I've fixed the issue in https://github.com/ruby/irb/pull/704 and released it in v1.8.1, so this patch is no longer needed 😄 

#skip-changelog

### DIFF
--- a/sentry-rails/Gemfile
+++ b/sentry-rails/Gemfile
@@ -24,6 +24,7 @@ if rails_version > Gem::Version.new("7.0.0")
   gem "rails", github: "rails/rails"
 else
   gem "rails", "~> #{rails_version}"
+  gem "psych", "~> 3.0.0"
 end
 
 gem "mini_magick"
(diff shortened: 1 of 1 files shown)


st0012:
@sl0thentr0py Thanks for the workaround btw. I didn't notice the issue until I saw your patch & another issue report about it.

sl0thentr0py (approved):

sentry-rails/Gemfile
```diff
@@ -24,6 +24,7 @@ if rails_version > Gem::Version.new("7.0.0")
   gem "rails", github: "rails/rails"
 else
   gem "rails", "~> #{rails_version}"
+  gem "psych", "~> 3.0.0"
```
st0012, line 27: Because Rails 6 doesn't require `psych`'s version, it could still be bumped to a incompatible version. So we need to lock it here.

PR #2118, opened 2023-09-28
Fix Sidekiq tracing headers not being overwritten in case of schedules and retries

sl0thentr0py:
https://github.com/getsentry/sentry-ruby/pull/1774 added `SentryContextClientMiddleware` to the server middleware chain too.

Sidekiq's [scheduler pushes to the client on the server again for schedules and retries](https://github.com/sidekiq/sidekiq/blob/aadc77a172f1490fb141c7936d2801ca3af925ef/lib/sidekiq/scheduled.rb#L39) which causes our trace propagation to be broken for this case.

Prioritize taking the trace propagation headers from the job whenever they exist.


szokeasaurusrex (approved):
looks good

PR #2120, opened 2023-10-01
Adopt Rails 7.1's new BroadcastLogger

st0012:
In https://github.com/rails/rails/pull/48615, Rails 7.1 introduced a new `BroadcastLogger` class that allows users to send logs to multiple loggers, which means we need to adjust the SDK's logger assignment for it.


sl0thentr0py (approved):

st0012:
It turned out I also found a Rails issue when working on this 😄  https://github.com/rails/rails/pull/49417#discussion_r1342516128

PR #2126, opened 2023-10-05
Update sentry-sidekiq for Sidekiq 7.1.5

st0012:
In https://github.com/sidekiq/sidekiq/pull/6051, Sidekiq changed the expected parameters on the error handler. However, because it assumes error handler to be a Proc, it uses `#arity` for checking the number of parameters. This is not compatible with our error handler, which is a class.


st0012:
@sl0thentr0py Oh nice. I guess we just need to wait then.

PR #2127, opened 2023-10-05
Add `SimpleCov.command_name` for isolated_specs

sl0thentr0py:
this ensures the coverage generated will be the merged version and not just from the last `isolated_specs` run.

see https://github.com/simplecov-ruby/simplecov#test-suite-names

#skip-changelog

### DIFF
--- a/sentry-ruby/spec/isolated/puma_spec.rb
+++ b/sentry-ruby/spec/isolated/puma_spec.rb
@@ -4,6 +4,8 @@ require_relative "../spec_helper"
 # Because puma doesn't have any dependency, if Rack is not installed the entire test won't work
 return if ENV["RACK_VERSION"] == "0"
 
+SimpleCov.command_name "RSpecIsolated"
+
 RSpec.describe Puma::Server do
   class TestServer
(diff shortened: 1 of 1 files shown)


sentry-ruby/spec/isolated/puma_spec.rb
```diff
@@ -4,6 +4,8 @@
 # Because puma doesn't have any dependency, if Rack is not installed the entire test won't work
 return if ENV["RACK_VERSION"] == "0"
 
+SimpleCov.command_name "RSpecIsolated"
```
st0012, line 7: Should we use something like `__FILE__` to make it more explicit (I'm not sure if we'll need to read it though)? But it will save us the headache to name it for future isolated tests.

sl0thentr0py: > Note that this only has to be invoked ONCE PER TEST SUITE, so even if you have 200 unit test files, specifying it in some_test.rb is enough.

so even if we add more isolated specs, just having this one here is sufficient

cleptric (approved):

Issue #2090, opened 2023-08-14
Sentry Crons support!


adamzapasnik:
Is there any plan to support Sentry Crons in Ruby sdk?

https://docs.sentry.io/product/crons/getting-started/

stephanie-anderson:
Hey Adam! Yes, development is planned within the next 2-3 weeks ✌️

sl0thentr0py:
@adamzapasnik yes we will do this soon, can you tell me what gems you'd like to see support for?

sl0thentr0py:
gems we will support first
- sidekiq-cron
- sidekiq-scheduler
- whenever
- clockwork

sl0thentr0py:
manual crons support is done, automatic for gems split into separate task

sl0thentr0py:
released in 5.12.0! try it out and let us know @adamzapasnik

swanson:
@sl0thentr0py we've been testing this out this week in our project. Do you plan to add a `heartbeat` method for times when we aren't reporting runtime? This seems like it reads better in our code -- even if it is just an alias for `capture_check_in` that doesn't return the check_id

natikgadzhi:
> Do you plan to add a heartbeat method for times when we aren't reporting runtime? 

You can just send a check_in :ok, ignoring the returned value, that'll do a heartbeat.
```ruby
  Sentry.capture_check_in(
    "rake-task-example",
    :ok, 
    monitor_config: monitor_config
  )
```

Does that solve it?

> I'm also finding it a bit confusing that the Ruby parameters are things like checkin_margin but the UI in Sentry for this same field in the UI is called Grace Period.

I _think_ the underlying field is internally called `checkin_margin` in the API as well. 

Interestingly, I didn't find documentation on how to programmatically create monitors (i.e. pass `monitor_config`) in Python, but I see it accepts `monitor_config`, and I'm certain I've seen a demo of it — so I assume it's just a hash, let me double-check what keys are there.

sl0thentr0py:
> I'm also finding it a bit confusing that the Ruby parameters are things like checkin_margin but the UI in Sentry for this same field in the UI is called Grace Period. Might be worth aliasing these.

@swanson  I passed this feedback to the product team, thanks!

Re: heartbeat, see [this](https://docs.sentry.io/platforms/ruby/crons/#heartbeat) (and also as @natikgadzhi said above).
We're still iterating on crons so we might make some of this APIs better eventually, thanks for the feedback again!

[...]

PR #2130, opened 2023-10-05
Add `Sentry::Cron::MonitorCheckIns` module for automatic monitoring of jobs

sl0thentr0py:
Standard job frameworks such as `ActiveJob` and `Sidekiq` can now use this module to automatically capture check ins.

```rb
class ExampleJob < ApplicationJob
  include Sentry::Cron::MonitorCheckIns

  sentry_monitor_check_ins

  def perform(*args)
    # do stuff
  end
end
```

```rb
class SidekiqJob
  include Sidekiq::Job
  include Sentry::Cron::MonitorCheckIns

  sentry_monitor_check_ins

  def perform(*args)
    # do stuff
  end
end
```

You can pass in optional attributes to `sentry_monitor_check_ins` as follows.
```rb
# slug defaults to the job class name
sentry_monitor_check_ins slug: 'custom_slug'

# define the monitor config with an interval
sentry_monitor_check_ins monitor_config: Sentry::Cron::MonitorConfig.from_interval(1, :minute)

# define the monitor config with a crontab
sentry_monitor_check_ins monitor_config: Sentry::Cron::MonitorConfig.from_crontab('5 * * * *')
```

part of #2090

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,15 +4,56 @@
 
 - Record client reports for profiles [#2107](https://github.com/getsentry/sentry-ruby/pull/2107)
 - Adopt Rails 7.1's new BroadcastLogger [#2120](https://github.com/getsentry/sentry-ruby/pull/2120)
-- Add `Sentry.capture_check_in` API for Cron Monitoring [#2117](https://github.com/getsentry/sentry-ruby/pull/2117)
+- Add [Cron Monitoring](https://docs.sentry.io/product/crons/) support
+  - Add `Sentry.capture_check_in` API for Cron Monitoring [#2117](https://github.com/getsentry/sentry-ruby/pull/2117)
 
-  You can now track progress of long running scheduled jobs.
+    You can now track progress of long running scheduled jobs.
 
-  ```rb
-  check_in_id = Sentry.capture_check_in('job_name', :in_progress)
-  # do job stuff
-  Sentry.capture_check_in('job_name', :ok, check_in_id: check_in_id)
-  ```
+    ```rb
+    check_in_id = Sentry.capture_check_in('job_name', :in_progress)
+    # do job stuff
+    Sentry.capture_check_in('job_name', :ok, check_in_id: check_in_id)
+    ```
+  - Add `Sentry::Cron::MonitorCheckIns` module for automatic monitoring of jobs [#2130](https://github.com/getsentry/sentry-ruby/pull/2130)
+
+    Standard job frameworks such as `ActiveJob` and `Sidekiq` can now use this module to automatically capture check ins.
+
+    ```rb
+    class ExampleJob < ApplicationJob
+      include Sentry::Cron::MonitorCheckIns
+
+      sentry_monitor_check_ins
+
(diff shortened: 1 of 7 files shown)


cleptric (approved):

trevorturk:
👋 Hello and apologies if this is the wrong place to ask, but I upgraded my gems today and was excited to see this new feature. However, after deploying and having some Sidekiq jobs running, I'm not seeing anything in the Sentry web interface. In reading the docs, I'm unclear if I need to do additional work (perhaps visiting /crons and then clicking "Add Monitor") or should I see the cron monitors appear automatically if I've done the `include Sentry::Cron::MonitorCheckIns` and `sentry_monitor_check_ins` dance?

trevorturk:
Thank you for the info! Just to follow up a bit, I do have the following:

```ruby
class Api::ExampleJob
  include Sidekiq::Job
  include Sentry::Cron::MonitorCheckIns
  sentry_monitor_check_ins monitor_config: Sentry::Cron::MonitorConfig.from_crontab("0 * * * * *")
```

...but I'm gathering from other GH chatter that perhaps I need to include a `slug` param in the `monitor_config` bit, perhaps like so?

```ruby
monitor_config = Sentry::Cron::MonitorConfig.from_crontab(
  "0 * * * * *",
  slug: "example_job"
)
```

I'd prefer to try to auto-configure these, so I don't have to use the web interface for other monitors, so I'm curious what else would be required. 

I believe the relevant docs I think could use additional detail would be here: https://docs.sentry.io/platforms/ruby/crons/#job-monitoring

I appreciate the help, thank you!

sl0thentr0py:
Yeah this should be made clear in the docs, sorry about that.

```ruby
  sentry_monitor_check_ins monitor_config: Sentry::Cron::MonitorConfig.from_crontab("0 * * * * *")
```

this will work, it'll set the slug to the job class name.

Or you can do the following for a custom slug
```ruby 
 sentry_monitor_check_ins slug: "example_job", monitor_config: Sentry::Cron::MonitorConfig.from_crontab("0 * * * * *")

trevorturk:
Gotcha, so I think I'm all set? Perhaps my confusion is just that I expected to see something on `sentry.io/crons` telling me that my cron monitor was set up and is working. Might be worth considering for the product roadmap, if possible. Now I think I'd just expect to see something on `sentry.io/issues` if/when something goes wrong, but it'd be nice to see something in the web interface that confirmed the cron is being monitored. Now that I think of it, I'm not sure what would happen if I intentionally removed this job that's being monitored -- perhaps I'd see one error and then no more?

(Apologies for the chatter here, just trying to help with the beta feedback!)

sl0thentr0py:
np! always happy to hear feedback and this is quite new so we're still iterating.
> but it'd be nice to see something in the web interface that confirmed the cron is being monitored.

if you've pushed this out, you should see check-ins start appearing in the Crons section in the product. And error statuses will get logged into Issues as well.

sl0thentr0py:
that should work, can you make another issue with some details if possible?

Issue #907, opened 2019-06-26
NoMethodError: undefined method `byteslice' for []:Array


ericproulx:
Hello,

Sidekiq integrations calls `Raven.capture_exception` like this:
```ruby
 Raven.capture_exception(
        ex,
        :message => ex.message,
        :extra => { :sidekiq => context }
      )
```

Unfortunately, if `ex.message` isn't a String (in our case, it was a Hash), it fails with ```NoMethodError: undefined method `byteslice' for []:Array```

I know it's weird that `ex.message` isn't a String, but it works when calling it like this :
```ruby
# ex is an exception
Raven.capture_exception(ex)
```

When looking at code, `event.message` is interpolated properly only if `Raven.capture_exception` wasn't call with options `message`
```ruby
# lib/raven/event.rb

def self.from_exception(exc, options = {}, &block)
      ...
      new(options) do |evt|
        evt.message = "#{exc.class}: #{exc.message}" # interpolated properly

        evt.add_exception_interface(exc)

        yield evt if block
      end
    end
```

Is it a bug ? I don't know, but I had to `to_s` my value before raising an exception.

Thanks

st0012:
I think it doesn't make much sense to accept a non-string `message`. However, as a error-reporting SDK it shouldn't cause any new error when taking non-string input. I will open a PR for this.

PR #2137, opened 2023-10-11
Allow non-string error message to be reported to sentry

the-spectator:
Ref: https://github.com/getsentry/sentry-ruby/issues/907#issuecomment-1752918307

## Description
When the def message in the error class is overridden to return a non-string value. Sentry is not able to report the error due to the error `NoMethodError: undefined method `byteslice' for []:Array`. Ideally, we shouldn't override the message and yet it was done 😓 .

Describe your changes:
Convert non-string error message to string.


natikgadzhi (approved):
Not related to the core team at Sentry, but the change looks reasonable to me. It's better to send out a longer, slightly awkward error message than to error out.

sentry-ruby/spec/sentry/client_spec.rb
```diff
@@ -209,6 +209,22 @@ def sentry_context
           expect(hash[:exception][:values][0][:value]).to eq("undefined method `[]' for nil:NilClass")
         end
       end
+
+      it "converts non-string error message" do
+        NonStringMessageError = Class.new(StandardError) do
+          def message
+            { foo: "bar" }
+          end
```
st0012, line 217: I think this is not needed because only `detailed_message` will be called?

the-spectator: ahh. good catch. let me fix this.

st0012 (approved):
Thank you for the fix. Can you also add a changelog entry for it?

the-spectator:
Thank you for the review! 🙏 
Added the changelog.

PR #2140, opened 2023-10-14
Drop sidekiq patch from the Gemfile

st0012:
The fix we were waiting for has been released in 7.1.6.

#skip-changelog


sl0thentr0py (approved):

st0012:
The test failures are all caused by `psych` with JRuby. So I'm merging this now and deal with them later.

PR #2148, opened 2023-10-19
add delayed job context into the sampling context

rlineweaver:
## Description

This makes additional job context available in the `sampling_context` via the Sentry transaction's `custom_sampling_context`, which makes it available for making sampling rate determinations in `traces_sampler`.  In our case, we want to base the sampling rate off of the job's priority.  This PR allows you to write code like:

```ruby
    config.traces_sampler = lambda do |sampling_context|
        sample_rate = 1

        if sampling_context[:transaction_context][:op] == Sentry::DelayedJob::Plugin::OP_NAME
           job_priority = sampling_context.dig(Sentry::DelayedJob::Plugin::DELAYED_JOB_CONTEXT_KEY, :priority)
           if job_priority && job_priority >= 99
              sample_rate = 0.01
           end
        end

        sample_rate
   end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased
 
+### Features
+
+- Make additional job context available to traces_sampler for determining sample rate (sentry-delayed_job) [#2148](https://github.com/getsentry/sentry-ruby/pull/2148)
+
 ### Bug Fixes
 
 - Fix puma integration for versions before v5 ([#2141](https://github.com/getsentry/sentry-ruby/pull/2141))
--- a/sentry-delayed_job/lib/sentry/delayed_job/plugin.rb
+++ b/sentry-delayed_job/lib/sentry/delayed_job/plugin.rb
@@ -20,7 +20,12 @@ module Sentry
             scope.set_contexts(**contexts)
             scope.set_tags("delayed_job.queue" => job.queue, "delayed_job.id" => job.id.to_s)
 
-            transaction = Sentry.start_transaction(name: scope.transaction_name, source: scope.transaction_source, op: OP_NAME)
+            transaction = Sentry.start_transaction(
+              name: scope.transaction_name,
+              source: scope.transaction_source,
+              op: OP_NAME,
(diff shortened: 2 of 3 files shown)


booleanbetrayal:
Agree that this could be super useful!

booleanbetrayal (approved):
Code changes LGTM and everything has been tested!

st0012:
@sl0thentr0py In other SDKs, is there a format to add job contexts as sampling context? Perhaps we should adopt this in other integrations as well? If so, then we should probably add `delayed_job` as the key here to distinguish context from different jobs given that some apps do use multiple background job libraries.

sl0thentr0py:
@st0012 yes let's put it under `delayed_job`, this is celery in python for instance
https://github.com/getsentry/sentry-python/blob/f570a9966252920bdb221101d596eb029497b0e9/sentry_sdk/integrations/celery.py#L247-L254

rlineweaver:
> @st0012 yes let's put it under `delayed_job`

Hi @st0012 , @sl0thentr0py , thanks for looking at this.

There are at most two keys being set in the `custom_sampling_context` here: [`:"Delayed-Job"` and `:"Active-Job"`](https://github.com/getsentry/sentry-ruby/blob/master/sentry-delayed_job/lib/sentry/delayed_job/plugin.rb#L8-L9) (the value of each is [another hash containing details](https://github.com/getsentry/sentry-ruby/blob/master/sentry-delayed_job/lib/sentry/delayed_job/plugin.rb#L42-L62) about the job).  So I think these keys are already namespaced well enough to avoid confusion/collisions.  I'd be happy to rename those keys to something else (like `delayed_job`), but I reused the existing keys for consistency with the scope's context.

sl0thentr0py:
ok fine with me then!

st0012 (approved):
Ah I see. I now regret defining those keys with camelcase + dash. But I guess it's too late to change 🤦‍♂️

st0012:
@rlineweaver Thank you for this PR 🙌

PR #2151, opened 2023-10-25
Enable opting out of patches

### LINKED ISSUE #2095: Allow opting out of patches
see original discussion here
https://github.com/getsentry/sentry-ruby/issues/2080#issuecomment-1660305576

sl0thentr0py:
closes #2095

### DIFF
--- a/sentry-ruby/spec/sentry_spec.rb
+++ b/sentry-ruby/spec/sentry_spec.rb
@@ -1032,4 +1032,56 @@ RSpec.describe Sentry do
       end.to change { new_transport.events.count }.by(1)
     end
   end
+
+  describe "patching" do
+    let(:target_class) { Class.new }
+
+    let(:module_patch) do
+      Module.new do
+        def foo; end
+      end
+    end
+
+    context "with target and patch to prepend" do
+      before do
+        described_class.register_patch(:module_patch, module_patch, target_class)
+      end
+
+      it "does not prepend patch if not in enabled_patches" do
+        perform_basic_setup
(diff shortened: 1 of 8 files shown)


sentry-ruby/spec/sentry_spec.rb
```diff
@@ -1032,4 +1032,53 @@
...
+          def foo; end
+        end
+      end
+
+      before do
+        described_class.register_patch(:module_patch, module_patch, target_class)
+      end
+
+      it "does not prepend patch if not in enabled_patches" do
+        expect(target_class).not_to receive(:prepend)
```
st0012, line 1051: Since we already checks the result, I don't think this is necessary?

sentry-ruby/spec/sentry_spec.rb
```diff
@@ -1032,4 +1032,53 @@
...
+        expect(target_class.ancestors.first).to eq(module_patch)
+        expect(target_class.instance_methods).to include(:foo)
+      end
+    end
+
+    context "with block" do
+      let(:dbl) { double("block") }
+
+      before do
+        described_class.register_patch(:block_patch) { dbl.foo }
```
st0012, line 1070: Can we do `target_class.send(:prepend, module_patch)`, verify the result by checking `instance_methods`, and not use the double?

I think by aligning the testing styles they'll be easier to understand.

st0012 (commented):
I think it works well 👍 Just suggestions on tests.

st0012 (approved):
👍

PR #2160, opened 2023-11-03
(sentry-sidekiq): Fixed a deprecation warning in error handler

### LINKED ISSUE #2157: Deprecation warning in sentry-sidekiq: Sidekiq exception handlers now take three arguments
### Issue Description

sentry-sidekiq triggers the following deprecation warning from sidekiq: ` Sidekiq exception handlers now take three arguments`.

Context:
 * Changelog from sidekiq: https://github.com/sidekiq/sidekiq/blob/main/Changes.md#715
 * https://github.com/appsignal/appsignal-ruby/commit/c8698dca465d84fdac33d88debc6fbb004458bf1 seems to fix this bug for appsignal.

### Reproduction Steps

Use sidekiq 7.1.6

### Expected Behavior

Don't trigger deprecation warning.

### Actual Behavior

Triggers deprecation warning.

### Ruby Version

3.2.2

### SDK Version

5.12.0

Sidekiq 7.1.6

### Sentry Config

_No response_

natikgadzhi:
## Summary

This PR fixes the deprecation warning that occurs when using Sidekiq >= 7.1.5: #2157. Closes #2157.

## Changes

I'm just getting started, so I documented some steps as I was onboarding: 
- Added more instructions to `CONTRIBUTING.md`, and bumped Rails version there.
- Bumped default Sidekiq version to 7.0 in `sentry-sidekiq` Gemfile.
- Fixed the deprecation warning in the error handler and added a comment there.
- Also added a changelog entry linking to this pull request.

## Next steps

We could in theory use some of those configuration values to enrich the reported errors: https://github.com/sidekiq/sidekiq/blob/main/lib/sidekiq/config.rb, but, they are not specific to the error at hand.

### DIFF
--- a/sentry-sidekiq/lib/sentry/sidekiq/error_handler.rb
+++ b/sentry-sidekiq/lib/sentry/sidekiq/error_handler.rb
@@ -7,11 +7,11 @@ module Sentry
 
       # @param ex [Exception] the exception / error that occured
       # @param context [Hash or Array] Sidekiq error context
-      # @param sidekiq_config [Sidekiq::Config] Sidekiq configuration,
-      #   defaults to Sidekiq's default configuration `Sidekiq.default_configuration`
+      # @param sidekiq_config [Sidekiq::Config, Hash] Sidekiq configuration,
+      #   Defaults to nil.
       #   Sidekiq will pass the config in starting Sidekiq 7.1.5, see
       #   https://github.com/sidekiq/sidekiq/pull/6051
-      def call(ex, context, sidekiq_config = ::Sidekiq.default_configuration)
+      def call(ex, context, sidekiq_config = nil)
         return unless Sentry.initialized?
 
         context_filter = Sentry::Sidekiq::ContextFilter.new(context)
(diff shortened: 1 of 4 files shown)


sentry-sidekiq/lib/sentry/sidekiq/error_handler.rb
```diff
@@ -5,7 +5,13 @@ module Sidekiq
...
       WITH_SIDEKIQ_7 = ::Gem::Version.new(::Sidekiq::VERSION) >= ::Gem::Version.new("7.0")
 
-      def call(ex, context)
+      # @param ex [Exception] the exception / error that occured
+      # @param context [Hash or Array] Sidekiq error context
+      # @param sidekiq_config [Sidekiq::Config] Sidekiq configuration,
+      #   defaults to Sidekiq's default configuration `Sidekiq.default_configuration`
+      #   Sidekiq will pass the config in starting Sidekiq 7.1.5, see
+      #   https://github.com/sidekiq/sidekiq/pull/6051
+      def call(ex, context, sidekiq_config = ::Sidekiq.default_configuration)
```
st0012, line 14: I think `::Sidekiq.default_configuration` was introduced in 7.0 so this will break older Sidekiq versions. Also, we need to update `retry_limit` to use the passed `sidekiq_config` if its present.

natikgadzhi: Hmmm. You're right — I think then we could set it to `nil` by default, but if it's present and has `retry_limit`, use that. I'll fix this up.

natikgadzhi: Thank you again for pointing this out, and for the deferred cleanup of the env var. I think I got to it, not sure if my code is the most elegant.

natikgadzhi:
@st0012 that one should be good to go now, I think.

sl0thentr0py (approved):
ty @natikgadzhi !

rafaelsales:
@sl0thentr0py  When will this be released?

sl0thentr0py:
@rafaelsales this was already released in https://github.com/getsentry/sentry-ruby/releases/tag/5.14.0

PR #2161, opened 2023-11-03
Allow proxy configuration via HTTP_PROXY env var

### LINKED ISSUE #2131: Sentry ignore proxy system configuration
### Issue Description

Sentry Ruby does not appear to utilize the HTTP_PROXY, HTTPS_PROXY, and NO_PROXY environment variables when a system configures to use a proxy. This issue is causing Sentry to be unable to communicate with Sentry.io

### Reproduction Steps

Our system has been set up with the following environment variables to enable proxy usage:

```
HTTPS_PROXY=http://proxy.internal:3128
no_proxy=localhost
NO_PROXY=localhost
https_proxy=http://proxy.internal:3128
http_proxy=http://proxy.internal:3128
HTTP_PROXY=http://proxy.internal:3128
```

When I attempted to validate the Sentry execution, I used the following code:

```
require 'sentry-ruby'

Sentry.init do |config|
  config.dsn = 'https://xxxx@yyyy.ingest.sentry.io/zzzz'

  config.debug = true

  # Sync events
  config.background_worker_threads = 0

  # or
    0.5
  end
end

Sentry.capture_message("test message")
```

I encountered the following error:

```
Event capturing failed: Failed to open TCP connection to yyyy.ingest.sentry.io:443 (execution expired)
```

Nevertheless, when I execute:

```
uri = URI('https://xxxx@yyyy.ingest.sentry.io/zzzz')
http = Net::HTTP.new(uri.host, uri.port)
http.use_ssl = true
http.get(uri.request_uri)
```

I'm able to connect to Sentry:
```
=> #<Net::HTTPNotFound 404 Not Found readbody=true>
```

So, to be able to connect Sentry I explicitly need to define `config.transport.proxy` on the initializer:

```
require 'sentry-ruby'

Sentry.init do |config|
  config.dsn = 'https://xxxx@yyyy.ingest.sentry.io/zzzz'

  config.debug = true
  config.transport.proxy = "http://proxy.internal:3128"

  # Sync events
  config.background_worker_threads = 0

  # or
    0.5
  end
end

Sentry.capture_message("test message")
```

### Expected Behavior

When the HTTP_PROXY, HTTPS_PROXY, and NO_PROXY environment variables are defined, Sentry Ruby should use these settings to route its network requests through the configured proxy,  without having to specify the proxy configuration using `config.transport.proxy`

### Actual Behavior

Currently, Sentry Ruby appears to ignore the HTTP_PROXY, HTTPS_PROXY, and NO_PROXY environment variables, which results in Sentry's inability to communicate with Sentry.io due to proxy-related issues.

### Ruby Version

3.2.0

### SDK Version

5.11.0

_No response_

### Sentry Config

_No response_

natikgadzhi:
## Summary

This PR allows `sentry-ruby` to use proxy config passed via environment variables, i.e. `HTTP_PROXY`. Closes #2131. The actual configuration settings is still respected over environment variables.

## Changes
- As @sl0thentr0py suggests, just removes the `nil` passed into `::Net::HTTP`, and allows it to do all the work for us.
- Adds a changelog entry. 

## TODOs
- [x] Add unit tests to https://github.com/getsentry/sentry-ruby/blob/c9b4eac913f0c6510178c451da57d91fc340fa20/sentry-ruby/spec/sentry/transport/http_transport_spec.rb#L94
- [x] I'll add a separate PR into `sentry-docs` to document how proxy config works. As of right now, `transport_configuration` block there is minimal. 
- [ ] Once this is shipped, further update `sentry-docs` with proxy via env var support.

## Review
- @stephanie-anderson, with this PR, `config.transport_configuration.proxy` is preferred over the environment values:
  - If the `config.transport_configuration.proxy` is present, but malformed (does not have the minimally required uri key), it will still be used over `HTTP_PROXY`, even if that has the correct setup. 
  - We should probably be consistent with how our other SDKs work with env vars — should we generally prefer the config passed in code, or the env vars?

### DIFF
--- a/sentry-ruby/lib/sentry/transport/http_transport.rb
+++ b/sentry-ruby/lib/sentry/transport/http_transport.rb
@@ -128,12 +128,15 @@ module Sentry
 
     def conn
       server = URI(@dsn.server)
-
+      
+      # connection respects proxy setting from @transport_configuration, or environment variables (HTTP_PROXY, HTTPS_PROXY, NO_PROXY)
+      # Net::HTTP will automatically read the env vars.
+      # See https://ruby-doc.org/3.2.2/stdlibs/net/Net/HTTP.html#class-Net::HTTP-label-Proxies
       connection =
         if proxy = normalize_proxy(@transport_configuration.proxy)
           ::Net::HTTP.new(server.hostname, server.port, proxy[:uri].hostname, proxy[:uri].port, proxy[:user], proxy[:password])
         else
-          ::Net::HTTP.new(server.hostname, server.port, nil)
+          ::Net::HTTP.new(server.hostname, server.port)
         end
 
       connection.use_ssl = server.scheme == "https"
@@ -148,6 +151,9 @@ module Sentry
       connection
     end
 
+    # @param proxy [String, URI, Hash] Proxy config value passed into `config.transport`.
+    #   Accepts either a URI formatted string, URI, or a hash with the `uri`, `user`, and `password` keys.
+    # @return [Hash] Normalized proxy config that will be passed into `Net::HTTP`
     def normalize_proxy(proxy)
       return proxy unless proxy
 
--- a/sentry-ruby/spec/sentry/transport/http_transport_spec.rb
+++ b/sentry-ruby/spec/sentry/transport/http_transport_spec.rb
@@ -63,7 +63,7 @@ RSpec.describe Sentry::HTTPTransport do
         expect(subject.send(:conn).port).to eq(80)
       end
     end
-    context "with http DSN" do
+    context "with https DSN" do
       let(:dsn) { "https://12345:67890@sentry.localdomain/sentry/42" }
 
       it "sets port to 443" do
@@ -127,6 +127,21 @@ RSpec.describe Sentry::HTTPTransport do
       end
 
       subject.send_data(data)
+    end
+
+    it "accepts a proxy from ENV[HTTP_PROXY]" do
+      ENV["http_proxy"] = "https://stan:foobar@example.com:8080"
+
+      stub_request(fake_response) do |_, http_obj|
+        expect(http_obj.proxy_address).to eq("example.com")
+        expect(http_obj.proxy_user).to eq("stan")
+        expect(http_obj.proxy_pass).to eq("foobar")
+        expect(http_obj.proxy_port).to eq(8080)
+      end
+
+      subject.send_data(data)
+
+      ENV["http_proxy"] = nil 
     end
 
     it "accepts custom timeout" do
--- a/sentry-ruby/spec/sentry/transport/http_transport_spec.rb
+++ b/sentry-ruby/spec/sentry/transport/http_transport_spec.rb
@@ -130,18 +130,20 @@ RSpec.describe Sentry::HTTPTransport do
     end
 
     it "accepts a proxy from ENV[HTTP_PROXY]" do
-      ENV["http_proxy"] = "https://stan:foobar@example.com:8080"
-
-      stub_request(fake_response) do |_, http_obj|
-        expect(http_obj.proxy_address).to eq("example.com")
-        expect(http_obj.proxy_user).to eq("stan")
-        expect(http_obj.proxy_pass).to eq("foobar")
-        expect(http_obj.proxy_port).to eq(8080)
+      begin
+        ENV["http_proxy"] = "https://stan:foobar@example.com:8080"
+  
+        stub_request(fake_response) do |_, http_obj|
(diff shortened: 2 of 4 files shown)


sentry-ruby/spec/sentry/transport/http_transport_spec.rb
```diff
@@ -129,6 +129,21 @@
...
+      stub_request(fake_response) do |_, http_obj|
+        expect(http_obj.proxy_address).to eq("example.com")
+        expect(http_obj.proxy_user).to eq("stan")
+        expect(http_obj.proxy_pass).to eq("foobar")
+        expect(http_obj.proxy_port).to eq(8080)
+      end
+
+      subject.send_data(data)
+
+      ENV["http_proxy"] = nil
```
st0012, line 144: This need to be put inside an ensure block so when it failed the env wouldn't leak to other tests.

```suggestion
      begin
        ENV["http_proxy"] = "https://stan:foobar@example.com:8080"
  
        stub_request(fake_response) do |_, http_obj|
          expect(http_obj.proxy_address).to eq("example.com")
          expect(http_obj.proxy_user).to eq("stan")
          expect(http_obj.proxy_pass).to eq("foobar")
          expect(http_obj.proxy_port).to eq(8080)
        end
  
        subject.send_data(data)
      ensure
        ENV["http_proxy"] = nil
      end
```

sentry-ruby/spec/sentry/transport/http_transport_spec.rb
```diff
@@ -63,7 +63,7 @@
         expect(subject.send(:conn).port).to eq(80)
       end
     end
-    context "with http DSN" do
+    context "with https DSN" do
```
st0012, line 66: Thx 👍

sentry-ruby/lib/sentry/transport/http_transport.rb
```diff
@@ -148,6 +151,9 @@ def conn
       connection
     end
 
+    # @param proxy [String, URI, Hash] Proxy config value passed into `config.transport`.
+    #   Accepts either a URI formatted string, URI, or a hash with the `uri`, `user`, and `password` keys.
+    # @return [Hash] Normalized proxy config that will be passed into `Net::HTTP`
```
st0012, line 156: I don't think it's necessary to document this method with YARD as it's internal to the SDK.

Maybe it's better to document [`Transport::Configuration#proxy`](https://github.com/getsentry/sentry-ruby/blob/master/sentry-ruby/lib/sentry/transport/configuration.rb#L6)instead?

natikgadzhi:
Curiously, it fails for Ruby 2.4: https://github.com/getsentry/sentry-ruby/actions/runs/6755752783/job/18365644960?pr=2161#step:6:1523 I'll try it out locally and make a workaround.

sentry-ruby/lib/sentry/transport/http_transport.rb
natikgadzhi: Yep, seemed like transport configuration could benefit from some documentation: https://github.com/getsentry/sentry-ruby/pull/2161/commits/d97e181c182aeb4750481a3ce343d9002f488e17

natikgadzhi:
I'm still on this — proved to be difficult to get to locally (2.4 doesn't easily install on M1 with `rbenv`, so I'll docker this tomorrow). I have no clue why this works with Ruby 2.4 on Rack 2, but not on Ruby 2.4 and Rack 3 =/

natikgadzhi:
@sl0thentr0py if you're in the modd to merge things — can you help me out with this one? I was hitting a wall on that Ruby 2.4 test.

sl0thentr0py:
i'll look tomorrow, irl now

st0012 (approved):
👍 Sorry for the delay

st0012:
The CI fails because Ruby didn't support getting user/pass from `http_proxy` until 2.5 with [this commit](https://github.com/ruby/ruby/commit/c7ec1b1f599de0297bbf2cba92a923a337df73a6). I think you need to add Ruby version condition on both `proxy_user` and `proxy_pass` assertions.

natikgadzhi:
@st0012 TIL. Fixed!

st0012:
Thank you for the fix as well as the new documentation!!

natikgadzhi:
I'll make a follow up docs PR when we ship this out.

natikgadzhi:
Since this is on the client SDK side, and it already shipped in 5.14, just update your sentry-ruby, it should already work against both SaaS, and a self-hosted Sentry. 


This patch only controls whether the client SDK will use a proxy to send data to the server. Point it to a self-hosted DSN and it should work.

PR #2169, opened 2023-11-13
Avoid invoking ActiveSupport::BroadcastLogger if not defined

PetrHeinz:
## Description

When `::ActiveSupport::BroadcastLogger` is not defined (Rails < 7.1) and `::Rails.logger` instance has `broadcasts` method defined, initialization of Sentry-ruby fails on `NameError: uninitialized constant ActiveSupport::BroadcastLogger (NameError)`.

I suggest changing the check to `::Rails.logger` being an instance of `::ActiveSupport::BroadcastLogger` (with a `defined?` check).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Bug Fixes
 
 - Fixed a deprecation in `sidekiq-ruby` error handler [#2160](https://github.com/getsentry/sentry-ruby/pull/2160)
+- Avoid instantiating `BroadcastLogger` in `sidekiq-ruby` if class not defined [#2169](https://github.com/getsentry/sentry-ruby/pull/2169)
 
 ## 5.13.0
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### Bug Fixes
 
 - Fixed a deprecation in `sidekiq-ruby` error handler [#2160](https://github.com/getsentry/sentry-ruby/pull/2160)
+- Avoid instantiating `BroadcastLogger` in `sidekiq-ruby` if class not defined [#2169](https://github.com/getsentry/sentry-ruby/pull/2169)
```
st0012, line 6:
```suggestion
- Avoid invoking ActiveSupport::BroadcastLogger if not defined [#2169](https://github.com/getsentry/sentry-ruby/pull/2169)
```

st0012 (approved):
I wasn't aware the other loggers could also have `braodcasts` method too. Thanks for the fix 👍

Issue #2134, opened 2023-10-10
Crons automatic support for gems


sl0thentr0py:
- sidekiq-cron
- sidekiq-scheduler
- ~whenever~ bash based, so just use `sentry-cli`
- ~clockwork~ also unaware of job system, so we can't really patch

_Originally posted by @sl0thentr0py in https://github.com/getsentry/sentry-ruby/issues/2090#issuecomment-1737351435_

natikgadzhi:
Cool if I work on `sidekiq-scheduler`? Using #2170 as a reference, that's pretty straightforward.

natikgadzhi:
Looking at [Whenever]https://github.com/javan/whenever#define-your-own-job-types, it's a bit funny:

> All jobs are by default run with bash -l -c 'command...'. Among other things, this allows your cron jobs to play nice with RVM by loading the entire environment instead of cron's somewhat limited environment. Read more: http://blog.scoutapp.com/articles/2010/09/07/rvm-and-cron-in-production
> 
> You can change this by setting your own :job_template.

I think we can make our own wrapper that: 
1. Works as a rake task, and therefore loads Sentry configuration from the existing Rails app
2. Starts, loads Sentry config, reports monitor check-in for the invoked command.
3. Executes the actual command by shelling out
4. If successful, reports successful monitor check-in.

I wonder if this should be configurable, so the user could opt-out of that behavior? I have a sneaky suspicion performance of starting a rake task only to ping Sentry, and then start another wrapped rake task could be not entirely great.

sl0thentr0py:
@natikgadzhi  hmm if it's bash based, I'd leave that out of the scope of this task.
We already have CLI support here, so then they can just [wrap their runs in `sentry-cli`](https://docs.sentry.io/product/crons/getting-started/cli/).

sl0thentr0py:
clockwork is also fully unaware of what scheduling system is in use, so we can't do much there either, let's just ship the sidekiq stuff for now.

natikgadzhi:
I looked into Clockwork yesterday and tried to prototype it. 

1. We can setup the error handler at least, relatively easy. Although it’s just three lines for the user to setup themselves. 
2. We could try and hook into their execute method that, in theory, wraps execution of _block_ tasks, but some tasks are allowed to be passed as _strings_ and I have no clue what the hell is going on. If it doesn’t look clear to me and you after taking a look, I doubt whatever integration we can do will be clear for users, and won’t get in their way. 

[...]

PR #2170, opened 2023-11-14
Add `sidekiq-cron` patch for automatic monitoring of jobs listed in the schedule

sl0thentr0py:
* optional patch under `sidekiq_cron`
 * patch the [`Sidekiq::Cron::Job#save`](https://github.com/sidekiq-cron/sidekiq-cron/blob/9823b8e741e6be877eb5971fd18966c225d08c2f/lib/sidekiq/cron/job.rb#L464) method and auto inject the `Sentry::MonitorCheckIns` module and turn monitoring on

part of #2134

### DIFF
--- a/sentry-sidekiq/Gemfile
+++ b/sentry-sidekiq/Gemfile
@@ -23,6 +23,7 @@ sidekiq_version = "7.0" if sidekiq_version.nil?
 sidekiq_version = Gem::Version.new(sidekiq_version)
 
 gem "sidekiq", "~> #{sidekiq_version}"
+gem "sidekiq-cron" if sidekiq_version >= Gem::Version.new("6.0")
 
 gem "rails", "> 5.0.0", "< 7.1.0"
 
--- a/sentry-sidekiq/lib/sentry/sidekiq/cron/job.rb
+++ b/sentry-sidekiq/lib/sentry/sidekiq/cron/job.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+return unless defined?(::Sidekiq::Cron::Job)
+
+module Sentry
+  module Sidekiq
+    module Cron
+      module Job
+        def save
+          # validation failed, do nothing
+          return false unless super
+
+          # fail gracefully if can't find class
+          klass_const =
+            begin
+              ::Sidekiq::Cron::Support.constantize(klass.to_s)
+            rescue NameError
+              return true
+            end
+
+          # only patch if not explicitly included in job by user
+          unless klass_const.send(:ancestors).include?(Sentry::Cron::MonitorCheckIns)
+            klass_const.send(:include, Sentry::Cron::MonitorCheckIns)
+            klass_const.send(:sentry_monitor_check_ins,
+                             slug: name,
+                             monitor_config: Sentry::Cron::MonitorConfig.from_crontab(cron))
+          end
+        end
+      end
+    end
+  end
+end
+
+Sentry.register_patch(:sidekiq_cron, Sentry::Sidekiq::Cron::Job, ::Sidekiq::Cron::Job)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@
 ### Features
 
 - Improve default slug generation for Crons [#2168](https://github.com/getsentry/sentry-ruby/pull/2168)
+- Automatic Crons support for scheduling gems
+  - Add support for [`sidekiq-cron`](https://github.com/sidekiq-cron/sidekiq-cron) [#2170](https://github.com/getsentry/sentry-ruby/pull/2170)
+
+    You can opt in to the `sidekiq-cron` patch and we will automatically monitor check-ins for all jobs listed in your `config/schedule.yml` file.
+
+    ```rb
+    config.enabled_patches += [:sidekiq_cron]
+    ```
 
 ### Bug Fixes
 
--- a/sentry-sidekiq/lib/sentry-sidekiq.rb
+++ b/sentry-sidekiq/lib/sentry-sidekiq.rb
@@ -39,3 +39,6 @@ Sidekiq.configure_client do |config|
     chain.add Sentry::Sidekiq::SentryContextClientMiddleware
   end
 end
+
+# patches
+require "sentry/sidekiq/cron/job"
--- a/sentry-sidekiq/spec/fixtures/schedule.yml
+++ b/sentry-sidekiq/spec/fixtures/schedule.yml
@@ -1,0 +1,11 @@
+happy:
+  cron: "* * * * *"
+  class: "HappyWorkerDup"
+
+manual:
+  cron: "* * * * *"
+  class: "SadWorkerWithCron"
+
+invalid_cron:
+  cron: "not a crontab"
+  class: "ReportingWorker"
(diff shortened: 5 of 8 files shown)


adinauer (approved):
LGTM if `perform` is actually being called for sidekiq which I assume you tested.

sentry-sidekiq/Gemfile
```diff
@@ -23,6 +23,7 @@ sidekiq_version = "7.0" if sidekiq_version.nil?
 sidekiq_version = Gem::Version.new(sidekiq_version)
 
 gem "sidekiq", "~> #{sidekiq_version}"
+gem "sidekiq-cron" if sidekiq_version >= Gem::Version.new("6.0")
```
natikgadzhi, line 26: If I read this correctly, this would bring `sidekiq-cron` as a dependency for `sentry-sidekiq`. Do we want to go with a similar path for other job schedulers, such as whenever, or do we want to keep them optional? 

Hypothetically, we could wrap the code that works with them in a check that verifies that they are indeed available, but I don't know if all that ceremony is better or worse than having several dependencies.

sentry-sidekiq/lib/sentry/sidekiq/cron/job.rb
```diff
@@ -0,0 +1,34 @@
...
+          klass_const =
+            begin
+              ::Sidekiq::Cron::Support.constantize(klass.to_s)
+            rescue NameError
+              return true
+            end
+
+          # only patch if not explicitly included in job by user
+          unless klass_const.send(:ancestors).include?(Sentry::Cron::MonitorCheckIns)
+            klass_const.send(:include, Sentry::Cron::MonitorCheckIns)
```
natikgadzhi, line 23: Would this make it so any invocation of the related Job class, including invocations triggered by `sidekiq-cron`, and invocations that the user did manually, with just `HappyWorkerJob.perform_async`? It looks like it would, right?

I wonder if there are situations when users want to kick the same job for a one-off thing that they don't want to be reported to Sentry. I, personally, think that since the defined `sidekiq-cron` job just wraps a `Sidekiq` job, than whenever that job was triggered, it should be reported. 

Do we want to add this to docs? Is this behavior consistent with other languages and frameworks Sentry supports?

/cc @sl0thentr0py

natikgadzhi (commented):
Don't mind me, I was just catching up.

sentry-sidekiq/Gemfile
sl0thentr0py: this is a dependency on _our_ side for testing, user dependencies are only these two 
https://github.com/getsentry/sentry-ruby/blob/c7c8e5b1495692aed987b43fb1e365101d8f8961/sentry-sidekiq/sentry-sidekiq.gemspec#L25-L26

sentry-sidekiq/lib/sentry/sidekiq/cron/job.rb
sl0thentr0py: I will add docs for all of this around releasing. 

> Would this make it so any invocation of the related Job class, including invocations triggered by sidekiq-cron, and invocations that the user did manually, with just HappyWorkerJob.perform_async? It looks like it would, right?

yes

> I wonder if there are situations when users want to kick the same job for a one-off thing that they don't want to be reported to Sentry. I, personally, think that since the defined sidekiq-cron job just wraps a Sidekiq job, than whenever that job was triggered, it should be reported.

This is a new feature, we will add special config once people use it and request it. This is fine as a first version, it's an opt-in patch either way.

sentry-sidekiq/lib/sentry/sidekiq/cron/job.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+return unless defined?(::Sidekiq::Cron::Job)
```
st0012, line 3: Since there's no guarantee that the SDK will be loaded AFTER `sidekiq-cron` is required, I think we need to require it here and rescue `LoadError` and return if it fails to load.

sentry-sidekiq/lib/sentry/sidekiq/cron/job.rb
st0012: It looks like there could be multiple ways to invoke `save` in addition to the initial creation. Although the current implementation doesn't seem to have obvious performance impact, I feel a better place to inject such one-off activation logic is `load_from_array` instead.
I understand that it's an experimental feature for now, so I'm not suggesting an immediate refactor. Just want to point it out for future improvements.

sentry-sidekiq/lib/sentry/sidekiq/cron/job.rb
```diff
@@ -0,0 +1,34 @@
...
+              return true
+            end
+
+          # only patch if not explicitly included in job by user
+          unless klass_const.send(:ancestors).include?(Sentry::Cron::MonitorCheckIns)
+            klass_const.send(:include, Sentry::Cron::MonitorCheckIns)
+            klass_const.send(:sentry_monitor_check_ins,
+                             slug: name,
+                             monitor_config: Sentry::Cron::MonitorConfig.from_crontab(cron))
+          end
```
st0012, line 27: Should we return `true` in the end?

st0012 (commented):
Sorry for the long delay. I was preparing for and attending RubyConf in the US.

sentry-sidekiq/lib/sentry/sidekiq/cron/job.rb
sl0thentr0py: if people are scheduling cron jobs with `Sidekiq::Cron::Job.create`, we patch those too automatically, either way shouldn't matter, we can react to feedback from users later depending on usage.

sentry-sidekiq/lib/sentry/sidekiq/cron/job.rb
sl0thentr0py: the upstream method doesn't return true either, just the falsey return is important.

st0012: But the upstream's last line is [`Sidekiq.logger.info`](https://github.com/sidekiq-cron/sidekiq-cron/blob/master/lib/sidekiq/cron/job.rb#L481), which returns `true`

```
> ::Sidekiq.logger.info(123)
=> true
```

And `unless ...` returns `nil`. So this patch actually changes the code's behaviour here:

```
errors[job.name] = job.errors unless job.save
```

sentry-sidekiq/lib/sentry/sidekiq/cron/job.rb
natikgadzhi: Honest question so I can draft the docs PR: 
- to enable `sidekiq-cron` and/or `sidekiq-scheduler` monitoring, the users will have to manually add `config.enabled_patches += [:sidekiq_cron]`, right? 
- When this patch is applied, and as long as `sidekiq-cron` is indeed available, we aim to instrument all jobs that it runs, all the time. The current logic that hooks into `save` does that. Correct?
- One side-iffect is that we're instrumenting the job class itself, so whether it was the `sidekiq-cron` schedule that invoked the job on a timer, or the user's code invoked `perform_*` on that job, we will monitor that too. That's by design for now, but we'll keep an eye out for feedback. Right?

sl0thentr0py: yes to all

natikgadzhi:
Yep. It doesn’t care where the schedule comes from.

PR #2172, opened 2023-11-20
Crons: add `sidekiq-scheduler` zero-config monitor check-ins

natikgadzhi:
## Summary

This pull request adds Crons support for [`sidekiq-scheduler`](https://github.com/sidekiq-scheduler/sidekiq-scheduler/) jobs with zero configuration. 

- Part of #2134
- Based on #2170 approach by @sl0thentr0py

## Changes

- 3543bec8 First commit moves `sidekiq-cron` instrumentation to have a clearer naming. I'm not attached to this — okay to revert if you don't feel like this is needed. I think it's nice when directory structure helps you understand what we're instrumenting specifically, but I don't know for absolutely certain if this will work nicely with zeitwerk?
- 0b495368 implements `sidekiq-scheduler` integration

## How this works

Similar to #2170, when `SidekiqScheduler` loads the jobs, we hook into it's `new_job` method, constantine the job class, and patch it with `MonitorCheckIns` module.

## Open questions

- [x] Is it okay to have both `sidekiq-cron` and `sidekiq-scheduler` in our `Gemfile`, and require both of them in `spec_helper` for `sentry-sidekiq`? I feel like folks could be surprised that Sentry tugs two extra dependencies, one of which they probably don't need at all.

  I don't know how to best work around this, though — if we drop the dependencies and wrap the `require` calls in error handling, I don't know if Bundler will allow us require anything from the global scope that is not a part of the bundle, right? So... 👀
- [x] Add a changelog entry.

### DIFF
--- a/sentry-sidekiq/Gemfile
+++ b/sentry-sidekiq/Gemfile
@@ -24,6 +24,7 @@ sidekiq_version = Gem::Version.new(sidekiq_version)
 
 gem "sidekiq", "~> #{sidekiq_version}"
 gem "sidekiq-cron" if sidekiq_version >= Gem::Version.new("6.0")
+gem "sidekiq-scheduler" if sidekiq_version >= Gem::Version.new("6.0")
 
 gem "rails", "> 5.0.0", "< 7.1.0"
 
--- a/sentry-sidekiq/Gemfile
+++ b/sentry-sidekiq/Gemfile
@@ -24,7 +24,7 @@ sidekiq_version = Gem::Version.new(sidekiq_version)
 
 gem "sidekiq", "~> #{sidekiq_version}"
 
-if sidekiq_version >= Gem::Version.new("6.0")
+if RUBY_VERSION.to_f >= 2.7 && sidekiq_version >= Gem::Version.new("6.0")
   gem "sidekiq-cron"
   gem "sidekiq-scheduler"
 end
--- a/sentry-sidekiq/Gemfile
+++ b/sentry-sidekiq/Gemfile
@@ -23,8 +23,11 @@ sidekiq_version = "7.0" if sidekiq_version.nil?
 sidekiq_version = Gem::Version.new(sidekiq_version)
 
 gem "sidekiq", "~> #{sidekiq_version}"
-gem "sidekiq-cron" if sidekiq_version >= Gem::Version.new("6.0")
-gem "sidekiq-scheduler" if sidekiq_version >= Gem::Version.new("6.0")
+
+if sidekiq_version >= Gem::Version.new("6.0")
+  gem "sidekiq-cron"
+  gem "sidekiq-scheduler"
+end
 
 gem "rails", "> 5.0.0", "< 7.1.0"
 
--- a/sentry-sidekiq/lib/sentry/sidekiq-scheduler/scheduler.rb
+++ b/sentry-sidekiq/lib/sentry/sidekiq-scheduler/scheduler.rb
@@ -10,7 +10,7 @@ module Sentry
         # SidekiqScheduler does not validate schedules
         # It will fail with an error if the schedule in the config is invalid.
         # If this errors out, let it fall through.
-        rufus_job = super(name, interval_type, config, schedule, options)
+        rufus_job = super
 
         # Constantize the job class, and fail gracefully if it could not be found
         klass_const =
@@ -20,7 +20,20 @@ module Sentry
           return rufus_job
         end
 
-        monitor_config = Sentry::Cron::MonitorConfig.from_crontab(schedule)
+        # For cron, every, or interval jobs — grab their schedule.
+        # Rufus::Scheduler::EveryJob stores it's frequency in seconds, 
+        # so we convert it to minutes before passing in to the monitor.
+        monitor_config = case interval_type
+          when "cron"
+            Sentry::Cron::MonitorConfig.from_crontab(schedule)
+          when "every", "interval" 
+            Sentry::Cron::MonitorConfig.from_interval(rufus_job.frequency.to_f / 60.0, :minute)
+        end
+
+        # If we couldn't build a monitor config, it's either an error, or
+        # it's a one-time job (interval_type is in, or at), in which case
+        # we should not make a monitof for it automaticaly.
+        return rufus_job if monitor_config.nil?
 
         # only patch if not explicitly included in job by user
         unless klass_const.send(:ancestors).include?(Sentry::Cron::MonitorCheckIns)
--- a/sentry-sidekiq/lib/sentry/sidekiq-scheduler/scheduler.rb
+++ b/sentry-sidekiq/lib/sentry/sidekiq-scheduler/scheduler.rb
@@ -1,5 +1,14 @@
 # frozen_string_literal: true
 
+# Try to require sidekiq-scheduler to make sure it's loaded before the integration.
+begin
+  require "sidekiq-scheduler"
+rescue LoadError
+  return
+end
+
+# If we've loaded sidekiq-scheduler, but the API changed,
+# and the Scheduler class is not there, fail gracefully.
 return unless defined?(::SidekiqScheduler::Scheduler)
 
 module Sentry
@@ -45,7 +54,7 @@ module Sentry
           ::Sidekiq.logger.info "Injected Sentry Crons monitor checkins into #{config.fetch("class")}"
         end
 
-        return rufus_job
+        rufus_job
       end
     end
   end
--- a/sentry-sidekiq/lib/sentry/sidekiq-scheduler/scheduler.rb
+++ b/sentry-sidekiq/lib/sentry/sidekiq-scheduler/scheduler.rb
@@ -21,22 +21,25 @@ module Sentry
         # If this errors out, let it fall through.
         rufus_job = super
 
+        klass = config.fetch("class")
+        return rufus_job unless klass
+
         # Constantize the job class, and fail gracefully if it could not be found
         klass_const =
-        begin
-          config.fetch("class").constantize
-        rescue NameError
-          return rufus_job
-        end
+          begin
+            Object.const_get(klass)
+          rescue NameError
+            return rufus_job
+          end
 
         # For cron, every, or interval jobs — grab their schedule.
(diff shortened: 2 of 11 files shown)


sl0thentr0py:
thx @natikgadzhi 

pls revert the renames, those are intentional and follow the namespacing of the gem we are patching `Sidekiq::Cron::Job -> Sentry::Sidekiq::Cron::Job`. In general if you want to do changes like that, please open a separate PR and keep features/bug fixes isolated.

I'll review the other stuff now.

sl0thentr0py:
>  Is it okay to have both sidekiq-cron and sidekiq-scheduler in our Gemfile, and require both of them in spec_helper for sentry-sidekiq? I feel like folks could be surprised that Sentry tugs two extra dependencies, one of which they probably don't need at all.

these are our dependencies for testing, not for our users.

sl0thentr0py (changes requested):

natikgadzhi:
> In general if you want to do changes like that, please open a separate PR and keep features/bug fixes isolated.

I even considered a separate PR first, but then settled on a separate commit for that exact reason. No problem, I'm a  noob here, so TIL. Will revert. 

I'll clean the rest of the feedback up in a bit and push up the fix, then work on documentation.

sentry-sidekiq/lib/sentry/sidekiq/sidekiq-scheduler/scheduler.rb
```diff
@@ -0,0 +1,41 @@
...
+
+        # Constantize the job class, and fail gracefully if it could not be found
+        klass_const =
+        begin
+          config.fetch("class").constantize
+        rescue NameError
+          return rufus_job
+        end
+
+        monitor_config = Sentry::Cron::MonitorConfig.from_crontab(schedule)
```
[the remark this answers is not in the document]
natikgadzhi: Added support for this, and working around the situation when some jobs can be one-off.

natikgadzhi:
@sl0thentr0py thank you a LOT for your feedback! I'm learning as I go, and I'm grateful for your patience with me.

I've worked through your suggestions, so this should be ready for another pass. Also added the changelog line.

sentry-sidekiq/Gemfile
```diff
@@ -24,6 +24,7 @@ sidekiq_version = Gem::Version.new(sidekiq_version)
 
 gem "sidekiq", "~> #{sidekiq_version}"
 gem "sidekiq-cron" if sidekiq_version >= Gem::Version.new("6.0")
+gem "sidekiq-scheduler" if sidekiq_version >= Gem::Version.new("6.0")
```
st0012, line 27: Let's merge the 2 under one condition?

sentry-sidekiq/lib/sentry/sidekiq-scheduler/scheduler.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+return unless defined?(::SidekiqScheduler::Scheduler)
```
st0012, line 3: I think we should try require [sidekiq-scheduler](https://github.com/sidekiq-scheduler/sidekiq-scheduler) here? Otherwise, it's not guaranteed that it will be required before `sentry-sidekiq` and then this will return even though the gem is present in Gemfile.

sentry-sidekiq/lib/sentry/sidekiq-scheduler/scheduler.rb
```diff
@@ -0,0 +1,54 @@
...
+        unless klass_const.send(:ancestors).include?(Sentry::Cron::MonitorCheckIns)
+          klass_const.send(:include, Sentry::Cron::MonitorCheckIns)
+          klass_const.send(:sentry_monitor_check_ins,
+                            slug: name,
+                            monitor_config: monitor_config)
+          
+          ::Sidekiq.logger.info "Injected Sentry Crons monitor checkins into #{config.fetch("class")}"
+        end
+
+        return rufus_job
```
st0012, line 48: This return is not necessary.

sentry-sidekiq/lib/sentry/sidekiq-scheduler/scheduler.rb
```diff
@@ -0,0 +1,54 @@
...
+            Sentry::Cron::MonitorConfig.from_interval(rufus_job.frequency.to_f / 60.0, :minute)
+        end
+
+        # If we couldn't build a monitor config, it's either an error, or
+        # it's a one-time job (interval_type is in, or at), in which case
+        # we should not make a monitof for it automaticaly.
+        return rufus_job if monitor_config.nil?
+
+        # only patch if not explicitly included in job by user
+        unless klass_const.send(:ancestors).include?(Sentry::Cron::MonitorCheckIns)
```
st0012, line 39: Is it possible to patch `load_schedules!` to patch all jobs at once instead of going through these conditions repeatedly?

sl0thentr0py: this check is still necessary if some user has included the mixin and overridden the slug themselves

st0012: Actually, I can see that `new_job` is called for new rufus jobs instead of the sidekiq jobs, so it won't be called as often as I thought. Please ignore this.

sentry-sidekiq/lib/sentry/sidekiq-scheduler/scheduler.rb
natikgadzhi: I think the upstream `new_job` returns the `Rufus::Job`. Wouldn't this method return whatever the `unless` above results in instead of the job, if we drop this line? (Just learning here, been a while since I wrote Ruby)

sentry-sidekiq/lib/sentry/sidekiq-scheduler/scheduler.rb
st0012: Yes so I also added a comment there. I think something like this should be enough?

```rb
begin
  require "sidekiq-scheduler"
rescue LoadError
  return
end
```

natikgadzhi:
@st0012 thank you for the review! I'll clean this up later tonight, should be ready tomorrow. 

I also see the Coverage decline, I'll add the test.

sentry-sidekiq/lib/sentry/sidekiq-scheduler/scheduler.rb
st0012: Oh I mean the `return` keyword here is not necessary as it's at the end of the method. We do need to return the same object as you said.

sentry-sidekiq/lib/sentry/sidekiq-scheduler/scheduler.rb
natikgadzhi: @st0012 @sl0thentr0py I'm happy to clean up both, sidekiq-cron in a separate PR to keep things clear.

sentry-sidekiq/lib/sentry/sidekiq-scheduler/scheduler.rb
natikgadzhi: Speaking of — should we perhaps wire up Rubocop and enforce formatting in a pre-commit hook? Or would that be an overkill?

natikgadzhi:
@st0012 @sl0thentr0py welp. Next steps seem to be:
- Code coverage: we _could_ craft tests, but there's not much value to test a scenario where a class name is broken since that would error out upstream anyway.
- Ruby 2.5 doesn't like `delegate`. ;-( I could add those methods implementations I think, but those are also just tests, the actual integration will work fine. Should I clean this up?

sentry-sidekiq/lib/sentry/sidekiq/sidekiq-scheduler/scheduler.rb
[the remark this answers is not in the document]
natikgadzhi: Done!

natikgadzhi:
@sl0thentr0py @st0012 Fixed the Ruby 2.5 compatibility, now the CI is just [trolling me](https://github.com/getsentry/sentry-ruby/actions/runs/6962834221/job/18947320234#step:6:1499).

sl0thentr0py:
re-ran just timestamp flakiness

sentry-sidekiq/lib/sentry/sidekiq-scheduler/scheduler.rb
```diff
@@ -0,0 +1,66 @@
...
+          end
+
+        # For cron, every, or interval jobs — grab their schedule.
+        # Rufus::Scheduler::EveryJob stores it's frequency in seconds,
+        # so we convert it to minutes before passing in to the monitor.
+        monitor_config = case interval_type
+          when "cron"
+            Sentry::Cron::MonitorConfig.from_crontab(schedule)
+          when "every", "interval"
+            Sentry::Cron::MonitorConfig.from_interval(rufus_job.frequency.to_i / 60, :minute)
```
sl0thentr0py, line 42: ok only thing i'm not sure is if the server will accept for instance `300 minutes` and convert that to 5 hours or not? I asked, but we can patch that later

sl0thentr0py (approved):
ok i made some minor changes and style cleanup, see last commit

good to go from my side, also tested, nice work @natikgadzhi !

sentry-sidekiq/lib/sentry/sidekiq-scheduler/scheduler.rb
natikgadzhi: I looked around, and in theory, we can have our own Fugit parser that would parse out the `schedule` string. But we'd have to get fugit as a dependency in. If we merge this, I would be down to test on my own and patch this up if it becomes an issue.

sl0thentr0py: yea we can do this later, no biggie

sl0thentr0py: 300 minutes is fine, server will take care of it

PR #2173, opened 2023-11-22
(fix) crons: require sidekiq-cron early

natikgadzhi:
#skip-changelog
(am I doing this right?)

## Summary

Addressing [this review](https://github.com/getsentry/sentry-ruby/pull/2170#discussion_r1401289145) by @st0012 on `sidekiq-cron` integration — the corresponsing fix for `sidekiq-scheduler` is in #2172.

### DIFF
--- a/sentry-sidekiq/lib/sentry/sidekiq/cron/job.rb
+++ b/sentry-sidekiq/lib/sentry/sidekiq/cron/job.rb
@@ -1,6 +1,12 @@
 # frozen_string_literal: true
 
-return unless defined?(::Sidekiq::Cron::Job)
+# Try requiring sidekiq-cron to ensure it's loaded before the integration.
+# If sidekiq-cron is not available, do nothing.
+begin
+  require "sidekiq-cron"
+rescue LoadError
+  return
+end
 
 module Sentry
   module Sidekiq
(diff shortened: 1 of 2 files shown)


sentry-sidekiq/lib/sentry/sidekiq/cron/job.rb
```diff
@@ -1,6 +1,12 @@
 # frozen_string_literal: true
 
-return unless defined?(::Sidekiq::Cron::Job)
```
natikgadzhi, line 3: Hm, but perhaps it's worth keeping the `defined?` check in case `sidekiq-cron` decides to change their API? Would be good to fail gracefully instead of erroring out.

st0012: In that case they'll need to rename the constant in a way that doesn't fit the convention anymore, which I think is relatively unlikely. And if that do happen, which will be a breaking change to us, we probably don't want it to always fail silently? 🤔

st0012 (approved):

st0012:
In the case of Ruby 2.6, it'll install an older version of sidekiq-cron (1.9.1), which will not be compatible with our current tests due to the change in `load_from_hash!`. I think we should make `job_spec` compatible with old sidekiq-cron too. Could you also update it for us too? If not, I'll open a separate PR for it later.

natikgadzhi:
Yep, I’m on it.

natikgadzhi:
@st0012 should be good to go now.

sl0thentr0py (approved):

PR #2174, opened 2023-11-22
Release detector: Use full SHA hash in default release name

### LINKED ISSUE #2114: Automatic release detection naming conflict with Github Action
Currently when you are utilizing the Sentry Ruby SDK if you do not specify a release name, it uses [git-based automatic detection](https://github.com/getsentry/sentry-ruby/blob/ace338c6d1e7a75885a8524991c88d491261aad0/sentry-ruby/lib/sentry/release_detector.rb#L31) which uses a `git rev-parse --short HEAD` command. 

This works nicely, however when creating a release using the [Github Action Sentry Release](https://github.com/marketplace/actions/sentry-release) the only option is to use the full SHA as it is all that Github exposes as a variable. This release is preferable to default to as it associates commits with the release and gives you a clearer picture of what is broken

For example you can wind up with two releases within your sentry project: 
- one `f029dd0a822e8e99e19ee50f21a7` with github commits associated and files changed
- another `f029dd0` that has the actual errors being sent to it

Would it be possible to change the Sentry Ruby Gem to use the full SHA with `git rev-parse HEAD` to align with the Github Action?

FWIW this change has already been discussed and changed by the sentry-python SDK a few years back. https://github.com/getsentry/sentry-python/issues/908

natikgadzhi:
## Summary

This pull request makes it so Sentry-Ruby will name releases by their full SHA commit hash instead of the short version. 
- Aligns with how `sentry-cli` does that
- Aligns with Python SDK: https://github.com/getsentry/sentry-python/issues/908
- Closes #2114

/cc @sl0thentr0py and @stephanie-anderson

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Features
 
 - Improve default slug generation for Crons [#2168](https://github.com/getsentry/sentry-ruby/pull/2168)
+- Change release name generator to use full SHA commit hash and align with `sentry-cli` and other Sentry SDKs [#2174](https://github.com/getsentry/sentry-ruby/pull/2174)
 - Automatic Crons support for scheduling gems
   - Add support for [`sidekiq-cron`](https://github.com/sidekiq-cron/sidekiq-cron) [#2170](https://github.com/getsentry/sentry-ruby/pull/2170)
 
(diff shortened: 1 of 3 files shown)


sl0thentr0py (approved):
thx

PR #2175, opened 2023-11-23
Spotlight Sidecar support in Sentry-Ruby

natikgadzhi:
## Summary

This PR adds Spotlight / Sidecar support to Sentry-Ruby!
- PHP example: https://github.com/getsentry/sentry-php/pull/1647 (thank you @cleptric!)
- JS Browser SDK: https://github.com/getsentry/sentry-javascript/pull/9629 by @Lms24 

This is still work in progress ;) /cc @sl0thentr0py, @cleptric, feedback very welcome! 

## Changes
- Added a separate `Spotlight::Configuration` object that has `enabled` and `sidecar_url` fields with defaults. 
- Added a separate `Spotlight::Transport` that is similar to `HTTPTransport`, but YOLO-grade. Doesn't care about rate limits.
- Added a hook into `http_transport.rb` (not proud of that one) that calls `Spotlight::Transport.new().send_data(data)` when it has stuff to send, but before validating rate limits. Aligned with PHP version.

## TODO
- [x] Changelog + Documentation in Sentry-Ruby
- [x] Follow-up with docs on Spotlight repo
- [x] See if we need to disable GZIP for now, or add GZIP support to Sidecar https://github.com/getsentry/spotlight/issues/102

## Reviewing
- I verified that it sends data to local Sidecar proxy
- Proxy says it receives events
- But, my client-side app struggles to connect to proxy — so I don't see the traces in the UI yet. Expect some messiness there. https://github.com/getsentry/spotlight/issues/116


HazAT:
@natikgadzhi Added support to the sidecar to accept gzip/deflate encoding

sl0thentr0py (changes requested):

natikgadzhi:
Okay, update time! 

- I removed the separate `Spotlight::Configuration` in favor of the simple Sentry-config-level setting.
- I verified that the thing actually works.

As long as @sl0thentr0py is happy, we can merge this as is, so folks can start using it. My approach would be to merge and follow-up with next steps, but while I have time, I'll add things in this PR:

- [ ] Improve `Spotlight` transport layer — try to re-use `HTTPTransport` as much as possible, perhaps inherit from it and override the methods that we won't use explicitly. 
- [ ] Add tests for Spotlight itself to verify it picks up config URL correctly.
- [ ] Add back gzip encoding (perhaps easiest to inherit from http_transport).

/cc @HazAT

natikgadzhi:
@HazAT, thank you for the Spotlight docs!

sl0thentr0py:
tested new changes without DSN

image
image

sl0thentr0py (approved):
ok i made some changes and added specs
@natikgadzhi @HazAT if you wanna take a quick look, otherwise i'll merge and release tomorrow

natikgadzhi:
I’ll look later today, but you should merge it, don’t block on me.

Huge thank you for picking this up ❤️‍🩹

sl0thentr0py:
discussed with @HazAT - i will add a max 3 retry on the spotlight transport so we don't spam logs in production tomorrow

natikgadzhi:
I saw that in JS SDK, sounds good to be consistent. 

Want me to add some docs to the readme here, and to sentry-docs?

sl0thentr0py:
i'm not sure if spotlight docs will be in sentry or separate, leaving that to @HazAT

PR #2178, opened 2023-11-23
Don't report network errors to Sentry

### LINKED ISSUE #1787: Net::HTTP errors are sent upstream
### Issue Description

Net::HTTP errors (here `Net::OpenTimeout`) don't derive from `StandardError` so they are not rescued in `send_event` and  sent upstream as exceptions to Sentry.

### Reproduction Steps

The following issue is from `SentryEventJob` with `config.async` but I think the problem holds generally.
https://sentry.io/share/issue/8e2e207b0fec41aaa2615f610a5437d6/

### Expected Behavior

Exceptions in the SDK should be gracefully ignored.

### Actual Behavior

They show up in Sentry.

### Ruby Version

2.6.8p205

### SDK Version

5.2.1

_No response_

### Sentry Config

_No response_

natikgadzhi:
## Summary

This pull request makes it so if Sentry encounters one of the network errors when sending data to Sentry, it won't report that error to Sentry and just re-raise it upstream instead.

Closes #1787.

## Changes

`::Net::HTTP` has quite a few possible errors, but I opted to rescue potential network errors (alongside `SocketError` that was already rescued) instead of rescuing `StandardError`, so that if something exotic comes up, the user does get a report in Sentry still.

Also added a changelog entry.

/cc @sl0thentr0py does that look like an adequate approach to you?


natikgadzhi:
Wow wow wow wait up, that coverage report does not look right.

natikgadzhi:
@st0012 and this one, pending the changelog resolution ;-)

sentry-ruby/lib/sentry/transport/http_transport.rb
```diff
@@ -14,6 +14,13 @@ class HTTPTransport < Transport
     RATE_LIMIT_HEADER = "x-sentry-rate-limits"
     USER_AGENT = "sentry-ruby/#{Sentry::VERSION}"
 
+    NET_HTTP_ERRORS = [
```
sl0thentr0py, line 17: should we maybe use [this list from bundler](https://github.com/ruby/ruby/blob/b0c639f249165d759596f9579fa985cb30533de6/lib/bundler/fetcher.rb#L281-L286)? it has a few more types

sl0thentr0py (changes requested):

sentry-ruby/lib/sentry/transport/http_transport.rb
natikgadzhi: We can! 

Pros:
- Better coverage! (I didn't think about Zlib errors)

Cons:
- Assumes Bundler is there. 


I think it's an okay trade-off to depend on Bundler?

sl0thentr0py: yes just copy over

natikgadzhi:
@sl0thentr0py, done — copied the list from Bundler, with two gotchas: 

- We don't do `::Net::HTTP::Persistent`, so it doesn't make sense to intercept it (I think?) 
- Bundler did not catch `ECONNREFUSED`, so I added it to our list. 

I think in general, having the most common errors is enough. If we miss an error, that's not the end of the world, that's just going to get reported upstream, no big deal. We just need to get 95% right. Right?

sl0thentr0py:
you'll need to move the changelog entry to a new 'Unreleased' section at the top now that i made a release

natikgadzhi:
Done! AFK for 2 hours.

PR #2179, opened 2023-11-24
Allow SENTRY_TRACE and SENTRY_BAGGAGE env vars to continue traces

### LINKED ISSUE #2099: Add a mechanism to continue a trace from environment variables.

natikgadzhi:
## Summary

This pull request is part 3 of https://github.com/getsentry/sentry-ruby/issues/2056. It adds support for continuing traces by passing `SENTRY_TRACE` and `SENTRY_BAGGAGE` environment variables.

Closes #2099. 

## Changes

- I've renamed `Baggage.from_header` to be `Baggage.from_baggage_string` and ditto for traces, just for readability.
- Added ENV vars support and unit tests.
- Added a changelog entry.

## Reviewing

I'm new to distributed tracing, and Sentry implementation in particular, so I might need more handholding with the next few PRs. 

In #2056:

> You may update the traceId and dynamicSamplingContext from the request headers during an incoming request or **if the process was exposed to a SENTRY_TRACE and/or SENTRY_BAGGAGE environment variable if performance is disabled**.

In my current implementation, if the trace string was not passed in headers otherwise, `PropagationContext` will try and read the environment variable. It does NOT check whether Sentry Performance is enabled. Is that correct?

/cc @sl0thentr0py

### DIFF
--- a/sentry-ruby/lib/sentry/baggage.rb
+++ b/sentry-ruby/lib/sentry/baggage.rb
@@ -26,11 +26,11 @@ module Sentry
     #
     # @param header [String] The incoming Baggage header string.
     # @return [Baggage, nil]
-    def self.from_incoming_header(header)
+    def self.from_baggage_string(baggage_string)
       items = {}
       mutable = true
 
-      header.split(',').each do |item|
+      baggage_string.split(',').each do |item|
         item = item.strip
         key, val = item.split('=')
 
--- a/sentry-ruby/lib/sentry/propagation_context.rb
+++ b/sentry-ruby/lib/sentry/propagation_context.rb
@@ -40,31 +40,37 @@ module Sentry
       @baggage = nil
       @incoming_trace = false
 
-      if env
-        sentry_trace_header = env["HTTP_SENTRY_TRACE"] || env[SENTRY_TRACE_HEADER_NAME]
-        baggage_header = env["HTTP_BAGGAGE"] || env[BAGGAGE_HEADER_NAME]
+      # Invoking code could pass a nil env, so let's ||= it so it's easier to work with.
+      env ||= {}
 
-        if sentry_trace_header
-          sentry_trace_data = self.class.extract_sentry_trace(sentry_trace_header)
+      # Trace string could be passed from the invoking code,
+      # or via the environment variable.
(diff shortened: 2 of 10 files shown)


natikgadzhi:
@st0012 if you feel like reviewing another one, this should be a quick and clean one.

sentry-ruby/lib/sentry/propagation_context.rb
```diff
@@ -40,31 +40,37 @@ def initialize(scope, env = nil)
...
-
-            @baggage.freeze!
-            @incoming_trace = true
-          end
+      # Invoking code could pass a nil env, so let's ||= it so it's easier to work with.
+      env ||= {}
+
+      # Trace string could be passed from the invoking code,
+      # or via the environment variable.
+      sentry_trace_string = env["HTTP_SENTRY_TRACE"] || env[SENTRY_TRACE_HEADER_NAME] || ENV["SENTRY_TRACE"]
```
st0012, line 48: I may be wrong, but I don't think the "environment variables" here means the system environment variables? It likely means request envs too?

sentry-ruby/lib/sentry/baggage.rb
```diff
@@ -26,11 +26,11 @@ def initialize(items, mutable: true)
     #
     # @param header [String] The incoming Baggage header string.
     # @return [Baggage, nil]
-    def self.from_incoming_header(header)
+    def self.from_baggage_string(baggage_string)
```
st0012, line 29: I'm not sure if this method is supposed to be public or not. If it is public for users, we can't simply rename it as it'll break users' apps.

st0012 (commented):
I left a few comments but I don't have much context on this part of tracing, so it's safer for @sl0thentr0py to give another review.

natikgadzhi:
Yep, I _might_ be wrong here. Because the request env vars were already respected, and the OS environment vars for this seem a _little bit off_ to me.

sentry-ruby/lib/sentry/baggage.rb
natikgadzhi: Yeah, you're right. Even if it's not public by design, folks might be using it. I could either make a deprecated wrapper with the old name and keep the new name, or revert the rename, either are fine by me.

natikgadzhi (commented):
Welp, I actually think I messed up and by "read trace from environment" we didn't mean OS env. In other SDKs (php, Python, etc), most instrumentations only pass request environment.

PR #2181, opened 2023-11-26
Tracing: don't record HTTP OPTIONS and HEAD transactions

### LINKED ISSUE #1864: Don't trace OPTIONS/HEAD requests by default
We decided that OPTIONS/HEAD requests shouldn't be traced by default since they eat into quota a lot for some users and don't add much value.
 
See https://github.com/getsentry/sentry-javascript/pull/5485

natikgadzhi:
## Summary

This pull request makes it so HTTP OPTIONS and HEAD requests will not be sampled (traced). 

Closes #1864. Based on the approach in https://github.com/getsentry/sentry-javascript/pull/5485

## Changes
- Adds a check for Rack env http method in `capture_exceptions` rack app middleware. It felt like the right spot for this, instead of shoving another check in `Transaction.set_initial_sample_decision`, because it depends directly on an HTTP request, and thus is rather specific to Rack env. 
- Adds a changelog entry
- [ ] This needs unit tests, but `capture_exceptions_spec.rb` is scary. I'll write them up tomorrow.

## Open questions

- @sl0thentr0py, @st0012, is this the right approach and the right place for this check? 
- Another note: this approach ignores the incoming trace. If I'm reading the Javascript implementation correctly, it does the same, since the HTTP method check is the first one in the tracing middleware.


sentry-ruby/lib/sentry/rack/capture_exceptions.rb
```diff
@@ -62,7 +63,17 @@ def capture_exception(exception, env)
...
-        options = { name: scope.transaction_name, source: scope.transaction_source, op: transaction_op }
+        options = {
+          name: scope.transaction_name,
+          source: scope.transaction_source, 
+          op: transaction_op
+        }
+        
+        # Tell Sentry to not sample this transaction if this is an HTTP OPTIONS or HEAD request.
+        if IGNORED_HTTP_METHODS.include?(env["REQUEST_METHOD"])
+          options.merge!(sampled: false)
```
st0012, line 74: I wonder if we should early return here instead so later computations (which are plenty of them) are avoided.

@sl0thentr0py WDYT?

sl0thentr0py: yes I was also about to write the same but got distracted, we should just return early

natikgadzhi: I'm on it!

natikgadzhi: Alright, catching up. 

If I read the code correctly, then if we bail out and return nil early, we won't start a transaction. The only side effect that I see is that we won't set the transaction on that scope, but we'll still track errors as usual, and we'll set the transaction name based on the request path.

I _think_ this is fine. Because if we do go and make the transaction with `@sampled = false`, it won't be much use anyway questionmark.

Pushing up the change now.

sl0thentr0py (changes requested):
you'll need the same logic in 
https://github.com/getsentry/sentry-ruby/blob/9a22f26b34fe77cd755354777d1e62185062bb04/sentry-rails/lib/sentry/rails/capture_exceptions.rb#L34
and 
https://github.com/getsentry/sentry-ruby/blob/9a22f26b34fe77cd755354777d1e62185062bb04/sentry-rails/lib/sentry/rails/action_cable.rb#L35

natikgadzhi:
Just passing by here — I'll clean this up today so we can get it merged! Sorry for the delay.

natikgadzhi:
@sl0thentr0py, thank you for the guidance <3 

I've pushed up the code, with two gotchas where I'd love your input and preference:
1. I've duplicated the `IGNORED_HTTP_METHODS` in the three classes that use it. In theory, we could just put it in `module Sentry`, and use it with `Sentry::IGNORED_HTTP_METHODS`, but I think having it in the same file is a bit more readable. Want me to DRY it out?
2. I have not yet added the tests for the Rails and ActionCable pieces. I can get back to this, or we can ship this as is — up to you!

natikgadzhi:
Hmmmmmmmm 

```
   1) Sentry::Client#event_from_transaction correct dynamic_sampling_context when head SDK
     Failure/Error:
       expect(event.dynamic_sampling_context).to eq({
         "environment" => "development",
         "public_key" => "12345",
         "sample_rate" => "1.0",
         "sampled" => "true",
         "transaction" => "test transaction",
         "trace_id" => transaction.trace_id
       })

       expected: {"environment"=>"development", "public_key"=>"12345", "sample_rate"=>"1.0", "sampled"=>"true", "trace_id"=>"f86126d5bfb14f5d8716c5459fe6495b", "transaction"=>"test transaction"}
            got: {"environment"=>"development", "public_key"=>"12345", "sample_rate"=>"0.5", "sampled"=>"true", "trace_id"=>"f86126d5bfb14f5d8716c5459fe6495b", "transaction"=>"test transaction"}

       (compared using ==)

       Diff:
       @@ -1,6 +1,6 @@
        "environment" => "development",
        "public_key" => "12345",
       -"sample_rate" => "1.0",
       +"sample_rate" => "0.5",
        "sampled" => "true",
        "trace_id" => "f86126d5bfb14f5d8716c5459fe6495b",
        "transaction" => "test transaction",
     # ./spec/sentry/client_spec.rb:183:in `block (3 levels) in <top (required)>'
     
 ```

sl0thentr0py:
will take this up separately

Issue #2138, opened 2023-10-13
Improve slug generation for crons


cleptric:
I have a question regarding the Ruby SDK. In the [documentation](https://docs.sentry.io/platforms/ruby/crons/#job-monitoring), it states

```ruby
# slug defaults to the job class name
sentry_monitor_check_ins slug: 'custom_slug'
```
Currently it appears that the `slug` parameter is always required, since Ruby class names can contain punctuation (`::`) and will have capital letters.

###### e.g.
```ruby
SuperFancy::PrivateGroups::SpecialOrganizations::ScheduleOutreachWorker
```
[In the code](https://github.com/getsentry/sentry-ruby/blob/master/sentry-ruby/lib/sentry/cron/monitor_check_ins.rb#L6), I would suggest that they use `self.class.name.parameterize` as the default, which will downcase the class name and replace the double colons with a dash.

```ruby
[1] pry(main)> e=SuperFancy::PrivateGroups::SpecialOrganizations::ScheduleOutreachWorker.new
=> #<SuperFancy::PrivateGroups::SpecialOrganizations::ScheduleOutreachWorker:0x00007f13c7a366f0>
[2] pry(main)> e.class.name.parameterize
=> "superfancy-specialgroups-specialorganizations-scheduleoutreachworker"
[3] pry(main)> e.class.name.parameterize.size
=> 68
```

This example demonstrates the other issue that we're running into, which is one of length. Since the Dashboard UI constrains the value to 50 characters (this not mentioned in the documentation), it appears the `slug` parameter will still have to be used.

Are these assumptions correct?

_Originally posted by @nimylian in https://github.com/getsentry/sentry/discussions/42283#discussioncomment-7274839_

swanson:
One concern I have is that it appears slugs can be duplicated across environments -- could/should the gem prefix the Rails environment to the slug?

natikgadzhi:
> This example demonstrates the other issue that we're running into, which is one of length. Since the Dashboard UI constrains the value to 50 characters (this not mentioned in the documentation), it appears the slug parameter will still have to be used.

Would it be better to try and omit pieces of the default slug if it runs out of 50 characters, or would it be more beneficial to omit the middle part of the longer slugs in the Dashboard UI? I would assume Dashboard will have to deal with many things that are too long to display, eventually?

sl0thentr0py:
thanks for the feedback everyone, I'll think about a good way to do this when I have some time to fix this. 
[...]

PR #2184, opened 2023-11-29
Improve default slug generation on `sidekiq-scheduler`

y-yagi:

**Other Notes**

## Description

`sidekiq-scheduler` allows to use a class name as a schedule name directly as follows.

```
Namspeced::CancelAbandonedOrders:
  cron: '0 */5 * * * *'
```

In this case, a slug is shown as `namspecedcancelabandonedorders` on Sentry now. I think this isn't good for readability.
This PR converts `::` to `-` as well as the cron default behavior of Crons mixin.

Related with: #2138.


sentry-ruby/lib/sentry/cron/monitor_check_ins.rb
```diff
@@ -44,9 +44,9 @@ def sentry_monitor_check_ins(slug: nil, monitor_config: nil)
           prepend Patch
         end
 
-        def sentry_monitor_slug
+        def sentry_monitor_slug(base_name = name)
```
st0012, line 47: Nit: let's keep calling it `name` as `base_name` could easily be associated `basename` methods in classes like `Pathname` and cause confusion. And IMO using karg will also make the API easier to use & maintain:

```rb
def sentry_monitor_slug(name: self.name)
end
```

sl0thentr0py (commented):
thx @y-yagi please make the change that @st0012 requested, otherwise lgtm

y-yagi:
@st0012 
Thanks for your review! I fixed your point!

@sl0thentr0py 
Done!

st0012 (approved):
Thank you!

PR #2204, opened 2023-12-19
Apply bundler-cache to CI's Ruby setup

st0012:
`bundler-cache` should automatically [pick the right bundler version](https://github.com/ruby/setup-ruby/blob/af848b40be8bb463a751551a1180d74782ba8a72/bundler.js#L126-L137) for us.
I also applied prettier formatting to the files as I do think it improves readability.


st0012:
I think danger is failing for an unrelated reason:

```
Failing the build, there are 7 fails.
Request failed [403]: https://api.github.com/user
Response: {
  "message": "Resource not accessible by integration",
  "documentation_url": "https://docs.github.com/rest/users/users#get-the-authenticated-user"
}
Request failed [403]: https://api.github.com/user
Response: {
  "message": "Resource not accessible by integration",
  "documentation_url": "https://docs.github.com/rest/users/users#get-the-authenticated-user"
}
```

sl0thentr0py (approved):
oh nice

sl0thentr0py:
danger is failing because it wants you to pin the SHA for github actions, we can ignore that

PR #2207, opened 2023-12-22
Refactor Sentry.close

st0012:
When reviewing #2206, I spotted 2 things that could be improved:

- When closing the SDK, we shouldn't need to enqueue the last client report sending to background worker again. This makes the behaviour of `close` more difficult to predict.
- Since the session flusher can still enqueue a job to background worker when flushing, we should postpone background worker's shutdown instead.

#skip-changelog

### DIFF
--- a/sentry-ruby/lib/sentry-ruby.rb
+++ b/sentry-ruby/lib/sentry-ruby.rb
@@ -233,13 +233,8 @@ module Sentry
     #
     # @return [void]
     def close
-      @background_worker.perform { get_current_client&.transport&.flush }
-      # session_flusher internally queues to the background worker too
-      @session_flusher&.flush
-
-      @background_worker.shutdown
-
       if @session_flusher
+        @session_flusher.flush
         @session_flusher.kill
         @session_flusher = nil
       end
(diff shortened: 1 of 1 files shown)


sl0thentr0py (commented):
the client reports are sent only once every interval, so we need to flush the pending stats on close. This was the main point of my change.

edit: ah you mean send them in the main thread, hmm

sl0thentr0py (approved):

st0012:
> ah you mean send them in the main thread

Yeah. I don't see a reason to send it async at that stage while waiting to shutdown the worker right away.

sentry-ruby/lib/sentry-ruby.rb
```diff
@@ -249,10 +244,16 @@ def close
         @backpressure_monitor = nil
       end
 
-      if configuration&.include_local_variables
-        exception_locals_tp.disable
+      if client = get_current_client
```
st0012, line 247: @sl0thentr0py Another "concern" I have is that we currently only focus on the main hub's top client, but not other clients. Ideally we should do that for all clients of the main hub, as well as other hubs' (which we don't track AFAIK) clients?

I think the former is probably fine as a hub is unlikely to have more than 1 client. But I do recall some customers tried doing multi-hub setup in the past (also rare tho).

I know both are fairly rare. I just want to raise it so we keep that in mind.

sl0thentr0py: yea both valid concerns but as a heads up, sometime next year we'll move to removing the Hub and only having a Scope (which will be closer to opentelemetry's context). 
rfc here - https://github.com/getsentry/rfcs/pull/122

st0012: Thank you for the context! I just gave it a first pass and got confused around different types of scopes too, as raised by you and many other comments.

But yeah if I understand it correctly, there will only be 1 client in the future and no hubs, which means this won't be a concern anymore.

PR #2208, opened 2023-12-27
Add Ruby 3.3 to CI matrix

st0012:
1. The `ruby_versions` workflow will save us from manually updating Ruby matrix in the future, a minor improvement.
    - It already adds Ruby 3.3 to the matrix to those adopted it.
1. We can't apply `ruby_versions` easily to sentry-rails & sentry-sidekiq as they have rather complicated Ruby vs lib version mapping.
    - So I manually added 3.3 to the Ruby matrix.

#skip-changelog


sl0thentr0py (approved):

sl0thentr0py:
you can ignore danger about those pins

PR #2211, opened 2023-12-28
Allow setting defaults for MonitorConfig objects with new Cron::Confi…

### LINKED ISSUE #2205: Ability to change the default grace period and timeout for all Cron Monitors when using the sidekiq_scheduler patches
**Describe the idea**

We find the sidekiq_scheduler patch extremely useful, especially as we have 100s of cron jobs defined there.

However, the default grace period of 1 minute does not work that well for our queue configuration: the overwhelming majority of our cron jobs will have latency of 1-3 minutes, and therefore Sentry will create cron issues every single time the checkin is delayed by a couple of minutes, resulting in a abysmal signal-to-noise ratio.

image

We would like the ability to change the checkin margins for all monitors created by the patch. An alternative could be an ability to configure the default values per Sentry project, but I'm not sure where to submit that, and what approach you like best.

I'm not sure about most users, but at this scale (100s of cron jobs) it is very likely that their latency will exceed the default 1 minute.

**Possible implementation**

The sidekiq_scheduler patch is currently using only the `schedule` positional argument to instantiate Monitors, but they support the necessary keyword arguments (`checkin_margin` and `max_runtime`) as well.

https://github.com/getsentry/sentry-ruby/blob/master/sentry-sidekiq/lib/sentry/sidekiq-scheduler/scheduler.rb#L38-L43

https://github.com/getsentry/sentry-ruby/blob/master/sentry-ruby/lib/sentry/cron/monitor_config.rb#L26-L31

Perhaps [Configuration](https://github.com/getsentry/sentry-ruby/blob/master/sentry-ruby/lib/sentry/configuration.rb) could be extended to allow setting those two values, and the default value for those keyword arguments could be taken from configuration instead of being nils?

If that sounds good, I'm happy to take a stab at contributing this.

sl0thentr0py:
closes #2205

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,17 @@
 - Implement proper flushing logic on ``close`` for Client Reports and Sessions [#2206](https://github.com/getsentry/sentry-ruby/pull/2206)
 - Support cron with timezone for `sidekiq-scheduler` patch [#2209](https://github.com/getsentry/sentry-ruby/pull/2209)
   - Fixes [#2187](https://github.com/getsentry/sentry-ruby/issues/2187)
+- Add `Cron::Configuration` object that holds defaults for all ``MonitorConfig`` objects [#2210](https://github.com/getsentry/sentry-ruby/pull/2210)
+
+    ```ruby
+    Sentry.init do |config|
+      # ...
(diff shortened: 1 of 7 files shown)


adinauer (approved):
LGTM

PR #2212, opened 2023-12-29
Improvements around the cron monitoring feature

st0012:
After reviewing the recent cron monitoring PRs, I want to make a few small improvements:

1. Testing the `MonitorCheckIns` module end-to-end: Since the module captures the core logic of the cron monitoring feature, I think making sure it works e2e in tests is very beneficial, especially when the test complexity doesn't really increase much (see the changes).
2. Applying the global cron configs in the `MonitorCheckIns`: this will keep the `MonitorConfig` simple and isolated.
3. Fixing `MonitorCheckIns#perform`'s rescue scope: it should only be applied AFTER the `start` local is defined. Otherwise, `Sentry.utc_now.to_i - start` would cause a new `TypeError` and overshadow the original exception.
4. There's an unnecessary extend

#skip-changelog

### DIFF
--- a/sentry-ruby/lib/sentry/cron/monitor_check_ins.rb
+++ b/sentry-ruby/lib/sentry/cron/monitor_check_ins.rb
@@ -39,6 +39,13 @@ module Sentry
 
       module ClassMethods
         def sentry_monitor_check_ins(slug: nil, monitor_config: nil)
+          if monitor_config && Sentry.configuration
+            cron_config = Sentry.configuration.cron
+            monitor_config.checkin_margin ||= cron_config.default_checkin_margin
+            monitor_config.max_runtime ||= cron_config.default_max_runtime
+            monitor_config.timezone ||= cron_config.default_timezone
+          end
+
           @sentry_monitor_slug = slug
           @sentry_monitor_config = monitor_config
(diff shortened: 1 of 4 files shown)


st0012:
@sl0thentr0py This also fixes an order-dependent CI failure in master, which can be reproduced with:

```
rspec './spec/sentry/client_spec.rb[1:7:2,1:7:3]' './spec/sentry/cron/monitor_config_spec.rb[1:1:1]' --seed 60911
```

It happens because the current `monitor_config_spec` sets `Sentry.configuration.cron`, but the value would be picked up by client specs as they don't initialise the SDK themselves.

sentry-ruby/lib/sentry/cron/monitor_check_ins.rb
```diff
@@ -57,8 +67,6 @@ def sentry_monitor_config
         end
       end
 
-      extend ClassMethods
```
sl0thentr0py, line 60: this was because I was earlier trying to make something that could be injected to both classes and modules, but now it's class only so it's not longer needed.

sl0thentr0py (approved):
thx! all make sense to me

PR #2216, opened 2024-01-02
Clean up logging

### LINKED ISSUE #2031: Reduce the logging
### Issue Description

Having this in my config isn't sufficient
```
config.logger.level = Logger::WARN
```

We are still getting a lot of spam messages about rate limits, transport, envelopes, traces (which we don't even have in our subscription plan), and a lot of other stuff that makes no sense to us and there is nothing we can do about it. So we basically don't need all those messages that we can't resolve.

### Reproduction Steps

Just enable Sentry in the app and observe there's a lot of stuff in the log.

### Expected Behavior

Ability to set the log level to something that'd only show us real warnings and problems we can resolve (like without upgrading our plan).

### Actual Behavior

Changing the log level doesn't reduce the logging

### Ruby Version

2.7.6

### SDK Version

5.7.0 & 5.8.0

_No response_

### Sentry Config

_No response_

sl0thentr0py:
closes #2031


sentry-ruby/lib/sentry/client.rb
```diff
@@ -177,11 +177,7 @@ def send_event(event, hint = nil)
 
       event
     rescue => e
-      loggable_event_type = event_type.capitalize
-      log_error("#{loggable_event_type} sending failed", e, debug: configuration.debug)
-
-      event_info = Event.get_log_message(event.to_hash)
```
st0012, line 183: In the past we relied on this message to look into the potentially problematic event payload. Do we have an alternative for this now?

sl0thentr0py: i want to move away from logging whole payloads for the end user, one can always log it in `before_send` if one really wants. Even on our side, we spent a lot of logging resources in python by logging huge payloads once.

st0012: The logging here should be triggered very rarely. Most expected errors should've been captured at `HTTPTransport#send_data` (network errors...etc.) so they don't even reach here.
What's most likely captured here are errors from the `before_send*` callbacks and event/envelope serialisation errors, which with the event data will make debugging much easier for users and us.

sl0thentr0py: `send_data` re-raises here
https://github.com/getsentry/sentry-ruby/blob/c5889798c9b35c0a10f03ac071c6e289da151592/sentry-ruby/lib/sentry/transport/http_transport.rb#L65-L69

so these will end up being logged for network errors too.

st0012: Sorry that I missed it 🤦‍♂️ 
Ok let's remove it first and find a better place to log them if we later find it necessary.

st0012 (approved):

sentry-ruby/lib/sentry/client.rb
sl0thentr0py: Yes but I agree that debugging `before_send` errors is not ideal. We've also discussed this internally.

sentry-ruby/lib/sentry/transport/http_transport.rb
```diff
@@ -53,18 +53,14 @@ def send_data(data)
       end
 
       if response.code.match?(/\A2\d{2}/)
-        if has_rate_limited_header?(response)
-          handle_rate_limited_response(response)
-        end
+        handle_rate_limited_response(response) if has_rate_limited_header?(response)
+      elsif response.code == "429"
```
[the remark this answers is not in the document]
sl0thentr0py: @charligoulet yes, this was deliberately done. In general, we shouldn't raise exceptions in userland for SDK internal things and if you were using this for seeing if you are limited on our SaaS offering, there is visibility in the product (Stats section).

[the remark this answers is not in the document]
sl0thentr0py: I see, thanks for the explanation of your use case. 
So a few points

* we never do a retry ever on our side, a dropped event is a dropped event
* the server's rate limiting is merely interpreted as 'stop doing stuff for a while'
* technically it is breaking yes, but this is considered internal to the http transport implementation and not really part of the API contract

That said, there are a couple of options for you now
* make your own transport deriving from `Sentry::HTTPTransport` and override `send_data` there and do what you wish there
* I can expose a `on_rate_limit` hook in the transport class similar to `on_error` to make that easier so you just need to define that one method to retry

does that work?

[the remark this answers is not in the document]
sl0thentr0py: cool I'll make a new issue to track that

PR #2217, opened 2024-01-04
Don't add most scope data to CheckInEvent

### LINKED ISSUE #2191: Don't apply all scope data to check-ins
avoid size limits

sl0thentr0py:
I also removed the option to pass `&block` to the top level `Sentry.capture_check_in` since that would be useless now (pretty sure no one was using that anyway).

closes #2191

### DIFF
--- a/sentry-ruby/spec/sentry/hub_spec.rb
+++ b/sentry-ruby/spec/sentry/hub_spec.rb
@@ -222,9 +222,16 @@ RSpec.describe Sentry::Hub do
       expect(event[:monitor_config]).to include({ schedule: { type: :crontab, value: "* * * * *" } })
     end
 
-    it_behaves_like "capture_helper" do
-      let(:capture_helper) { :capture_check_in }
-      let(:capture_subject) { [slug, :ok] }
+    context "with sending not allowed" do
+      before do
+        expect(configuration).to receive(:sending_allowed?).and_return(false)
+      end
+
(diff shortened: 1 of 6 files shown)


sentry-ruby/spec/sentry/hub_spec.rb
```diff
@@ -222,9 +222,12 @@
       expect(event[:monitor_config]).to include({ schedule: { type: :crontab, value: "* * * * *" } })
     end
 
-    it_behaves_like "capture_helper" do
-      let(:capture_helper) { :capture_check_in }
-      let(:capture_subject) { [slug, :ok] }
+    it "doesn't send the event nor assign last_event_id" do
```
st0012, line 225: Let's wrap this inside a context like `"with sending not allowed"` so it doesn't look like `capture_check_in` should not be sent.

st0012 (commented):
It looks like `CheckInEvent` shouldn't even inherit `Event` as we don't want any of the attributes from it anyway?

sl0thentr0py:
@st0012 

> It looks like CheckInEvent shouldn't even inherit Event anymore as we don't want any of the attributes from it anyway?

we might introduce a couple back (like tags/extras), we're still iterating on the design, so I don't want to refactor too much. This was mainly to remove breadcrumbs because they caused payload bloat and caused missed check-ins. I removed the other stuff for now and we might back add some later.

st0012 (approved):

Issue #2221, opened 2024-01-09
Missing release on rubygems.org?


gael-ian:
Hello,

All sentry gems for Ruby seem to have been updated to 5.16.0 but `sentry-rails` and `sentry-delayed_job` are still only available as 5.15.2 on rubygems.org. Maybe there was a bug with the release script? Can you push them manually?

Thanks.

sl0thentr0py:
oh damn, on it

sl0thentr0py:
sigh, github is also having an incident right now, I'll try to get it out as soon as they're back up.

sl0thentr0py:
cause was failure to install latest `sqlite3` release 
```
sqlite3-1.7.0-x86_64-darwin requires ruby version < 3.4.dev, >= 3.0, which is incompatible with the current version, ruby 2.7.8p225
```

sl0thentr0py:
@gael-ian 5.16.1 was released correctly for all gems, sorry for the inconvenience!

PR #2224, opened 2024-01-09
Make sure batch_build exit code is correct and require artifacts in craft

sl0thentr0py:
make sure stuff like #2221 doesn't happen again

#skip-changelog


Lms24 (approved):
Craft config change LGTM

PR #2229, opened 2024-01-13
Don't lock Rails under 7.1

st0012:
It's been over half of a year since Rails 7.1's release. I think all our test suites should include it in the test targets.

#skip-changelog

### DIFF
--- a/.github/workflows/sentry_delayed_job_test.yml
+++ b/.github/workflows/sentry_delayed_job_test.yml
@@ -56,6 +56,7 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby_version }}
           bundler-cache: true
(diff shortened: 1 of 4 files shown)


.github/workflows/sentry_delayed_job_test.yml
```diff
@@ -55,7 +55,6 @@ jobs:
         uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby_version }}
-          bundler-cache: true
```
st0012, line 58: For some reason, the next step always require `bundle install` for resolving git-based gems. If we were to keep it, then this is not necessary anymore.

sl0thentr0py (approved):

PR #2231, opened 2024-01-15
Make sure sending_allowed? is respected irrespective of spotlight configuration

### LINKED ISSUE #2226: Sending to sentry is always enabled when using spotlight
### Issue Description

When enabling spotlight, the sentry configuration permits sending events to sentry through `sending_allowed?` regardless of `enabled_environments`.
https://github.com/getsentry/sentry-ruby/blob/5.15.0/sentry-ruby/lib/sentry/configuration.rb#L463

Not setting the `dsn` did not work as a workaround, since `sending_allowed?` also requires a valid DSN before `SpotlightTransport.send_data` is invoked.

### Reproduction Steps

Set `config.spotlight` to `true`, `config.dsn` to a valid DSN, and do not include `config.environment` in `config.enabled_environments`. Trigger an event to be sent threough for example `raise 'test'`.

### Expected Behavior

Events are sent to the spotlight sidecar only, not sentry

### Actual Behavior

Events are sent to both of them

### Ruby Version

3.2.2

### SDK Version

5.15

Rails 7.0.5.1, sidekiq 6.1

### Sentry Config

Sentry.init do |config|
  # Enable Sentry in production and conditionally in development
  config.environment = 'development'
  config.enabled_environments = ['production']

  config.dsn = DSN

  config.spotlight = true
end

sl0thentr0py:
we need to make a difference between the 'event building' and the 'sending' so I made this more explicit. 

fixes #2226

### DIFF
--- a/sentry-ruby/lib/sentry/configuration.rb
+++ b/sentry-ruby/lib/sentry/configuration.rb
@@ -485,9 +485,13 @@ module Sentry
     end
 
     def sending_allowed?
+      spotlight || sending_to_dsn_allowed?
+    end
+
+    def sending_to_dsn_allowed?
       @errors = []
 
-      spotlight || (valid? && capture_in_environment?)
+      valid? && capture_in_environment?
     end
 
     def sample_allowed?
--- a/sentry-ruby/spec/sentry/configuration_spec.rb
+++ b/sentry-ruby/spec/sentry/configuration_spec.rb
@@ -276,6 +276,17 @@ RSpec.describe Sentry::Configuration do
       subject.spotlight = true
       expect(subject.sending_allowed?).to eq(true)
     end
+
+    it "true when sending to dsn allowed" do
+      allow(subject).to receive(:sending_to_dsn_allowed?).and_return(true)
+      expect(subject.sending_allowed?).to eq(true)
+    end
+
+    it "false when no spotlight and sending to dsn not allowed" do
+      allow(subject).to receive(:sending_to_dsn_allowed?).and_return(false)
+      subject.spotlight = false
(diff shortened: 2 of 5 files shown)


sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -478,10 +478,14 @@ def profiles_sample_rate=(profiles_sample_rate)
       @profiles_sample_rate = profiles_sample_rate
     end
 
+    def event_building_allowed?
```
sl0thentr0py, line 481: @st0012 lmk if you have a better name for this :)

sentry-ruby/spec/sentry/configuration_spec.rb
```diff
@@ -271,10 +271,21 @@
     end
   end
 
-  describe "#sending_allowed?" do
```
st0012, line 274: Shouldn't we still have tests for this method?

sentry-ruby/lib/sentry/configuration.rb
st0012: I feel this should be called `sending_allowed?` and the original `sending_allowed?` should be something like `sending_to_sentry?`

sandstrom:
Just curious, is this something that will be merged soon?

We've tried out Spotlight, but with this bug it's also caused all dev events to flow from developer machines to Sentry, with no easy way to turn it off.

Would be great for us if this could be merged!

cc @sl0thentr0py @st0012

sl0thentr0py:
sorry slipped through the cracks

szokeasaurusrex (approved):
stamp

Issue #1904, opened 2022-09-28
Add distributed tracing and dynamic sampling support for all background queues


sl0thentr0py:
use `continue_trace` before `start_transaction` in

- [ ] resque (native, non-ActiveJob jobs only)
- [X] delayed_job
- [X] active_job
- [X] sidekiq (native, non-ActiveJob jobs)

## Status (checked against sentry-ruby source, 2026-09-08)

* `active_job` — done. `Sentry::Rails::ActiveJobExtensions::SentryReporter.record` (`sentry-rails/lib/sentry/rails/active_job.rb`) calls `Sentry.continue_trace(trace_headers, **transaction_options)` before `Sentry.start_transaction`, shipped in 6.7.0 via [#2947](<https://github.com/getsentry/sentry-ruby/issues/2947>). This covers Sidekiq+AJ, Resque+AJ, DelayedJob+AJ, and SolidQueue, since all AJ-backed adapters route perform through this shared reporter.
* `delayed_job` — done. `Sentry::DelayedJob::Plugin.start_transaction` (`sentry-delayed_job/lib/sentry/delayed_job/plugin.rb`) calls `Sentry.continue_trace(env, **options)` before `Sentry.start_transaction`.
* `resque` — still open, but only for native (non-ActiveJob) Resque workers. `Sentry::Resque::SentryReporter.record` (`sentry-resque/lib/sentry/resque.rb`) calls `Sentry.start_transaction` directly with no `continue_trace` step, so trace propagation is missing for jobs that don't go through the `ResqueAdapter`/ActiveJob wrapper. Jobs enqueued via ActiveJob on the Resque adapter already get continuation through the `active_job` path above.

* `sidekiq` — done, and predates 6.7.0. `Sentry::Sidekiq::SentryContextServerMiddleware#start_transaction` (`sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb`) already calls `Sentry.continue_trace(env, **options)` before `Sentry.start_transaction`. Not in the original checklist but confirmed complete for native Sidekiq jobs; AJ-wrapped Sidekiq jobs also get continuation via the `active_job` path.

Remaining work: add `continue_trace` support to the native `sentry-resque` reporter (or confirm out of scope now that `sentry-resque` is maintenance-only per the Queue Integrations Overhaul project).

PR #2233, opened 2024-01-16
Add support for distributed tracing for DelayedJob

sl0thentr0py:
* works for `PerformableMethod`
* activejob will be done more generally

part of #1904

### DIFF
--- a/sentry-delayed_job/lib/sentry/delayed_job/plugin.rb
+++ b/sentry-delayed_job/lib/sentry/delayed_job/plugin.rb
@@ -10,7 +10,12 @@ module Sentry
       OP_NAME = "queue.delayed_job".freeze
 
       callbacks do |lifecycle|
+        lifecycle.before(:enqueue) do |job, *args, &block|
+          inject_trace_data(job) if Sentry.initialized?
+        end
+
         lifecycle.around(:invoke_job) do |job, *args, &block|
+          env = extract_trace_data(job)
           next block.call(job, *args) unless Sentry.initialized?
 
           Sentry.with_scope do |scope|
@@ -20,12 +25,7 @@ module Sentry
             scope.set_contexts(**contexts)
             scope.set_tags("delayed_job.queue" => job.queue, "delayed_job.id" => job.id.to_s)
 
-            transaction = Sentry.start_transaction(
-              name: scope.transaction_name,
-              source: scope.transaction_source,
-              op: OP_NAME,
-              custom_sampling_context: contexts
-            )
+            transaction = start_transaction(scope, env, contexts)
             scope.set_span(transaction) if transaction
 
             begin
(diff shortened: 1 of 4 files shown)


sentry-delayed_job/lib/sentry/delayed_job/plugin.rb
```diff
@@ -91,12 +91,42 @@ def self.report?(job)
...
+      def self.extract_trace_data(job)
+        payload_object = job.payload_object
+        return nil unless payload_object.is_a?(Delayed::PerformableMethod)
+
+        ind = payload_object.args.index { |a| a.is_a?(Hash) && a.key?(:sentry) }
+        return nil unless ind
+
+        env = payload_object.args[ind][:sentry]
+        payload_object.args.delete_at(ind)
+        env
```
st0012, line 128: Can this work?

```suggestion
        target_payload = payload_object.args.find { |a| a.is_a?(Hash) && a.key?(:sentry) }
        return nil unless target_payload
        payload_object.args.delete(target_payload)
        target_payload[:sentry]
```

st0012 (commented):
Instead of injecting trace data to args, can we maybe use a custom ivar? Like `@_sentry_trace_data`. That way the extraction will be easier and we don't need to risk breaking the call completely if the args deletion somehow goes wrong.

sl0thentr0py:
> Instead of injecting trace data to args, can we maybe use a custom ivar? 

we'll need to patch delayed_job's YAML stuff

sentry-delayed_job/lib/sentry/delayed_job/plugin.rb
sl0thentr0py: that delete will do a find again under the hood and I'm not sure it works with an object like that.

st0012:
> we'll need to patch delayed_job's YAML stuff here then

Ah thanks for this context. Let's first go with the current implementation then.

sentry-delayed_job/lib/sentry/delayed_job/plugin.rb
st0012: It works because `delete` uses `==` under the hood. And since the value here is a hash with string values, there should be no issue. I also have tested it with the test suites.

st0012 (approved):

sentry-delayed_job/spec/spec_helper.rb
```diff
@@ -1,4 +1,5 @@
 require "bundler/setup"
+require "debug" if RUBY_VERSION.to_f >= 2.6 && RUBY_ENGINE == "ruby"
 require "pry"
```
st0012, line 3: Not related: do you still use pry?

sl0thentr0py: nope! I've moved to `debugger` if available!

PR #2235, opened 2024-01-16
Add hint: support to Sentry::Rails::ErrorSubscriber

### LINKED ISSUE #2228: add hint support to Sentry::Rails::ErrorSubscriber
if i want to explicitly send an exception to `sentry` that would otherwise be ignored because of `Sentry.configuration.excluded_exceptions` i can do this:

```ruby
Sentry.capture_exception(ignored_exception, hint: { ignore_exclusions: true }) 
```
however, as far as i can tell this isn't possible if i'm using the rails error reporting interface. 

it would be useful to expose this there.

i think it could be achieved the same way `tags` are handled here:

https://github.com/getsentry/sentry-ruby/blob/master/sentry-rails/lib/sentry/rails/error_subscriber.rb#L17C38-L20

so you could just provide a `hint` key in `context:`:
```ruby
Rails.error.report exception, e, handled: true, severity: :error, context: { hint: { ignore_exclusions: true } }
```

and you could then pass it in via this (untested) implementation:
```ruby
      def report(error, handled:, severity:, context:, source: nil)
        tags = { handled: handled }

        if source
          return if SKIP_SOURCES.match?(source)
          tags[:source] = source
        end

        if context[:tags].is_a?(Hash)
          context = context.dup
          tags.merge!(context.delete(:tags))
        end

        hint = {}
        if context[:hint].is_a?(Hash)
          context = context.dup
          hint.merge!(context.delete(:hint)
       end

        Sentry::Rails.capture_exception(error, level: severity, contexts: { "rails.error" => context }, tags: tags, hint: hint)
      end
```

modosc:
closes #2228

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 ## Unreleased
 
+### Features
+
 - Fix warning about default gems on Ruby 3.3.0 ([#2225](https://github.com/getsentry/sentry-ruby/pull/2225))
+- Add `hint:` support to `Sentry::Rails::ErrorSubscriber` [#2235](https://github.com/getsentry/sentry-ruby/pull/2235)
 
 ## 5.16.1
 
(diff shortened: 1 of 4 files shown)


sl0thentr0py (approved):
thx @modosc

PR #2241, opened 2024-02-04
Adopt rubocop and format the project

st0012:
In addition to having a consistent coding style through CI checks, anyone uses a LSP server (whether `Ruby LSP` or `Solargraph`) will have auto-formatting in their editor as well.

1. After checking different rubocop rules, I think [rubocop-rails-omakase](https://github.com/rails/rubocop-rails-omakase)'s rules match my taste the most but also doesn't require much changes in our codebase.
    - I didn't use `standard` as it's not supported by `Ruby LSP` atm
1. I ran `rubocop -a` on the entire project to apply the styles, with a couple of manual adjustments.
1. I also added a new CI job for linting.

#skip-changelog

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,9 @@
+inherit_gem:
+  rubocop-rails-omakase: rubocop.yml
+
+Layout/SpaceInsideArrayLiteralBrackets:
+  Enabled: false
+
+AllCops:
+  Exclude:
+    - 'sentry-raven/**/*'
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,16 @@
+name: Lint
+
+on:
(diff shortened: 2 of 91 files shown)


.rubocop.yml
```diff
@@ -0,0 +1,9 @@
+inherit_gem:
+  rubocop-rails-omakase: rubocop.yml
+
+Layout/SpaceInsideArrayLiteralBrackets:
+  Enabled: false
```
st0012, line 5: In addition to not liking this style myself, having it enabled requires changing almost all array literals in the codebase (~500 more offences). So I decided to not take this one in.

st0012:
@sl0thentr0py 👋 let me know if you have concern about this?

sl0thentr0py (approved):
nice, thx!

Issue #2246, opened 2024-02-12
[ruby] Add metrics


sl0thentr0py:
https://github.com/getsentry/team-sdks/issues/36
```[tasklist]
### Tasks
- [ ] https://github.com/getsentry/sentry-ruby/pull/2247
- [ ] https://github.com/getsentry/sentry-ruby/pull/2254
- [ ] https://github.com/getsentry/sentry-ruby/pull/2255
- [ ] https://github.com/getsentry/sentry-ruby/pull/2258
- [ ] https://github.com/getsentry/sentry-ruby/pull/2263
```

PR #2247, opened 2024-02-16
Add Metrics API and aggregator

sl0thentr0py:
* new `Sentry::Metrics` module with 4 apis that map to the new 4 `Sentry::Metrics::Metric` classes
  * `increment` - simple counter
  * `distribution` -  array of observations
  * `gauge` - statistics (last/min/max/sum/count)
  * `set` - unique values
* new `Sentry::Metrics::Aggregator` that starts a thread that flushes pending metric buckets in 5 second intervals 
  * buckets are a nested hash of timestamp (rolled to 10 second intervals) -> metric keys -> actual metric instance
  * there is a random `flush_shift` once per startup to create jittering
  * flushable buckets are sent in a new `statsd` type envelope that is not json so made a small change to the `Envelope::Item`
  * tag key/values are sanitized for unicode/special characters according to the two regexes

Reference spec - https://develop.sentry.dev/sdk/metrics/

part of #2246

### DIFF
--- a/sentry-ruby/lib/sentry/envelope.rb
+++ b/sentry-ruby/lib/sentry/envelope.rb
@@ -7,11 +7,12 @@ module Sentry
       STACKTRACE_FRAME_LIMIT_ON_OVERSIZED_PAYLOAD = 500
       MAX_SERIALIZED_PAYLOAD_SIZE = 1024 * 1000
 
-      attr_accessor :headers, :payload
+      attr_accessor :headers, :payload, :is_json
 
-      def initialize(headers, payload)
+      def initialize(headers, payload, is_json: true)
         @headers = headers
         @payload = payload
+        @is_json = is_json
       end
 
       def type
@@ -19,7 +20,7 @@ module Sentry
       end
 
       def to_s
-        [JSON.generate(@headers), JSON.generate(@payload)].join("\n")
+        [JSON.generate(@headers), @is_json ? JSON.generate(@payload) : @payload].join("\n")
       end
 
       def serialize
@@ -78,8 +79,8 @@ module Sentry
       @items = []
     end
 
-    def add_item(headers, payload)
-      @items << Item.new(headers, payload)
+    def add_item(headers, payload, is_json: true)
+      @items << Item.new(headers, payload, is_json: is_json)
     end
 
     def item_types
--- a/sentry-ruby/lib/sentry/envelope.rb
+++ b/sentry-ruby/lib/sentry/envelope.rb
@@ -7,12 +7,11 @@ module Sentry
       STACKTRACE_FRAME_LIMIT_ON_OVERSIZED_PAYLOAD = 500
       MAX_SERIALIZED_PAYLOAD_SIZE = 1024 * 1000
 
-      attr_accessor :headers, :payload, :is_json
+      attr_accessor :headers, :payload
 
-      def initialize(headers, payload, is_json: true)
+      def initialize(headers, payload)
         @headers = headers
         @payload = payload
-        @is_json = is_json
       end
 
       def type
@@ -20,7 +19,7 @@ module Sentry
       end
 
       def to_s
-        [JSON.generate(@headers), @is_json ? JSON.generate(@payload) : @payload].join("\n")
+        [JSON.generate(@headers), @payload.is_a?(String) ? @payload : JSON.generate(@payload)].join("\n")
       end
 
       def serialize
@@ -79,8 +78,8 @@ module Sentry
       @items = []
     end
 
-    def add_item(headers, payload, is_json: true)
-      @items << Item.new(headers, payload, is_json: is_json)
+    def add_item(headers, payload)
+      @items << Item.new(headers, payload)
     end
 
     def item_types
--- a/sentry-ruby/lib/sentry/metrics.rb
+++ b/sentry-ruby/lib/sentry/metrics.rb
@@ -5,12 +5,27 @@ require 'sentry/metrics/counter_metric'
 require 'sentry/metrics/distribution_metric'
 require 'sentry/metrics/gauge_metric'
 require 'sentry/metrics/set_metric'
+require 'sentry/metrics/aggregator'
 
 module Sentry
   module Metrics
     class << self
       # TODO-neel-metrics define units, maybe symbols
-      def incr(key, value: 1.0, unit: 'none', tags: nil, timestamp: nil)
+
+      def incr(key, value = 1.0, unit = 'none', tags: {}, timestamp: nil)
+        Sentry.metrics_aggregator&.add(:c, key, value, unit, tags: tags, timestamp: timestamp)
+      end
+
+      def distribution(key, value, unit = 'none', tags: {}, timestamp: nil)
+        Sentry.metrics_aggregator&.add(:d, key, value, unit, tags: tags, timestamp: timestamp)
+      end
+
+      def set(key, value, unit = 'none', tags: {}, timestamp: nil)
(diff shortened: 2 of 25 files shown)


sentry-ruby/lib/sentry/metrics.rb
```diff
@@ -0,0 +1,30 @@
...
+require 'sentry/metrics/counter_metric'
+require 'sentry/metrics/distribution_metric'
+require 'sentry/metrics/gauge_metric'
+require 'sentry/metrics/set_metric'
+require 'sentry/metrics/aggregator'
+
+module Sentry
+  module Metrics
+    class << self
+      def incr(key, value = 1.0, unit: 'none', tags: {}, timestamp: nil)
```
cleptric, line 13: While Python nor the docs are updated (yet), we agreed on calling it `increment()`. Up to you.

sentry-ruby/lib/sentry/metrics/aggregator.rb
```diff
@@ -0,0 +1,185 @@
...
+      # exposed only for testing
+      attr_reader :thread, :buckets, :flush_shift
+
+      def initialize(configuration, client)
+        @client = client
+        @logger = configuration.logger
+
+        @default_tags = {}
+        @default_tags['release'] = configuration.release if configuration.release
+        @default_tags['environment'] = configuration.environment if configuration.environment
```
cleptric, line 30: Do we have the fallback to `production` somewhere?

cleptric (commented):
Are the units defined somewhere? Or do we rely on people just following docs for the supported default units?

sentry-ruby/lib/sentry/metrics/aggregator.rb
```diff
@@ -0,0 +1,185 @@
...
+
+      # serialize buckets to statsd format
+      def serialize_buckets(buckets)
+        buckets.map do |timestamp, timestamp_buckets|
+          timestamp_buckets.map do |metric_key, metric|
+            type, key, unit, tags = metric_key
+            values = metric.serialize.join(':')
+            sanitized_tags = tags.map { |k, v| "#{sanitize_key(k)}:#{sanitize_value(v)}" }.join(',')
+
+            "#{sanitize_key(key)}@#{unit}:#{values}|#{type}|\##{sanitized_tags}|T#{timestamp}"
```
cleptric, line 154: You could omit unit if `none` and tags if empty. In case we always have an `environment` set, it does not matter ofc.

sentry-ruby/lib/sentry/metrics/aggregator.rb
sl0thentr0py: yes, but it's not production
https://github.com/getsentry/sentry-ruby/blob/c1217cd450adb5343ed287aa92b8a6d45c5bb91d/sentry-ruby/lib/sentry/configuration.rb#L637-L639

sentry-ruby/lib/sentry/metrics/aggregator.rb
sl0thentr0py: i followed whatever python does

sl0thentr0py:
will add unit lists in separate PR

sentry-ruby/lib/sentry/metrics/aggregator.rb
```diff
@@ -0,0 +1,185 @@
...
+        @mutex.synchronize do
+          @buckets[bucket_timestamp] ||= {}
+
+          if @buckets[bucket_timestamp][bucket_key]
+            @buckets[bucket_timestamp][bucket_key].add(value)
+          else
+            @buckets[bucket_timestamp][bucket_key] = METRIC_TYPES[type].new(value)
+          end
+        end
+      end
```
antonpirker, line 71: I guess considering a force flush when the new weight exceeds the limit follows in another PR?

sentry-ruby/lib/sentry/metrics/aggregator.rb
```diff
@@ -0,0 +1,185 @@
...
+
+          if force
+            flushable_buckets = @buckets
+            @buckets = {}
+          else
+            cutoff = Sentry.utc_now.to_i - ROLLUP_IN_SECONDS - @flush_shift
+            flushable_buckets = @buckets.select { |k, _| k <= cutoff }
+            @buckets.reject! { |k, _| k <= cutoff }
+          end
+
```
antonpirker, line 141: Calculating the new total weight of remaining buckets. I guess this is also coming in a later PR?

antonpirker (commented):
Some comments.

sentry-ruby/lib/sentry/metrics/aggregator.rb
sl0thentr0py: yes

sentry-ruby/lib/sentry/metrics/aggregator.rb
sl0thentr0py: yes

sentry-ruby/spec/sentry/metrics/aggregator_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+
+      it 'returns early and does nothing with force' do
+        expect(sentry_envelopes.count).to eq(0)
+        subject.flush(force: true)
+      end
+    end
+
+    context 'with pending buckets' do
+      before do
+        allow(Time).to receive(:now).and_return(fake_time)
```
st0012, line 207: Maybe we should start using [timecop](https://github.com/travisjeffery/timecop) for these tests.

sentry-ruby/lib/sentry/envelope.rb
```diff
@@ -7,19 +7,20 @@ class Item
...
+        @is_json = is_json
       end
 
       def type
         @headers[:type] || 'event'
       end
 
       def to_s
-        [JSON.generate(@headers), JSON.generate(@payload)].join("\n")
+        [JSON.generate(@headers), @is_json ? JSON.generate(@payload) : @payload].join("\n")
```
st0012, line 23: Before this PR, every item's `@payload` is a hash. And if I understand it correctly, metrics payload would be strings instead (as we call `bytesize` on them)? If that's the case, maybe we can use `@payload` type as the condition here and don't need the `is_json` attribute for `Item`? Like:

```rb
[JSON.generate(@headers), @payload.is_a?(String) ? @payload : JSON.generate(@payload)].join("\n")
```

sentry-ruby/lib/sentry/metrics/aggregator.rb
```diff
@@ -0,0 +1,185 @@
...
+      def kill
+        log_debug('[Metrics::Aggregator] killing thread')
+
+        @exited = true
+        @thread&.kill
+      end
+
+      private
+
+      def ensure_thread
```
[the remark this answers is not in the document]
sl0thentr0py: can do that separately
https://github.com/getsentry/sentry-ruby/issues/2261

sentry-ruby/spec/sentry/metrics/aggregator_spec.rb
sl0thentr0py: for now it's simple enough, if you really want I can refactor all the time related tests separately.

antonpirker (approved):
From my limited understanding of Ruby this looks OK. Could not find anything that was screaming at me. (It also looks quite similar to the Python implementation.)

sl0thentr0py:
@st0012 im gonna start merging these because we want to ship this week. If you have things you want changed, feel free to leave reviews anyway, I can change them separately.

Issue #2249, opened 2024-02-17
sentry-rails 5.15.2 to 5.16.1 update breaks gitlab - TZInfo::InvalidTimezoneIdentifier: Invalid identifier: (GMT+00:00) UTC


pravi:
### Issue Description

When sentry-rails is updated from 5.10.0 to 5.16.1 (later we found upto 5.15.2 is good) in gitlab, all rake tasks fails with `TZInfo::InvalidTimezoneIdentifier: Invalid identifier: (GMT+00:00) UTC`
https://gitlab.com/gitlab-org/gitlab/-/merge_requests/144973

### Reproduction Steps

For now this gitlab MR reproduces the error

https://gitlab.com/gitlab-org/gitlab/-/merge_requests/144973

### Expected Behavior

This should not break gitlab or any rails app

### Actual Behavior

I think sentry is adding extra '(GMT+00:00)' to a valid timezone 'UTC'. I tested with many other timezone values in config/application.rb and all fails.

### Ruby Version

3.1

### SDK Version

3.1

pravi:
If we can confirm this is a bug in gitlab, then we can close this issue.

sl0thentr0py:
hey, here it says it's the string so I think this is on your end
```ruby
# application.rb:
class Application < Rails::Application
  config.time_zone = 'Eastern Time (US & Canada)'
end
```
https://api.rubyonrails.org/classes/ActiveSupport/TimeZone.html

hoverjet:
I have a similar issue during building stage. Downgrading to 5.15.2 helped.

```
bin/rails aborted!
TZInfo::InvalidTimezoneIdentifier: Invalid identifier: nil (TZInfo::InvalidTimezoneIdentifier)
/app/vendor/bundle/ruby/3.2.0/gems/tzinfo-2.0.6/lib/tzinfo/data_source.rb:316:in `validate_timezone_identifier'
/app/vendor/bundle/ruby/3.2.0/gems/tzinfo-2.0.6/lib/tzinfo/data_sources/zoneinfo_data_source.rb:307:in `load_timezone_info'
/app/vendor/bundle/ruby/3.2.0/gems/tzinfo-2.0.6/lib/tzinfo/data_source.rb:195:in `get_timezone_info'
/app/vendor/bundle/ruby/3.2.0/gems/tzinfo-2.0.6/lib/tzinfo/timezone.rb:128:in `get'
[...]

PR #2250, opened 2024-02-18
Read time_zone.name instead of time_zone

pravi:
## Description
Without adding time_zone.name, the returned value is considered invalid by tzdata.

Closes: #2249

### DIFF
--- a/sentry-rails/lib/sentry/rails/railtie.rb
+++ b/sentry-rails/lib/sentry/rails/railtie.rb
@@ -72,8 +72,8 @@ module Sentry
     end
 
     def configure_cron_timezone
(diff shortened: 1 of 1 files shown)


pravi:
Not that simple, will try a different fix.

pravi:
This looks like a bug in gitlab now, as https://gitlab.com/gitlab-org/gitlab/-/blob/master/config/initializers/time_zone.rb?ref_type=heads sets

`Rails.application.config.time_zone = Time.zone`

```
puts Time.zone
(GMT+00:00) UTC
```
Which is not a valid identifier.

sl0thentr0py:
im closing this, please reopen if necessary

PR #2251, opened 2024-02-20
Fix mention of renamed attribute

c960657:
`rails_report_rescued_exceptions` was changed to `rails.report_rescued_exceptions` in #1070. The inline docuentation was copied without updating the attribute name.


st0012 (approved):
Good catch, thx 👍

PR #2255, opened 2024-03-06
Metric summaries on span

sl0thentr0py:
* add a `LocalAggregator` instance on spans to duplicate metrics on the span as a gauge metric
* proxy the main aggregator add calls to the local aggregator if a span is running
* start a `metric.timing` span in the `Sentry::Metrics.timing` API

part of https://github.com/getsentry/sentry-ruby/issues/2246


antonpirker (approved):
Looks like it does the same as Python. (Again, I am no Ruby dev so some things (in tests) look like black magic) But I guess this is ok.

PR #2258, opened 2024-03-08
Add before_emit callback for metrics

sl0thentr0py:
part of #2246

### DIFF
--- a/sentry-ruby/lib/sentry/utils/argument_checking_helper.rb
+++ b/sentry-ruby/lib/sentry/utils/argument_checking_helper.rb
@@ -15,5 +15,11 @@ module Sentry
         raise ArgumentError, "expect the argument to be one of #{values.map(&:inspect).join(' or ')}, got #{argument.inspect}"
       end
     end
+
+    def check_callable!(name, value)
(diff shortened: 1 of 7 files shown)


sentry-ruby/lib/sentry/utils/argument_checking_helper.rb
```diff
@@ -15,5 +15,11 @@ def check_argument_includes!(argument, values)
         raise ArgumentError, "expect the argument to be one of #{values.map(&:inspect).join(' or ')}, got #{argument.inspect}"
       end
     end
+
+    def check_callable!(name, value)
```
st0012, line 19: Unless we expect users to pass a command object here (Non-Proc objects that define `call` as the unify interface), we can use `check_argument_type!(argument, ::Proc, NilClass)` without introducing a new method. (Which I should've done for Configuration a while ago).

sl0thentr0py: this method was just pulled out of configuration.rb, we can do that refactoring separately

antonpirker (approved):
Look good from my side

PR #2269, opened 2024-03-14
Fix NoMethodError / Make session_tracking check consistent

knapo:
## Description
https://github.com/getsentry/sentry-ruby/pull/2245 (cc @st0012 ) introduced changes around checking if session_tracking is enabled or not.. However, it wasn't updated in all necessary places and make it inconsistent within the codebase. As result, when not `enabled_in_current_env?` we're getting a following error:

```
     NoMethodError:
       undefined method `add_session' for nil:NilClass
     # <REDACTED>/gems/sentry-ruby-5.17.0/lib/sentry/hub.rb:244:in `end_session'
     # <REDACTED>/gems/sentry-ruby-5.17.0/lib/sentry/hub.rb:253:in `with_session_tracking'
     # <REDACTED>/gems/sentry-ruby-5.17.0/lib/sentry-ruby.rb:403:in `with_session_tracking'
```
     
Affected version: `sentry-ruby 5.17.0`

PR #2275, opened 2024-03-17
fix: Don't mutate original enabled_environments

ixti:
Ensure enabled_environments contains test environment without mutating original list. Otherwise it can cause unexpected spec failures in case original config has unfrozen Array, or fail in attempt to mutate said array.

Alternative approach would be to override `Sentry::Configuration#dup` to dup underlying arrays and hashes.

See: https://github.com/getsentry/sentry-ruby/pull/2269#issuecomment-2002637738

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Bug Fixes
 
 - Don't instantiate connection in `ActiveRecordSubscriber` ([#2278](https://github.com/getsentry/sentry-ruby/pull/2278))
+- Don't mutate original `enabled_environments` when using test helpers ([#2275](https://github.com/getsentry/sentry-ruby/pull/2275))
 
 ## 5.17.1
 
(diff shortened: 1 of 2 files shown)


st0012:
Thanks for the PR, but why did you find it necessary to freeze the `enabled_environments` array in the first place?

ixti:
My tests were randomly passing randomly failing. Adding a freeze helped to nail down that Sentry helper was changing configuration inplace - thus specs became order-dependent

Vagab:
we have ran into the same problem. This pr didn't fix it(not sure why?, will investigate), but I have also noticed that the origin of the issue is mutation of that array

sl0thentr0py:
superseded by #2317

ixti:
I still think `|=` better reflects the desire than `+=` :D

Vagab:
didn’t work for me :(

ixti:
> didn’t work for me :(

Oh. That's weird. Either way, as long as Sentry does not mutate original config - we're good :D

:heart: :heart: :heart:

PR #2278, opened 2024-03-20
Fix: Get connection from connection_pool instead of ActiveRecord::Base.connection

Iwaide:
# Issue
When attempting to retrieve a connection in `ActiveRecordSubscriber`, a loop occurs, eventually leading to `ActiveRecord::ConnectionTimeoutError`.

In versions of Rails lower than 6, since the payload does not contain a connection, it tries to use `ActiveRecord::Base.connection` to get the current connection. However, it seems that this causes `subscribe!` to be called again.
I am not sure if this issue is specific to MySQL. Please refer to the following repository for a reproduction environment:
https://github.com/Iwaide/re_app_for_sentry/

# Fix
I have modified the code to search for and retrieve a connection from `ActiveRecord::Base.connection_pool.connections`.
I am not very familiar with connection management, so I am not entirely confident that this is the appropriate approach. Please make any necessary corrections.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Bug Fixes
+
+- Don't instantiate connection in `ActiveRecordSubscriber` ([#2278](https://github.com/getsentry/sentry-ruby/pull/2278))
+
 ## 5.17.1
 
 ### Bug Fixes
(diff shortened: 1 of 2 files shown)


sl0thentr0py (approved):
thanks @Iwaide !

PR #2280, opened 2024-03-26
Add mechanism interface and default to handled false in integrations

### LINKED ISSUE #1743: Add `mechanism` to the exception interfaces
Add the ability to pass [`mechanism`](https://develop.sentry.dev/sdk/event-payloads/exception/#exception-mechanism) to `capture_exception`. Among other things, this will let us use the `handled` flag to set the session status to `crashed`/`errored` based on whether it was handled or not in #1715.

sl0thentr0py:
closes #1743


sentry-ruby/lib/sentry/interfaces/single_exception.rb
```diff
@@ -29,17 +29,19 @@ def initialize(exception:, stacktrace: nil)
       @module = exception.class.to_s.split('::')[0...-1].join('::')
       @thread_id = Thread.current.object_id
       @stacktrace = stacktrace
+      @mechanism = mechanism || Mechanism.new
```
st0012, line 32: Can we move this default assignment upstream? I think if we assign a default in `Hub#capture_exception` as well, then we can just assume `mechanism` will always be present when it reached client and interfaces?

sentry-ruby/lib/sentry/interfaces/single_exception.rb
```diff
@@ -29,17 +29,19 @@ def initialize(exception:, stacktrace: nil)
...
     end
 
     def to_hash
       data = super
       data[:stacktrace] = data[:stacktrace].to_hash if data[:stacktrace]
+      data[:mechanism] = data[:mechanism].to_hash if data[:mechanism]
```
st0012, line 38: With the fallback in constructor, this should always be `true`?

sentry-ruby/lib/sentry/interfaces/single_exception.rb
sl0thentr0py: I moved it to `Client#event_from_exception` since we'll have to break the signature of that method otherwise since it uses positional and not keyword arguments.

sentry-ruby/lib/sentry/client.rb
```diff
@@ -80,8 +80,9 @@ def capture_event(event, scope, hint = {})
     # Initializes an Event object with the given exception. Returns `nil` if the exception's class is excluded from reporting.
     # @param exception [Exception] the exception to be reported.
     # @param hint [Hash] the hint data that'll be passed to `before_send` callback and the scope's event processors.
+    # @param mechanism [Mechanism] the mechanism that holds a type and whether the exception was handled or not.
     # @return [Event, nil]
-    def event_from_exception(exception, hint = {})
+    def event_from_exception(exception, hint = {}, mechanism = nil)
```
st0012, line 85: Can we make `mechanism` part of the `hint` here? 
The current approach limits the ways to pass hint, from both `key: value` and `{ key: value }`, to just the latter. IMO it's a breaking change (albeit a minor one), and it's not really necessary. If we simply treat `mechanism` the same as `ignore_exclusions` here, then we won't have the problem.

st0012 (commented):
Just one comment left, otherwise it's good.

antonpirker (approved):
rubber stamp

PR #2281, opened 2024-03-27
Use mechanism to track crashed sessions for unhandled exceptions

sl0thentr0py:
part of #1743

### DIFF
--- a/sentry-ruby/lib/sentry/session.rb
+++ b/sentry-ruby/lib/sentry/session.rb
@@ -4,9 +4,8 @@ module Sentry
   class Session
     attr_reader :started, :status, :aggregation_key
 
-    # TODO-neel add :crashed after adding handled mechanism
-    STATUSES = %i[ok errored exited]
-    AGGREGATE_STATUSES = %i[errored exited]
+    STATUSES = %i[ok errored crashed exited]
+    AGGREGATE_STATUSES = %i[errored crashed exited]
 
     def initialize
       @started = Sentry.utc_now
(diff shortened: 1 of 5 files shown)


sentry-ruby/lib/sentry/session.rb
```diff
@@ -17,9 +16,11 @@ def initialize
       @aggregation_key = Time.utc(@started.year, @started.month, @started.day, @started.hour, @started.min)
     end
 
-    # TODO-neel add :crashed after adding handled mechanism
-    def update_from_exception(_exception = nil)
+    def update_from_event(event = nil)
```
st0012, line 19: Why are we calling it `update_from_event` but in fact there may be no event? It looks like the only place where it could be `nil` is in tests, can we improve the setup there so `event` becomes a required argument?

sentry-ruby/lib/sentry/session.rb
```diff
@@ -17,9 +16,11 @@ def initialize
...
       @status = :errored
+      return unless event&.is_a?(ErrorEvent)
+
+      @status = :crashed if event.exception.values.any? { |e| e.mechanism&.handled == false }
```
st0012, line 23: From #2280, it looks like there will always be a mechanism in the exception interface, so we won't need `&` here?

sentry-ruby/lib/sentry/session.rb
```diff
@@ -17,9 +16,11 @@ def initialize
       @aggregation_key = Time.utc(@started.year, @started.month, @started.day, @started.hour, @started.min)
     end
 
-    # TODO-neel add :crashed after adding handled mechanism
-    def update_from_exception(_exception = nil)
+    def update_from_event(event)
       @status = :errored
```
st0012, line 20: Should this be moved behind the return below? Since we name it `update_from_event` now, setting the session to `errored` when it's called with any event is weird (e.g. technically `CheckInEvent` can also be passed here.)

IMO we should probably name it `update_from_exception_event` unless we plan to expand its usage for other event types too.

sl0thentr0py: done

PR #2286, opened 2024-03-31
Add generator for initializer generation

### LINKED ISSUE #2218: Add rake command for adding rails initializer
we want to make it easier for rails users to setup sentry, see
https://fly.io/docs/reference/sentry/#ruby-on-rails
and 
https://github.com/getsentry/sentry-wizard

* add new rake command like `bundle exec sentry:setup`
* auto create default initializer
* some questions TBD (see sentry-wizard and how they do it)
   * DSN
   * enable performance
   * enable profiling

st0012:
### Usage

```shell
$ bin/rails generate sentry 
$ bin/rails generate sentry --dsn my_dsn
```

Implements #2218

### DIFF
--- a/sentry-rails/lib/generators/sentry_generator.rb
+++ b/sentry-rails/lib/generators/sentry_generator.rb
@@ -1,0 +1,18 @@
+require "rails/generators/base"
+
+class SentryGenerator < ::Rails::Generators::Base
+  class_option :dsn, type: :string, desc: "Sentry DSN"
+
+  def copy_initializer_file
+    dsn = options[:dsn] ? "'#{options[:dsn]}'" : "ENV['SENTRY_DSN']"
+
+    create_file "config/initializers/sentry.rb", <<~RUBY
+      # frozen_string_literal: true
+
+      Sentry.init do |config|
+        config.breadcrumbs_logger = [:active_support_logger]
+        config.dsn = #{dsn}
+      end
+    RUBY
+  end
+end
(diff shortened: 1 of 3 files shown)


sentry-rails/lib/generators/sentry_generator.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def copy_initializer_file
+    dsn = options[:dsn] ? "'#{options[:dsn]}'" : "ENV['SENTRY_DSN']"
+
+    create_file "config/initializers/sentry.rb", <<~RUBY
+      # frozen_string_literal: true
+
+      Sentry.init do |config|
+        config.breadcrumbs_logger = [:active_support_logger]
+        config.dsn = #{dsn}
```
sl0thentr0py, line 14: we'd like to enable performance by default, so let's add
```ruby
config.enable_tracing = true
```

sl0thentr0py (changes requested):

sentry-rails/lib/generators/sentry_generator.rb
st0012: I was concerned that enabling tracing by default without setting any sampling rate/logic would cause unexpected quota consumption. But it's hard to set a good sampling strategy statically, so I decided to not go for it first.

st0012: I've updated it though.

sl0thentr0py: We're generally trying to move away from client side sampling and let people control their sampling dynamically on the server. Of course there are nuances for very large apps with 100% sampling so we will not get rid of the client side sampling configs, but we do want to encourage them less for newer users.

sl0thentr0py (approved):

PR #2298, opened 2024-04-15
Fix new patch changelog entries to internal

sl0thentr0py:
i updated the github releases directly

#skip-changelog


st0012 (approved):
Nice 👍

PR #2303, opened 2024-04-29
Notify users when their custom options are discarded

st0012:
This is a followup on #2301

It should save users from some debugging if we can tell them what options they pass are discarded and how they can fix it.

The warning is only enabled at debug level so it should not be too noisy.


sl0thentr0py (approved):
thx!

PR #2304, opened 2024-05-01
Extract thread control logic into a new `ThreadedPeriodicWorker` class

### LINKED ISSUE #2261: Unify threading logic under a base class/module

st0012:
Instead of letting BackpressureMonitor, SessionFlusher, and Metrics::Aggregator handle their own threading logic, we can extract that into a new class that can be reused by all of them.

Closes #2261

#skip-changelog


sentry-ruby/lib/sentry/threaded_periodic_worker.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+    def ensure_thread
+      return false if @exited
+      return true if @thread&.alive?
+
+      @thread = Thread.new do
+        loop do
+          sleep(@interval)
+          run
```
st0012, line 21: I prefer using the generic `run` instead of `flush` as the API contract of this abstract class. Individual subclasses and simply alias `run` to `flush`.

sl0thentr0py (approved):
ty! <3

PR #2306, opened 2024-05-06
Remove `config.capture_exception_frame_locals`

st0012:
It's been marked as deprecated for a while now and should be removed in the next major release.

Additional note: some TracePoint events can invalidate YJIT's optimization, which was a main concern of the current implementation around locals capturing. But at least until Ruby 3.3, the `raise` event we're using is NOT one of them.


sl0thentr0py:
i'll take care of this separately as i'm going through all deprecations for the major

PR #2311, opened 2024-05-14
Make links more specific in gemspecs

AlexWayfer:
> - Finally, please add an entry to the corresponding changelog

I'm not sure should I add an entry to the changelogs, and each of them or just one.

## Description

I believe each separate gem should have correct (direct, specific) links in its gemspec. In this architecture with single-repo it's achievable. Also it'll help us to list dependencies in our project without common duplicates to the root of the repo.

Additionally gems with specific version (such as  `5.17.3`) should point to documentation, changelog, sources, etc. of this version, not upstream (`master`).

Also links to `rubydoc.info` and GitHub issues were added.

#skip-changelog


sl0thentr0py (approved):
thx @AlexWayfer fixed the changelogs slightly

PR #2313, opened 2024-05-26
Add query source support to `sentry-rails`

### LINKED ISSUE #2242: Add "Query Sources" Ruby support
### Problem Statement

[Query Sources](https://docs.sentry.io/product/performance/queries/#query-sources) are supported in Python frameworks and in Laravel, but not in Ruby. Opening this issue so we can keep track of customer requests, which have come up a few times already.

### Solution Brainstorm

There's some prior art on this in our own SDKs _and_ in the OSS world, since Rails has built-in query source collection. Maybe we can hook into that, or just do something similar.

### Product Area

Performance - Queries

st0012:
Closes #2242

This PR adds query source information to the tracing spans captured by `sentry-rails`'s `ActiveSupportSubscriber`, which should be picked up by Sentry and displayed like:

Screenshot 2024-05-26 at 16 43 23

The implementation is inspired by Rails' [`ActiveRecord::LogSubscriber`](https://github.com/rails/rails/blob/1b3fc3c82e36a1c5f19f174e318166a11bd0b301/activerecord/lib/active_record/log_subscriber.rb) but doesn't depend on it.

There are several requirements for this feature to work:
- Ruby 3.2+ because we need `Thread.each_caller_location` to minimise the performance impact
- Rails 7.1+ because we need `ActiveSupport::BacktraceCleaner#clean_frame`
- `config.rails.enable_db_query_source` is set to `true` (default)
- `sentry-rails`'s `ActiveRecordSubscriber` is included in the `config.rails.tracing_subscribers` list (default)

And to reduce the feature's overhead, only queries that exceed `config.rails.db_query_source_threshold_ms` will have source recorded, which is set to 100ms by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,11 @@
       config.enabled_patches += [:graphql]
     end
     ```
+
 - Add `Sentry.get_trace_propagation_meta` helper for injecting meta tags into views ([#2314](https://github.com/getsentry/sentry-ruby/pull/2314))
+- Add query source support to `sentry-rails` ([#2313](https://github.com/getsentry/sentry-ruby/pull/2313))
+
+  The feature is only activated in apps that use Ruby 3.2+ and Rails 7.1+. By default only queries that take longer than 100ms will have source recorded, which can be adjusted by updating the value of `config.rails.db_query_source_threshold_ms`.
 
 ### Bug Fixes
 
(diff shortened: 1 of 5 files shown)


sl0thentr0py (changes requested):
Thanks for adding this!
We also have an option called `db_query_source_threshold_ms` that defaults to 100ms and the source is recorded on the span only if it takes more than that threshold.
Could you also add that?

https://github.com/getsentry/sentry-python/blob/b922f40ad9cff1f97286030f71a22e0000e9b4e2/sentry_sdk/tracing_utils.py#L185-L186

sl0thentr0py (approved):
ty <3

PR #2334, opened 2024-06-27
feat: allow to set custom status code for different exception classes

### LINKED ISSUE #2333: Custom transaction status code for exceptions
**Describe the idea**

The idea is to allow us to set custom status code for when a exception occurs during a transaction, something like

```ruby
Sentry.init do |config|
  config.exception_status_code = ->(exception) do
    case exception
    when ActiveRecord::RecordNotFound, ActionController::RoutingError
      404
    else
      500
    end
  end
end
```

Sentry assumes status code 500 (internal error) for all of the exceptions during transactions. The problem is that in a production environment, by default Rails will raise `ActionController::RoutingError` whenever somebody tries to access a non-existing page, which will result in "internal error" transactions.

This feature allows the user to map different exceptions to different status codes, so they can ignore that without having to change the default Rails behavior.

**Possible implementation**

In the `Sentry::Rack::CaptureExceptions` we can replace the hardcoded 500 status code from the `rescue Exception` with a configurable lambda that returns the desired status code for the specific exception.

Another possibility, is to bake this rule into Sentry (404 for routing errors), but I think that would be a shift in the current behavior which might not be ideal

gustavodiel:
Fixes #2333

## Description
This PR adds a new option to specify the status code for any unhandled exception during a transaction

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@
   The feature is only activated in apps that use Ruby 3.2+ and Rails 7.1+. By default only queries that take longer than 100ms will have source recorded, which can be adjusted by updating the value of `config.rails.db_query_source_threshold_ms`.
 - Log envelope delivery message with debug instead of info ([#2320](https://github.com/getsentry/sentry-ruby/pull/2320))
 
+- Add `exception_status_code` configuration option that allows to specify which transaction status code should be used for exceptions ([#2333](https://github.com/getsentry/sentry-ruby/issues/2333))
+
 ### Bug Fixes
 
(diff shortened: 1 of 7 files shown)


mhfs:
Could someone on the Sentry side review this, please? 🙏

solnic:
Hey thanks for the PR! I'm wondering if this could also potentially help with #2451?

gustavodiel:
@solnic

Hmm I don't think so, because I think it's being filtered before this step.
Once that's fixed, we can use this PR to flag which status code this error should get on Sentry

PR #2337, opened 2024-07-03
Ignore Sidekiq::JobRetry::Handled exception

karlentwistle:
Adds `Sidekiq::JobRetry::Handled` exception to default Sidekiq ignore list.

`Sidekiq::JobRetry::Skip` exceptions are currently ignored, however, since Sidekiq 7.3.0 shipped, I think `Sidekiq::JobRetry::Handled` exceptions also need to be ignored.

You can see the full diff between [7.2.4..7.3.0](https://github.com/sidekiq/sidekiq/compare/30786e082c70349ab27ffa9eccc42fb0c696164d...602d04640c7dd0c16222800f24bd3d3c32031a98) and I'd particularly like to draw your attention to the changes within `lib/sidekiq/job_retry.rb`. 

You can see in certain scenarios Sidekiq now raises [`Sidekiq::JobRetry::Handled`](https://github.com/sidekiq/sidekiq/blob/602d04640c7dd0c16222800f24bd3d3c32031a98/lib/sidekiq/job_retry.rb#L106) (7.3.0) instead of [`Sidekiq::JobRetry::Skip`](https://github.com/sidekiq/sidekiq/blob/30786e082c70349ab27ffa9eccc42fb0c696164d/lib/sidekiq/job_retry.rb#L132) (7.2.4).

Without this change, some customers will see the following in their Sentry Issues after upgrading to Sidekiq 7.3.0.

Screenshot 2024-07-03 at 08 27 40

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Bug Fixes
 
+- Ignore internal Sidekiq::JobRetry::Handled exception [#2337](https://github.com/getsentry/sentry-ruby/pull/2337)
 - Fix Vernier profiler not stopping when already stopped [#2429](https://github.com/getsentry/sentry-ruby/pull/2429)
 
 ## 5.21.0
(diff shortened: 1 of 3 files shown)


natematykiewicz:
+1. I have a bunch of these in my Sentry as well, and they're always next to a second error event that's the actual problem. They can be safely ignored.

solnic (approved):
This makes sense. Thank you!

Issue #2297, opened 2024-04-11
Consider avoiding Concurrent.processor_count when setting default background_worker_threads


trevorturk:
### Issue Description

Since https://github.com/getsentry/sentry-ruby/pull/1155 we have the ability to configure `background_worker_threads` which currently defaults to `Concurrent.processor_count. However, this config is likely to be too high in a shared hosting environment, leading to unnecessary memory usage. It may make sense to change the default to, say, a static 1 or 3 and add documentation about how best to configure?

See also: https://github.com/rails/rails/pull/50669/files#diff-e25cab4832f078ca2f9dffa8918f895c8474f8086878e4ced9f5795131fc2de2R32

/cc @st0012 

Feel free to disregard, since this is easy to correct, but in my case I was surprised at the impact lowering from 8 to 1 made on my memory use on Heroku. 

### Reproduction Steps

```
$ heroku run rails c
Running rails c on ⬢ APP_NAME... up, run.4033 (Standard-1X)
Loading production environment (Rails 7.1.3.2)
irb(main):001> Concurrent.processor_count
=> 8
```

### Expected Behavior

Consider defaulting to a static/lower number of background_worker_threads.

### Actual Behavior

Defaults to `Concurrent.processor_count`. 

### Ruby Version

3.3.3

### SDK Version

5.17.3

trevorturk:
I was thinking about this a bit more (apologies if this is more complicating than helping) but it seems to me that if we did 1 thread by default, the only real risk is that you might have more than 30 (the default) items in the queue at which point they'd start being dropped (at least that's my understanding according to the docs). My thinking is that we could try pushing down to 1 thread and sending an event (or logs or something) if we see items being dropped, so the customer would know they need to increase the thread count?

st0012:
I agree that we should probably lower the worker threads, thanks for raising this.

> My thinking is that we could try pushing down to 1 thread and sending an event (or logs or something) if we see items being dropped, so the customer would know they need to increase the thread count?

This potentially means that some customers would need to redeploy (potentially multiple times) to adjust the thread count after we push such changes, which would likely be a pretty negative experience. Considering that the current default has existed for more than 3 years and we only have received a report now, IMO the potential benefit doesn't justify the risk.

> I believe anything between 2-4 threads as a default might be ok.

How about we start with `Concurrent.processor_count / 2`?

[...]

PR #2339, opened 2024-07-06
Use Concurrent.usable_processor_count when it is available

st0012:
It's a new API introduced in concurrent-ruby 1.3.1, which works better in the container environment.

https://github.com/ruby-concurrency/concurrent-ruby/pull/1038

Since there are gems like sorbet-runtime that still use older versions of concurrent-ruby, we can't directly bump concurrent-ruby's requirement, but need to check if the method is available before calling it.

Big thanks to @trevorturk for [proposing this idea](https://github.com/getsentry/sentry-ruby/issues/2297#issuecomment-2137623858).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Internal
+
+- Use Concurrent.usable_processor_count when it is available ([#2339](https://github.com/getsentry/sentry-ruby/pull/2339))
+
 ## 5.18.1
 
 ### Bug Fixes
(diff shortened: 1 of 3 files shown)


trevorturk:
Nice, thanks! I wonder if we should move that `/ 2.0` into the new helper method since we can (hopefully) trust the new container version to do the right thing? Either way looks good enough. Thank you!

sl0thentr0py (approved):

sl0thentr0py:
ty both!

PR #2345, opened 2024-07-18
Support for tracing Faraday requests

### LINKED ISSUE #1795: Add tracing support for faraday requests
Right now we just have `Net:HTTP` support for outgoing http requests and if Faraday uses some other adapter we don't get any spans.
https://github.com/lostisland/awesome-faraday/#adapters

solnic:
This adds support for automatic tracing of HTTP requests when using Faraday. It leverages its instrumentation middleware to do the magic.

Screenshot 2024-07-25 at 12 52 21

---

## Todo

- [x] Figure out how to use the instrumentation middleware
- [x] ~Add a Sentry middleware that will set everything up~ turned out we needed a monkeypatch after all
- [x] Figure out how to enable the integration automatically
- [x] Cover the same cases that are specified by the net/http_spec
- [x] ~(optional/stretch) unify implementation between net/http MP and faraday extension~ if we decide to do it, it should go to a separate PR
- [x] Figure out what to do with the default net/http patch

Closes #1795


sentry-ruby/lib/sentry/faraday.rb
```diff
@@ -0,0 +1,52 @@
...
+      #  ssl=#<struct Faraday::SSLOptions verify=nil, verify_hostname=nil, ca_file=nil, ca_path=nil, verify_mode=nil, cert_store=nil, client_cert=nil, client_key=nil, certificate=nil, private_key=nil, verify_depth=nil, version=nil, min_version=nil, max_version=nil>,
+      #  parallel_manager=nil,
+      #  params=nil,
+      #  response=nil,
+      #  response_headers=nil,
+      #  status=nil,
+      #  reason_phrase=nil,
+      #  response_body=nil>]
+      def self.instrument(name, env, &block)
+        Sentry.with_child_span(op: "http.faraday", start_timestamp: Sentry.utc_now.to_f, origin: "auto.http.faraday") do |sentry_span|
```
sl0thentr0py, line 41: we have some conventions here
https://develop.sentry.dev/sdk/performance/span-operations/#web-server
so this should be `http.client` or `http.client.faraday` if you want to be more precise.

sl0thentr0py:
looks great so far!

as you noted in the description, you need to further
* [add a breadcrumb](https://github.com/getsentry/sentry-ruby/blob/36866c54ea67933c36a5d8ce2b52dae972e3edd1/sentry-ruby/lib/sentry/net/http.rb#L42)
* [propagate the tracing headers](https://github.com/getsentry/sentry-ruby/blob/36866c54ea67933c36a5d8ce2b52dae972e3edd1/sentry-ruby/lib/sentry/net/http.rb#L37-L39) (`sentry-trace` and `baggage`
  * [docs here](https://develop.sentry.dev/sdk/distributed-tracing/) are slightly outdated because we don't add `traceparent` anymore, I will try to clean those up
* extract relevant request data 
  * the span data field keys recorded in [`DataConventions`](https://github.com/getsentry/sentry-ruby/blob/36866c54ea67933c36a5d8ce2b52dae972e3edd1/sentry-ruby/lib/sentry/span.rb#L10-L47) here are kept in sync with [OpenTelemetry Semantic Conventions](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client)

sentry-ruby/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 Individual gem's changelog has been deprecated. Please check the [project changelog](https://github.com/getsentry/sentry-ruby/blob/master/CHANGELOG.md).
 
+## Unreleased
+
+- Support for tracing Faraday requests ([#2345](https://github.com/getsentry/sentry-ruby/pull/2345))
```
sl0thentr0py, line 7: we need to add this in the root `CHANGELOG.md`

sl0thentr0py (changes requested):

sl0thentr0py:
@solnic yes please!

sentry-ruby/lib/sentry/faraday.rb
```diff
@@ -0,0 +1,108 @@
...
+    module Connection
+      def new(*)
+        connection = super
+
+        unless connection.builder.adapter === ::Faraday::Adapter::NetHttp
+          connection.request :instrumentation, name: OP_NAME, instrumenter: Instrumenter.new
+        end
+
+        # Ensure that the instrumentation middleware is the first one in the stack
+        connection.builder.handlers.reverse!
```
solnic, line 20: We need to do this so that the instrumentation will measure the entire execution stack, not just sending a request, unless we actually do want to only measure the request?

sl0thentr0py: okay so this is bad because any monkeypatch has to be _idempotent_, i.e. leave userland behaviour unchanged. This breaks that contract. Instead of `reverse`, can we just make sure our patch is the first in the list?

That doesn't change the ordering of the existing stuff the user might have added.

solnic: @sl0thentr0py the trick is that we don't know the adapter until we yield. I guess I could add the instrumentation and then remove it if the adapter ends up being net/http. Does it make sense?

sl0thentr0py: you can just add the instrumentation always and early return in `instrument` below if the adapter is net/http.
My point is about the `reverse!` because that changes userland behaviour if they also added stuff in the `handlers` list. 
But I also need to read that corresponding faraday code because I don't know how that's working, will get back in a bit.

solnic: @sl0thentr0py oh d'ooh of course, yeah I just wanted to put it first, but reverse changes the whole handlers array. I will fix it / make it better!

sentry-ruby/lib/sentry/faraday.rb
```diff
@@ -0,0 +1,116 @@
...
+          # because if is is, then the net/http instrumentation will take care of it
+          if connection.builder.adapter.name != "Faraday::Adapter::NetHttp"
+            connection.request :instrumentation, name: OP_NAME, instrumenter: Instrumenter.new
+
+            # Make sure that it's going to be the first middleware so that it can capture
+            # the entire request processing involving other middlewares
+            handlers = connection.builder.handlers
+            instrumentation = handlers.pop
+
+            handlers.unshift(instrumentation)
```
sl0thentr0py, line 26: `handlers` actually exposes an `insert` method (which is more low level than `use` sadly) so maybe we can change it later and avoid the array weirdness but it's fine for now this way 
https://github.com/lostisland/faraday/blob/4860f75372bdfac5612c76d76bd713e72b3f45b3/lib/faraday/rack_builder.rb#L118-L123

sentry-ruby/lib/sentry/faraday.rb
```diff
@@ -0,0 +1,116 @@
...
+      end
+
+      def instrument(op_name, env, &block)
+        return unless Sentry.initialized?
+
+        Sentry.with_child_span(op: op_name, start_timestamp: Sentry.utc_now.to_f, origin: SPAN_ORIGIN) do |sentry_span|
+          request_info = extract_request_info(env)
+
+          if propagate_trace?(request_info[:url])
+            set_propagation_headers(env[:request_headers])
```
sl0thentr0py, line 46: just to make sure, the modified `env` with the extra headers gets used in the actual request right?

sentry-ruby/lib/sentry/faraday.rb
```diff
@@ -0,0 +1,116 @@
...
+    OP_NAME = "http.client"
+    SPAN_ORIGIN = "auto.http.faraday"
+    BREADCRUMB_CATEGORY = "http"
+
+    module Connection
+      # Since there's no way to preconfigure Faraday connections and add our instrumentation
+      # by default, we need to extend the connection constructor and do it there
+      #
+      # @see https://lostisland.github.io/faraday/#/customization/index?id=configuration
+      def new(url = nil, options = nil)
```
sl0thentr0py, line 14: nit: 
any reason for overriding `new` and `extend`ing rather than overriding `initialize` and `prepend`ing ? just a bit cleaner since most of our patches are `prepend`ing

sentry-ruby/spec/sentry/faraday_spec.rb
```diff
@@ -0,0 +1,255 @@
...
+
+    let(:url) { "http://example.com" }
+
+    it "skips instrumentation" do
+      transaction = Sentry.start_transaction
+      Sentry.get_current_scope.set_span(transaction)
+
+      _response = http.get("/test")
+
+      request_span = transaction.span_recorder.spans.last
```
sl0thentr0py, line 249: I guess we should also check that none of the spans are from faraday either?

sl0thentr0py (changes requested):
great first PR @solnic! 
just a few nits and comments

sentry-ruby/lib/sentry/faraday.rb
sl0thentr0py: nvm there's a spec that covers this, ignore

sl0thentr0py (approved):
perfect, ty and grats on landing your first PR!

solnic:
Ahh yes! Thanks for the reviews!

PR #2349, opened 2024-07-22
Suppress the unnecessary “unsupported options notice”

moznion:
A related change was introduced in #2303.

## Description

In `Scope#update_from_options()`, the method strips key-value pairs from the `options` according to its parameters. When all keys in `options` are supported, it still logs an “unsupported options notice” for an empty `unsupported_option_keys` value with empty array literal in `Sentry::Hub#capture_event`, like: "Options [] are not supported and will not be applied to the event."

**Example:**

When calling `subject.capture_event(event, level: 'DEBUG')`, the `capture_event` method should handle the options like this:

```ruby
# In this case, options == {:level=>'DEBUG'}
unsupported_option_keys = scope.update_from_options(**options)
# unsupported_option_keys should be [], but the following debug log will be shown
# like "Options [] are not supported and will not be applied to the event."
configuration.log_debug <<~MSG
  Options #{unsupported_option_keys} are not supported and will not be applied to the event.
  You may want to set them under the `extra` option.
MSG
```

This patch changes the logic to check whether `unsupported_option_keys` is empty before logging the notice, thus suppressing unnecessary logs.

### DIFF
--- a/sentry-ruby/spec/sentry/hub_spec.rb
+++ b/sentry-ruby/spec/sentry/hub_spec.rb
@@ -350,7 +350,7 @@ RSpec.describe Sentry::Hub do
       expect(string_io.string).to include("Options [:unsupported] are not supported and will not be applied to the event.")
     end
 
-    it "should not show the unsupported options notice even if it has a special parameter" do
+    it "does not warn about unsupported options if all passed options are supported" do
       expect do
         subject.capture_event(event, level: 'DEBUG')
       end.not_to raise_error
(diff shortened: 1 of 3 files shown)


sentry-ruby/spec/sentry/hub_spec.rb
```diff
@@ -350,6 +350,14 @@
       expect(string_io.string).to include("Options [:unsupported] are not supported and will not be applied to the event.")
     end
 
+    it "should not show the unsupported options notice even if it has a special parameter" do
```
st0012, line 353:
```suggestion
    it "does not warn about unsupported options if all passed options are supported" do
```

st0012 (approved):
Thanks for the fix 👍

PR #2350, opened 2024-07-22
Don't overwrite ip_address if already set on user

### LINKED ISSUE #2347: User ip_address Overwritten Despite Being Set via Sentry.set_user
### Issue Description

When the `@send_default_pii` flag is set to true, the `ip_address` set via `Sentry.set_user` is being overwritten by the IP address calculated from the Rack environment (`calculate_real_ip_from_rack`). This behavior is not intended and leads to incorrect IP address information being sent to Sentry.

After attempting to disable the send_default_pii flag, the ip_address set via Sentry.set_user was correctly reflected in Sentry. This confirms that the issue is related to the send_default_pii flag causing the IP address to be overwritten by `calculate_real_ip_from_rack`.

https://github.com/getsentry/sentry-ruby/blob/36866c54ea67933c36a5d8ce2b52dae972e3edd1/sentry-ruby/lib/sentry/event.rb#L103-L115

### Reproduction Steps

```ruby
  Sentry.set_user(
    ip_address: '123.123.123.123'
  )
```

### Expected Behavior

The ip_address set via Sentry.set_user using request.remote_ip should be correctly recorded and not overwritten by `calculate_real_ip_from_rack`.

### Actual Behavior

When @send_default_pii is true, the ip_address set via Sentry.set_user is overwritten by the IP address calculated from `calculate_real_ip_from_rack`. This results in Cloudflare’s IP address being recorded in Sentry instead of the actual client’s IP address. This discrepancy may arises because the cloudflare-rails gem modifies the IP addresses to Cloudflare’s proxies without change trusted proxies.

### Ruby Version

3.2.2

### SDK Version

5.18.1

_No response_

### Sentry Config
```ruby
Sentry.init do |config|
  config.dsn = ''
  config.breadcrumbs_logger = %i[active_support_logger http_logger]
end
```

sl0thentr0py:
Fixes #2347

### DIFF
--- a/sentry-ruby/spec/sentry/event_spec.rb
+++ b/sentry-ruby/spec/sentry/event_spec.rb
@@ -141,6 +141,12 @@ RSpec.describe Sentry::Event do
         expect(event.to_hash[:user][:ip_address]).to eq("2.2.2.2")
       end
 
+      it "doesn't overwrite already set ip address" do
+        Sentry.set_user({ ip_address: "3.3.3.3" })
+        Sentry.get_current_scope.apply_to_event(event)
+        expect(event.to_hash[:user][:ip_address]).to eq("3.3.3.3")
+      end
+
       context "with config.trusted_proxies = [\"2.2.2.2\"]" do
         before do
           Sentry.configuration.trusted_proxies = ["2.2.2.2"]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## Unreleased
 
+### Bug Fixes
+
+- Don't overwrite `ip_address` if already set on `user` [#2350](https://github.com/getsentry/sentry-ruby/pull/2350)
+  - Fixes [#2347](https://github.com/getsentry/sentry-ruby/issues/2347)
+
 ### Internal
 
 - Use Concurrent.usable_processor_count when it is available ([#2339](https://github.com/getsentry/sentry-ruby/pull/2339))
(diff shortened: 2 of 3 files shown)


solnic (approved):
LGTM!

sentry-ruby/spec/sentry/event_spec.rb
```diff
@@ -141,6 +141,12 @@
         expect(event.to_hash[:user][:ip_address]).to eq("2.2.2.2")
       end
 
+      it "doesn't overwrite already set ip address" do
+        Sentry.set_user({ ip_address: "3.3.3.3" })
+        Sentry.get_current_scope.apply_to_event(event)
+        expect(event.to_hash[:user][:ip_address]).to eq("3.3.3.3")
```
solnic, line 147: @sl0thentr0py note for the future - this should be `event.to_h`, because `to_hash` is a special coercion method that's used implicitly in some places in Ruby and may cause weird issues. For example `**` uses it, see:

```ruby
(ruby) {foo: "bar", **event}
{:foo=>"bar",
 :event_id=>"266e3389d9b64f57b48da6f3d0969a93",
 :level=>:error,
 :timestamp=>"2024-07-22T16:22:13Z",
 :environment=>"development",
 :server_name=>"44b9b235e7c3",
 :modules=>
  {"rake"=>"12.3.3",
 # ...
```

It should only be used when the object that implements it is actually a hash-like object, and events are not like that.

st0012: Thanks! For context, I took it from the [old SDK](https://github.com/getsentry/sentry-ruby/blob/36866c54ea67933c36a5d8ce2b52dae972e3edd1/sentry-raven/lib/raven/event.rb#L135-L148) when implementing `sentry-ruby`. So we probably can make the switch rather freely. I'll put up a PR to see how it goes.

st0012 (approved):

PR #2351, opened 2024-07-23
Migrate from to_hash to to_h

st0012:
As @solnic pointed out in https://github.com/getsentry/sentry-ruby/pull/2350#discussion_r1686820195 `to_hash` has special meaning in Ruby and could be called implicitly in contexts like double splatting argument. So we should switch to `to_h` to avoid potential issues.

Additionally, Matz also [explained](https://bugs.ruby-lang.org/issues/10180#note-1) that in the explicit conversion usages (which is the case for the SDK), `to_h` should be used.

### Compatibility Concern

Since this PR changes the serialization interface of almost all SDK's internal components, if users/libs relied on patching `to_hash` to extend/modify the data, this could become a breaking change for them. And because `to_hash` has been that interface since the old `sentry-raven` SDK (for at least [12 years](https://github.com/getsentry/sentry-ruby/commit/5c045f80b49bd52831f3f37af5c121749a4e4bb1)), I think we can assume at least some legacy apps had built customization on top of it.

For this reason, maybe this should be shipped in 6.0?

(In cases where `Configuration#async` is set, we use `Event#to_json_compatible` instead, so they won't be affected)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Breaking Changes
 
 - Remove `config.async` [#1894](https://github.com/getsentry/sentry-ruby/pull/1894)
+- Migrate from to_hash to to_h ([#2351](https://github.com/getsentry/sentry-ruby/pull/2351))
 
 ## Unreleased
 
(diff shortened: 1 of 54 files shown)


sl0thentr0py:
@st0012 I created a 6.0-dev branch 
https://github.com/getsentry/sentry-ruby/pull/2352

can you base this rebase this branch on that branch and change the base there?

solnic:
Thanks for tackling this! I did not expect you'd address it so quickly ❤️ This is definitely a potentially breaking change, but luckily it's easily fixable in people's codebases.

st0012:
@sl0thentr0py rebased 👍

sl0thentr0py (approved):

PR #2357, opened 2024-07-29
File/attachment uploads

### LINKED ISSUE #1548: File/attachment uploads
Practically all the other SDKs support file/attachment uploads.

We are new to Sentry and are looking to capture screenshots for errors.

Is this on the roadmap? Is there a short term workaround or solution?

Thanks!

solnic:
This adds support for attaching files to the scope like this:

```ruby
# Using path option
Sentry.add_attachment(path: "/your/attachment.txt")

# Using filename and bytes
Sentry.add_attachment(filename: "attachment.txt", bytes: File.read("/your/attachment.txt"))
```

Screenshot 2024-08-06 at 12 54 36

### TODO

- [x] Add `Sentry.add_attachment(**opts)`
- [x] Extend Transport and Envelope with support for sending attachments
- [x] Add support for `:path` option
- [x] Add support for figuring out `content_type` automatically based on `filename`
- [x] Ensure that either bytes or file content as a string are supported and that multi-line content is working

---

Closes #1548

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@
   - Closes [#1795](https://github.com/getsentry/sentry-ruby/issues/1795)
   - Please note that the Faraday instrumentation has some limitations in case of async requests: https://github.com/lostisland/faraday/issues/1381
 - Support for attachments ([#2357](https://github.com/getsentry/sentry-ruby/pull/2357))
+  Usage:
+
+  ```ruby
+  Sentry.add_attachment(path: '/foo/bar.txt')
+  Sentry.add_attachment(filename: 'payload.json', bytes: '{"value": 42}'))
+  ```
 
 ## 5.18.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Support for tracing Faraday requests ([#2345](https://github.com/getsentry/sentry-ruby/pull/2345))
   - Closes [#1795](https://github.com/getsentry/sentry-ruby/issues/1795)
   - Please note that the Faraday instrumentation has some limitations in case of async requests: https://github.com/lostisland/faraday/issues/1381
+- Support for attachments ([#2357](https://github.com/getsentry/sentry-ruby/pull/2357))
 
 ## 5.18.2
 
--- a/sentry-ruby/lib/sentry/attachment.rb
+++ b/sentry-ruby/lib/sentry/attachment.rb
@@ -14,7 +14,7 @@ module Sentry
     end
 
     def to_envelope_headers
-      { type: 'attachment', filename: filename }
+      { type: 'attachment', filename: filename, length: payload.bytesize }
     end
 
     def payload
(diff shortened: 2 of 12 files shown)


sentry-ruby/sentry-ruby.gemspec
```diff
@@ -26,6 +26,7 @@ Gem::Specification.new do |spec|
 
   spec.require_paths = ["lib"]
 
-  spec.add_dependency "concurrent-ruby", '~> 1.0', '>= 1.0.2'
+  spec.add_dependency "concurrent-ruby", "~> 1.0", ">= 1.0.2"
+  spec.add_dependency "mime-types", "~> 3.0"
```
sl0thentr0py, line 30: ugh sad that this isn't in stdlib, we are actually very strict about introducing deps. 
Since we are sending the `filename` anyway, relay could in theory do the MIME extraction server side, so I would leave this out of the client and remove the `infer_content_type` above. It's only a nice to have and I would rather not add the new dependency.

sl0thentr0py (changes requested):

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Support for tracing Faraday requests ([#2345](https://github.com/getsentry/sentry-ruby/pull/2345))
   - Closes [#1795](https://github.com/getsentry/sentry-ruby/issues/1795)
   - Please note that the Faraday instrumentation has some limitations in case of async requests: https://github.com/lostisland/faraday/issues/1381
+- Support for attachments ([#2357](https://github.com/getsentry/sentry-ruby/pull/2357))
```
sl0thentr0py, line 8: for new features, we add a small snippet below the changelog entry that shows users how to use the new feature. So here,

```ruby
Sentry.add_attachment(path: '/foo/bar.txt')
Sentry.add_attachment(filename: 'payload.json', bytes: '{"value": 42}'))
```

sentry-ruby/lib/sentry/attachment.rb
```diff
@@ -0,0 +1,43 @@
...
+    PathNotFoundError = Class.new(StandardError)
+
+    attr_reader :bytes, :filename, :path, :content_type, :add_to_transactions
+
+    def initialize(bytes: nil, filename: nil, content_type: nil, path: nil, add_to_transactions: false)
+      @bytes = bytes
+      @filename = infer_filename(filename, path)
+      @path = path
+      @content_type = content_type
+      @add_to_transactions = add_to_transactions
```
sl0thentr0py, line 14: we're still not using `add_to_transactions` in the logic below anywhere.
Now that I think about it, I don't want to have this option in ruby and we can always send the attachment, so I would remove this  and remove the `Event` check in `transport.rb#148`

sl0thentr0py (changes requested):
couple of comments

sentry-ruby/lib/sentry/attachment.rb
solnic: @sl0thentr0py if I remove the Event check, spec breaks because sometimes the event is a plain Hash, what should I do with this?

sl0thentr0py: ahh it's Event, sorry yes keep it as is then

sl0thentr0py (approved):

PR #2360, opened 2024-07-31
Unify Net::HTTP and Faraday implementations

solnic:
Bonus refa after #2345

⚠️ this also fixes a bug that I found in Faraday instrumenter where it would early return `nil` if Sentry is not initialized instead of issuing a request. See 259f223965ebd7230415e15ad27e27ad833ec898

#skip-changelog

### DIFF
--- a/sentry-ruby/lib/sentry/utils/http_tracing.rb
+++ b/sentry-ruby/lib/sentry/utils/http_tracing.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+module Sentry
+  module Utils
+    module HttpTracing
+      def set_span_info(sentry_span, request_info, response_status)
+        sentry_span.set_description("#{request_info[:method]} #{request_info[:url]}")
+        sentry_span.set_data(Span::DataConventions::URL, request_info[:url])
+        sentry_span.set_data(Span::DataConventions::HTTP_METHOD, request_info[:method])
+        sentry_span.set_data(Span::DataConventions::HTTP_QUERY, request_info[:query]) if request_info[:query]
+        sentry_span.set_data(Span::DataConventions::HTTP_STATUS_CODE, response_status)
+      end
+
+      def set_propagation_headers(req)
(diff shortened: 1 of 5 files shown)


sentry-ruby/lib/sentry/utils/http_tracing.rb
```diff
@@ -0,0 +1,40 @@
...
+        Sentry.add_breadcrumb(crumb)
+      end
+
+      def record_sentry_breadcrumb?
+        Sentry.initialized? && Sentry.configuration.breadcrumbs_logger.include?(:http_logger)
+      end
+
+      def propagate_trace?(url)
+        url &&
+          Sentry.configuration.propagate_traces &&
```
sl0thentr0py, line 35: there's an annoying threaded race condition sometimes with `Sentry.close` where `Sentry.configuration` can be `nil` so if you access `Sentry.configuration` directly like this you need an additional `Sentry.initialized?` check here too like `record_sentry_breadcrumb?` above.

sl0thentr0py (changes requested):
just one comment

sentry-ruby/lib/sentry/utils/http_tracing.rb
```diff
@@ -0,0 +1,41 @@
...
+      end
+
+      def record_sentry_breadcrumb?
+        Sentry.initialized? && Sentry.configuration.breadcrumbs_logger.include?(:http_logger)
+      end
+
+      def propagate_trace?(url)
+        url &&
+          Sentry.initialized? &&
+          Sentry.configuration.propagate_traces &&
```
solnic, line 36: @sl0thentr0py given that direct access to the configuration is not thread-safe, we could have a method for checking configuration values that would perform this additonal `initialize?` check, something like:

```
Sentry.configuration(:propagate_traces)
```

sl0thentr0py: feel free to to add a helper but the semantics of that (truthy or falsey?) are not always clean.

sl0thentr0py (approved):
ty!

PR #2369, opened 2024-08-13
Include Sentry meta tag in the layout automatically

### LINKED ISSUE #2368: Auto inject meta (and JS SDK) into rails root layout as optional part of generator

solnic:
Closes #2368

### DIFF
--- a/sentry-rails/lib/generators/sentry_generator.rb
+++ b/sentry-rails/lib/generators/sentry_generator.rb
@@ -16,4 +16,10 @@ class SentryGenerator < ::Rails::Generators::Base
       end
     RUBY
   end
+
+  def inject_code_into_layout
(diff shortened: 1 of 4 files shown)


sentry-rails/lib/generators/sentry_generator.rb
```diff
@@ -16,4 +16,10 @@ def copy_initializer_file
       end
     RUBY
   end
+
+  def inject_code_into_layout
+    inject_into_file "app/views/layouts/application.html.erb", before: "</head>\n" do
```
sl0thentr0py, line 21: Can we gate this behind a default on boolean option like `--inject-meta`?

sl0thentr0py (changes requested):

sl0thentr0py (approved):

PR #2372, opened 2024-08-14
Experimental support for multi-threaded profiling using Vernier

solnic:
Experimental support for Vernier.

### Screenshots

Screenshot 2024-08-21 at 13 41 53

Screenshot 2024-08-21 at 13 37 24

Screenshot 2024-08-21 at 13 38 07

Screenshot 2024-08-21 at 13 39 04

Screenshot 2024-08-21 at 13 39 34


sentry-ruby/lib/sentry/vernier/output.rb
```diff
@@ -0,0 +1,97 @@
...
+            in_app: in_app
+          }.compact
+        }
+      end
+
+      def stacks
+        sample_stacks = profile.threads.flat_map { |_, thread_info|
+          samples = thread_info[:samples].map { |stack_id| profile.stack(stack_id) }
+
+          samples.map { |stack| [stack.idx, stack.frames.map(&:idx)] }
```
sl0thentr0py, line 82: just to confirm, vernier will assign unique `idx`s to stacks across threads right?

sentry-ruby/lib/sentry/vernier/output.rb
```diff
@@ -0,0 +1,97 @@
...
+        }.to_h
+
+        stacks = profile._stack_table.stack_count.times.map do |stack_id|
```
sl0thentr0py, line 85: I don't understand this step, are there some stacks that are not in `profile.threads` but are in `profile._stack_table`? Maybe add a comment clarifying the structure.

sl0thentr0py (changes requested):
amazing work @solnic 
did a first pass, some questions and comments

will do a detailed test round later in the week

sentry-ruby/lib/sentry/profiler/helpers.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+require "securerandom"
+
+module Sentry
+  class Profiler
+    module Helpers
+      def in_app?(abs_path)
+        abs_path.match?(@in_app_pattern)
```
st0012, line 9: Since this helper module relies on multiple instance variables, I feel in this case it's better to create a base profiler class that has these helpers but also define the ivars in the constructor.

sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -393,6 +397,8 @@ def initialize
       self.traces_sampler = nil
       self.enable_tracing = nil
 
+      self.profiler = Sentry::Profiler
```
st0012, line 400: Can we name this `profiler_class` instead?

sentry-ruby/lib/sentry/vernier/output.rb
```diff
@@ -0,0 +1,97 @@
...
+            }
+          }.compact
+        }
+      end
+
+      def frames
+        funcs = stack_table_hash[:frame_table].fetch(:func)
+        lines = stack_table_hash[:func_table].fetch(:first_line)
+
+        funcs.map { |idx|
```
st0012, line 60: nit: let's use do/end block for multi-line blocks if possible? I can also enable the rubocop rule for it to save some manual work.

sentry-ruby/lib/sentry/vernier/output.rb
st0012, line 82: I think we don't need to do 2 loops here:

```suggestion
          thread_info[:samples].map do |stack_id|
            stack = profile.stack(stack_id)
            [stack.idx, stack.frames.map(&:idx)]
          end
```

sentry-ruby/lib/sentry/vernier/profiler.rb
```diff
@@ -0,0 +1,126 @@
...
+
+      attr_reader :started, :event_id, :result
+
+      def initialize(configuration)
+        @event_id = SecureRandom.uuid.delete("-")
+
+        @started = false
+        @sampled = nil
+
+        @profiling_enabled = defined?(Vernier) && configuration.profiling_enabled?
```
st0012, line 25: Is it possible for this part being loaded but `Vernier` is not being defined?

sentry-ruby/lib/sentry/vernier/profiler.rb
```diff
@@ -0,0 +1,126 @@
...
+        @profiles_sample_rate = configuration.profiles_sample_rate
+        @project_root = configuration.project_root
+        @app_dirs_pattern = configuration.app_dirs_pattern || Backtrace::APP_DIRS_PATTERN
```
st0012, line 28: We can remove the fallback once #2390 is merged.

sentry-ruby/lib/sentry/vernier/output.rb
sl0thentr0py: yes I was also thinking the same but maybe there's a reason I couldn't understand

sentry-ruby/lib/sentry/configuration.rb
solnic: Good point! Renamed.

sentry-ruby/lib/sentry/profiler/helpers.rb
solnic: @st0012 yeah I was planning to make a bigger refactoring of these classes in a separate PR but I can introduce an abstract class already here. Definitely better than a Helper module :)

sentry-ruby/lib/sentry/vernier/output.rb
solnic: ah ok, I am used to using {..} for blocks that return values, regardless if they are single line or not.

sentry-ruby/lib/sentry/profiler/helpers.rb
solnic: I ended up postponing this because Vernier uses its Output class to generate the profile hash, and that one uses Helpers. What I'd like to do eventually is to refactor both profiler classes so that use the same structure (like using Output class for example) and then unifying them will become simpler.

sentry-ruby/lib/sentry/vernier/output.rb
solnic: Oh I had it like that for debugging purposes :) It's a single loop now.

sentry-ruby/lib/sentry/vernier/output.rb
solnic: @sl0thentr0py that's what I'm seeing yes:

```
/usr/local/rvm/gems/default/bin/rspec:80
/usr/local/rvm/gems/default/bin/rspec:83
thread-bar-0:88
thread-bar-0:87
thread-bar-0:88
thread-bar-0:87
thread-bar-0:89
thread-bar-0:88
thread-bar-0:87
thread-bar-0:88
thread-bar-0:90
thread-bar-0:87
thread-bar-0:91
thread-bar-0:88
thread-bar-0:87
thread-bar-0:88
thread-bar-0:90
thread-bar-0:87
thread-bar-0:88
thread-bar-0:91
thread-bar-0:87
thread-bar-0:88
thread-bar-0:89
thread-bar-0:88
thread-bar-0:87
thread-bar-0:88
thread-bar-0:90
thread-bar-0:88
thread-bar-0:89
thread-bar-0:88
thread-bar-0:87
thread-bar-0:88
thread-bar-0:87
thread-bar-0:88
thread-bar-0:89
thread-bar-0:88
thread-bar-0:87
thread-bar-0:88
thread-bar-0:93
thread-bar-1:80
thread-bar-1:89
thread-bar-1:90
thread-bar-1:91
thread-bar-1:88
thread-bar-1:87
thread-bar-1:88
thread-bar-1:87
thread-bar-1:91
thread-bar-1:88
thread-bar-1:89
thread-bar-1:88
thread-bar-1:89
thread-bar-1:88
thread-bar-1:89
thread-bar-1:87
thread-bar-1:88
thread-bar-1:91
thread-bar-1:88
thread-bar-1:87
thread-bar-1:88
thread-bar-1:90
thread-bar-1:87
thread-bar-1:90
thread-bar-1:88
thread-bar-1:90
thread-bar-1:88
thread-bar-1:91
thread-bar-1:87
thread-bar-1:89
thread-bar-1:88
thread-bar-1:89
thread-bar-1:91
thread-bar-1:88
thread-bar-1:87
thread-bar-1:88
thread-bar-1:91
thread-bar-1:87
thread-bar-1:93
Did not find stack 0 in sample stacks
Did not find stack 1 in sample stacks
Did not find stack 2 in sample stacks
Did not find stack 3 in sample stacks
Did not find stack 4 in sample stacks
Did not find stack 5 in sample stacks
Did not find stack 6 in sample stacks
Did not find stack 7 in sample stacks
Did not find stack 8 in sample stacks
Did not find stack 9 in sample stacks
Did not find stack 10 in sample stacks
Did not find stack 11 in sample stacks
Did not find stack 12 in sample stacks
Did not find stack 13 in sample stacks
Did not find stack 14 in sample stacks
Did not find stack 15 in sample stacks
Did not find stack 16 in sample stacks
Did not find stack 17 in sample stacks
Did not find stack 18 in sample stacks
Did not find stack 19 in sample stacks
Did not find stack 20 in sample stacks
Did not find stack 21 in sample stacks
Did not find stack 22 in sample stacks
Did not find stack 23 in sample stacks
Did not find stack 24 in sample stacks
Did not find stack 25 in sample stacks
Did not find stack 26 in sample stacks
Did not find stack 27 in sample stacks
Did not find stack 28 in sample stacks
Did not find stack 29 in sample stacks
Did not find stack 30 in sample stacks
Did not find stack 31 in sample stacks
Did not find stack 32 in sample stacks
Did not find stack 33 in sample stacks
Did not find stack 34 in sample stacks
Did not find stack 35 in sample stacks
Did not find stack 36 in sample stacks
Did not find stack 37 in sample stacks
Did not find stack 38 in sample stacks
Did not find stack 39 in sample stacks
Did not find stack 40 in sample stacks
Did not find stack 41 in sample stacks
Did not find stack 42 in sample stacks
Did not find stack 43 in sample stacks
Did not find stack 44 in sample stacks
Did not find stack 45 in sample stacks
Did not find stack 46 in sample stacks
Did not find stack 47 in sample stacks
Did not find stack 48 in sample stacks
Did not find stack 49 in sample stacks
Did not find stack 50 in sample stacks
Did not find stack 51 in sample stacks
Did not find stack 52 in sample stacks
Did not find stack 53 in sample stacks
Did not find stack 54 in sample stacks
Did not find stack 55 in sample stacks
Did not find stack 56 in sample stacks
Did not find stack 57 in sample stacks
Did not find stack 58 in sample stacks
Did not find stack 59 in sample stacks
Did not find stack 60 in sample stacks
Did not find stack 61 in sample stacks
Did not find stack 62 in sample stacks
Did not find stack 63 in sample stacks
Did not find stack 64 in sample stacks
Did not find stack 65 in sample stacks
Did not find stack 66 in sample stacks
Did not find stack 67 in sample stacks
Did not find stack 68 in sample stacks
Did not find stack 69 in sample stacks
Did not find stack 70 in sample stacks
Did not find stack 71 in sample stacks
Did not find stack 72 in sample stacks
Did not find stack 73 in sample stacks
Did not find stack 74 in sample stacks
Did not find stack 75 in sample stacks
Did not find stack 76 in sample stacks
Did not find stack 77 in sample stacks
Did not find stack 78 in sample stacks
Did not find stack 79 in sample stacks
Found stack 80 in sample stacks
Did not find stack 81 in sample stacks
Did not find stack 82 in sample stacks
Found stack 83 in sample stacks
Did not find stack 84 in sample stacks
Did not find stack 85 in sample stacks
Did not find stack 86 in sample stacks
Found stack 87 in sample stacks
Found stack 88 in sample stacks
Found stack 89 in sample stacks
Found stack 90 in sample stacks
Found stack 91 in sample stacks
Did not find stack 92 in sample stacks
Found stack 93 in sample stacks
```

sentry-ruby/lib/sentry/vernier/output.rb
solnic: @sl0thentr0py gaah, good catch - seems like it's not unique per thread. I need to rework this a bit and fix!

solnic:
@sl0thentr0py @st0012 thanks for the comments! There are a couple of things that need be fixed. I'll ping you once it's done next week.

sentry-ruby/lib/sentry/vernier/profiler.rb
solnic: @st0012 yes, we rescue from LoadError if vernier is not bundled. It's the same with stackprof, its Profiler gets required even if stackprof is not available (see transaction.rb). I think we can make this whole conditional code loading better in the future and limit the amount of files that sentry requires. ie we can have some API for configuring features and then based on its configuration a smart code loader would know what to do.

sentry-ruby/lib/sentry/vernier/output.rb
solnic: @sl0thentr0py this is fixed via 3a9367b03b65f0f575d87a3e8ee5ddf9b2b5529b

sentry-ruby/lib/sentry/vernier/profiler.rb
st0012: I think we can `return` in the rescue block then we won't need to call `defined?` at all. For example:

```rb
begin
  require "foo"
rescue LoadError
  return
end

Foo.bar
```

st0012 (approved):
Left another [comment](https://github.com/getsentry/sentry-ruby/pull/2372#discussion_r1741267077) wrt the `defined?` check. But otherwise looks good to me 👍

sl0thentr0py:
@solnic we're not requiring the new `sentry/vernier/profiler` anywhere? We should not force end users to require another obscure path from within the gem, they should all be setup internally.

sentry-ruby/lib/sentry/vernier/profiler.rb
sl0thentr0py: sorry to come back to this, but can we keep both similar. Right now stackprof has one way and vernier has another.

sl0thentr0py:
I tried on a simple rails controller with some database calls and the profile is not sent due to being oversized. 
However, the [size limits](https://develop.sentry.dev/sdk/envelopes/#size-limits) for a profile payload are much larger (50MB) than for errors/transactions (1MB), so we need to add this logic in `Envelope::Item`.

solnic:
> @solnic we're not requiring the new `sentry/vernier/profiler` anywhere? We should not force end users to require another obscure path from within the gem, they should all be setup internally.
> 
> I would also prefer a warning message in `Configuration` when people try to set the profiler to vernier but vernier is not installed as a peer dependency gem.

@sl0thentr0py OK I made it so that sentry/vernier/profiler is required automatically and if you try to set it in the config when vernier is not available, the you'll get a meaningful error message:

```
3.1.6 :002 > Sentry::Configuration.new.tap { |c| c.profiler_class = Sentry::Vernier::Profiler }
/workspaces/sentry/sentry-ruby/sentry-ruby/lib/sentry/configuration.rb:506:in `rescue in profiler_class=': Please add the 'vernier' gem to your Gemfile to use the Vernier profiler with Sentry. (ArgumentError)
        from /workspaces/sentry/sentry-ruby/sentry-ruby/lib/sentry/configuration.rb:503:in `profiler_class='
        from (irb):2:in `block in <top (required)>'
        from <internal:kernel>:90:in `tap'
        from (irb):2:in `<main>'
```

solnic:
> I tried on a simple rails controller with some database calls and the profile is not sent due to being oversized.
> However, the [size limits](https://develop.sentry.dev/sdk/envelopes/#size-limits) for a profile payload are much larger (50MB) than for errors/transactions (1MB), so we need to add this logic in `Envelope::Item`.

@sl0thentr0py wdym by "this logic" exactly?

sl0thentr0py:
we need to generalize [this logic to have different limits](https://github.com/getsentry/sentry-ruby/blob/04e030f974558aa750e26afc458c48e663ddcfcc/sentry-ruby/lib/sentry/envelope.rb#L42-L56) according to different envelope `item.type` but let's make a separate issue/PR for that.

solnic:
@sl0thentr0py alright this is now using the increased envelope profile item limit

sl0thentr0py:
omg the profiles look perfect <3

sl0thentr0py (approved):
incredible work!!! so happy to ship this

solnic:
@sl0thentr0py yessssss! Thanks for the reviews and testing! Let's 🚢 it 😄

PR #2374, opened 2024-08-15
feat: Add support for $SENTRY_DEBUG and $SENTRY_SPOTLIGHT

BYK:
Part of getsentry/spotlight#482. Similar to sentry-python#3443 but it also adds support for `$SENTRY_DEBUG` which was lacking in the Ruby SDK.

### DIFF
--- a/sentry-ruby/lib/sentry/utils/env_helper.rb
+++ b/sentry-ruby/lib/sentry/utils/env_helper.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+module Sentry
+  module EnvHelper
+    TRUTHY_ENV_VALUES = %w(t true yes y 1 on).freeze
+    FALSY_ENV_VALUES = %w(f false no n 0 off).freeze
+
+    def env_to_bool(value, strict=false)
+      value = value.to_s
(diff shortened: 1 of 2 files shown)


BYK:
Astonished that it only took me like 5 tries to get it right 🤣

sentry-ruby/lib/sentry/utils/env_helper.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Sentry
+  module Utils
+    module EnvHelper
+      TRUTHY_ENV_VALUES = %w[t true yes y 1 on].freeze
+      FALSY_ENV_VALUES = %w[f false no n 0 off].freeze
+
+      def EnvHelper.env_to_bool(value, strict: false)
```
sl0thentr0py, line 9:
```suggestion
      def self.env_to_bool(value, strict: false)
```

sl0thentr0py (changes requested):
I would like a couple of tests in `spotlight_transport_spec.rb` but if you don't wanna figure out rspec I can also take care of that, lmk

sl0thentr0py (approved):

PR #2377, opened 2024-08-16
Fix rubocop setup (for some envs)

solnic:
This makes it possible to use rubocop when working from inside a subdir.
It also fixes running rubocop from the root where it would actually use
the config that we have. Previously it would ignore it for some reason,
symlinking did the trick.

The rubocop-packaging was needed because something somewhere tries to
require it.

#skip-changelog


st0012 (approved):
🙌

PR #2382, opened 2024-08-17
Support Rails 7.2 on CI

st0012:
#skip-changelog

On my machine, running `sentry-rails`'s full test suite with `rake test` against Rails 7.1 takes less than 8 seconds. But it now takes ~33 seconds against Rails 7.2 even after I solved the slowdown issue caused by the eager load namespace accumulation issue. That said, I think we can start running tests against Rails 7.2 and optimize the test setup further in another PR.

### DIFF
--- a/.github/workflows/sentry_rails_test.yml
+++ b/.github/workflows/sentry_rails_test.yml
@@ -43,6 +43,9 @@ jobs:
           - { ruby_version: "2.7", rails_version: 5.2.0 }
           - { ruby_version: "2.7", rails_version: 6.0.0 }
           - { ruby_version: "2.7", rails_version: 6.1.0 }
+          - { ruby_version: "3.1", rails_version: 7.2.0 }
+          - { ruby_version: "3.2", rails_version: 7.2.0 }
+          - { ruby_version: "3.3", rails_version: 7.2.0 }
           - { ruby_version: "jruby", rails_version: 6.1.0 }
           - {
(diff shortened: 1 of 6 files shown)


.github/workflows/sentry_rails_test.yml
```diff
@@ -43,6 +43,9 @@ jobs:
           - { ruby_version: "2.7", rails_version: 5.2.0 }
           - { ruby_version: "2.7", rails_version: 6.0.0 }
           - { ruby_version: "2.7", rails_version: 6.1.0 }
+          - { ruby_version: "3.1", rails_version: 7.2.0 }
+          - { ruby_version: "3.2", rails_version: 7.2.0 }
+          - { ruby_version: "3.3", rails_version: 7.2.0 }
```
st0012, line 48: I know it's ultimately a business decision, but IMO we should drop support or at least stop running CI against Ruby 2.4, 2.5 and Rails 5.0 and 5.1.
Otherwise, with the release of Rails 8.0 in the near future, we'll be running so many jobs on CI for each PR.

sl0thentr0py (approved):

.github/workflows/sentry_rails_test.yml
sl0thentr0py: I will prepare a major soon and remove, just need some time since I'm busy with another larger project

PR #2387, opened 2024-08-22
Improve sidekiq-cron patch

adamzapasnik:
## Description
Describe your changes:

I've added two changes:

1st `name.to_s` cause one may provide hash with symbols to `load_from_hash` and `name` ends up being a symbol, and then it ends up raising an error in `sentry-ruby/lib/sentry/hub.rb` cause it expects a string...
```ruby
    def capture_check_in(slug, status, **options)
      check_argument_type!(slug, ::String)
```

2nd change is much more important though, it's required to support natural-language-format, because the current version doesn't create a monitor successfully.  natural language format is something like `every 5 minutes` which gets translated to `'*/5 * * * *'` as you can see it in the tests

https://github.com/sidekiq-cron/sidekiq-cron?tab=readme-ov-file#natural-language-formats

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 ## Unreleased
 
+### Features
+
 - Add support for $SENTRY_DEBUG and $SENTRY_SPOTLIGHT ([#2374](https://github.com/getsentry/sentry-ruby/pull/2374))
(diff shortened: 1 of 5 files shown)


sl0thentr0py (approved):
ty @adamzapasnik!

PR #2388, opened 2024-08-26
Improve CI

solnic:
- Exlude rubocop from CI's bundle - this was needed because we want to test things when rack is not installed too and unfortunately rubocop-rails depends on rack, so rack was always installed
- Measure coverage on latest Ruby 3.3 - we're gonna have specs that run only under 3.3 once Vernier branch is merged and w/o this tweak coverage would go down
- Add rack 3.1 and redis 5.3 to the test matrix

#skip-changelog


solnic:
The diff in cov against master is most likely caused by the fact that this PR runs coverage under 3.3 and master 3.2, so we can just ignore it

sl0thentr0py (approved):

PR #2394, opened 2024-09-02
Verifies that client is available before sending breadcrumbs

lomefin:
## Description

During build, we have encountered situations where Sentry wants to send
a `breadcrumb` but it fails to accomplish the task because of
`NoMethodError: undefined method `configuration' for nil:NilClass`.
After one of this error is triggered, Sentry fails to work at all,
passing from 0 to several hundred errors in our pipeline, and then back
to zero once we start rebuilding.

Our pipeline tests using parallelism and there is a ton of concurrency,
so we believe that has something to do with it.

The backtrace points out to `Hub#add_breadcrumb` and inside of it to
`configuration`. It is possible that while the new "local" hub is
available, the client is not and fails to work.

We have observed that several other method have checks
in place to verify that the client is really available but not on
`add_breadcrumb` so we added a verification before starting the rest
of the process.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Add support for $SENTRY_DEBUG and $SENTRY_SPOTLIGHT ([#2374](https://github.com/getsentry/sentry-ruby/pull/2374))
 - Support human readable intervals in `sidekiq-cron` ([#2387](https://github.com/getsentry/sentry-ruby/pull/2387))
 - Set default app dirs pattern ([#2390](https://github.com/getsentry/sentry-ruby/pull/2390))
+- Verifies presence of client before adding a breadcrumb ([#2394](https://github.com/getsentry/sentry-ruby/pull/2394))
 
 ## 5.19.0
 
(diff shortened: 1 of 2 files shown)


st0012:
> It is possible that while the new "local" hub is
available, the client is not and fails to work.

Can you elaborate more on this? I think this indicates a bigger problem either in your application's Sentry setup OR in the SDK itself.

I'd like to check a few things like:
- Does the app itself create and manage `Hub`s? For example, by calling `Hub#new_from_top` or `Hub#clone`
- Does any part of it call `Hub#pop_scope`?
    - @sl0thentr0py @solnic Although it may not be the cause of this issue, I feel like [pop_scope](https://github.com/getsentry/sentry-ruby/blob/c3bcfa0401856d9ee7616c95ab9dfe988a89da30/sentry-ruby/lib/sentry/hub.rb#L75-L77) is a very dangerous API to expose as it pops client as well. I think when there's only one layer left, `pop_scope` should retain the layer but just replace the scope?

sl0thentr0py:
@st0012 this is just part of the design problem of the older hub/scopes, we will eventually consolidate into a single OTEL-like context eventually so there will be no more `pop_scope`.

Issue #2378, opened 2024-08-16
NoMethodError: undefined method `add_session` for nil


MrSerth:
### Issue Description

In my server logs (but not in Sentry), I found another occurrence of #2268: `NoMethodError: undefined method 'add_session for nil`

### Reproduction Steps

Unfortunately, I don't have clear reproduction steps. I know that the error occurred during [this trace](https://codeocean.sentry.io/performance/trace/83013e024cd640d1b58ecd8bd77e9d12) and doesn't occur regularly.

### Expected Behavior

No error is generated in Sentry, and `add_session` can be called successfully.

### Actual Behavior

A `NoMethodError` is raised.

![Image](https://github.com/user-attachments/assets/de7571e2-5aa8-4a76-848f-11bce952711d)

### Ruby Version

3.3.4

### SDK Version

5.18.2

### Integration and Its Version

Rails 7.1.3.4

PR #2396, opened 2024-09-02
fix: Safe-navigate to session flusher

ixti:
`Sentry.session_flusher` is `nil`-able. And under some circumstances (which I still can't figure out) in specs it's `nil` causing our spec suite randomly (but as of recent, pretty all the time) fail.

### Example failure

```
     Failure/Error: post(path, env:, headers:, xhr: true)
     
     NoMethodError:
       undefined method `add_session' for nil
     # /home/ixti/.gem/ruby/3.3.4/gems/sentry-ruby-5.19.0/lib/sentry/hub.rb:251:in `end_session'
     # /home/ixti/.gem/ruby/3.3.4/gems/sentry-ruby-5.19.0/lib/sentry/hub.rb:260:in `with_session_tracking'
     # /home/ixti/.gem/ruby/3.3.4/gems/sentry-ruby-5.19.0/lib/sentry-ruby.rb:411:in `with_session_tracking'
     # /home/ixti/.gem/ruby/3.3.4/gems/sentry-ruby-5.19.0/lib/sentry/rack/capture_exceptions.rb:21:in `block in call'
     # /home/ixti/.gem/ruby/3.3.4/gems/sentry-ruby-5.19.0/lib/sentry/hub.rb:59:in `with_scope'
     # /home/ixti/.gem/ruby/3.3.4/gems/sentry-ruby-5.19.0/lib/sentry-ruby.rb:391:in `with_scope'
     # /home/ixti/.gem/ruby/3.3.4/gems/sentry-ruby-5.19.0/lib/sentry/rack/capture_exceptions.rb:20:in `call'
     # /home/ixti/.gem/ruby/3.3.4/gems/rails_semantic_logger-4.17.0/lib/rails_semantic_logger/rack/logger.rb:45:in `call_app'
     # /home/ixti/.gem/ruby/3.3.4/gems/rails_semantic_logger-4.17.0/lib/rails_semantic_logger/rack/logger.rb:26:in `block in call'
     # /home/ixti/.gem/ruby/3.3.4/bundler/gems/semantic_logger-093c0258b053/lib/semantic_logger/base.rb:190:in `block in tagged'
     # /home/ixti/.gem/ruby/3.3.4/bundler/gems/semantic_logger-093c0258b053/lib/semantic_logger/semantic_logger.rb:395:in `named_tagged'
     # /home/ixti/.gem/ruby/3.3.4/bundler/gems/semantic_logger-093c0258b053/lib/semantic_logger/base.rb:197:in `tagged'
     # /home/ixti/.gem/ruby/3.3.4/gems/rails_semantic_logger-4.17.0/lib/rails_semantic_logger/rack/logger.rb:26:in `call'
     # /home/ixti/.gem/ruby/3.3.4/gems/rack-3.1.7/lib/rack/runtime.rb:24:in `call'
     # /home/ixti/.gem/ruby/3.3.4/gems/rack-3.1.7/lib/rack/sendfile.rb:114:in `call'
     # /home/ixti/.gem/ruby/3.3.4/gems/rack-cors-2.0.2/lib/rack/cors.rb:102:in `call'
     # /home/ixti/.gem/ruby/3.3.4/gems/railties-7.2.1/lib/rails/engine.rb:535:in `call'
     # /home/ixti/.gem/ruby/3.3.4/gems/rack-test-2.1.0/lib/rack/test.rb:360:in `process_request'
     # /home/ixti/.gem/ruby/3.3.4/gems/rack-test-2.1.0/lib/rack/test.rb:153:in `request'
     # ./spec/support/helpers/protobufable_test_helper.rb:76:in `rpc'
     # ./spec/requests/asgard/auth_controller_spec.rb:220:in `block (4 levels) in <top (required)>'
     # /home/ixti/.gem/ruby/3.3.4/gems/webmock-3.23.1/lib/webmock/rspec.rb:39:in `block (2 levels) in <top (required)>'
```

### RSpec Config (Sentry related)

``` ruby
require "sentry/test_helper"

RSpec.configure do |config|
  config.include Sentry::TestHelper
  config.before { setup_sentry_test }
  config.after { teardown_sentry_test }
end
```

### Workaround

As a workaround, we just disabled session tracking:

``` ruby
Sentry.init do |config|
  # ...

  # See: https://github.com/getsentry/sentry-ruby/pull/2337
  config.excluded_exceptions << "Sidekiq::JobRetry::Handled"

  # See: https://github.com/getsentry/sentry-ruby/pull/2396
  config.auto_session_tracking = false

  # ...
end
```

---

Resolves: https://github.com/getsentry/sentry-ruby/issues/2378

### DIFF
--- a/sentry-ruby/lib/sentry/hub.rb
+++ b/sentry-ruby/lib/sentry/hub.rb
@@ -255,7 +255,11 @@ module Sentry
 
       return unless session
       session.close
-      Sentry.session_flusher.add_session(session)
+
+      # NOTE: Under some circumstances, session_flusher nilified out of sync
+      #   See: https://github.com/getsentry/sentry-ruby/issues/2378
+      #   See: https://github.com/getsentry/sentry-ruby/pull/2396
+      Sentry.session_flusher&.add_session(session)
     end
 
     def with_session_tracking(&block)


ixti:
We tried and it still fails:

``` ruby
require "sentry/test_helper"

RSpec.configure do |config|
  config.include Sentry::TestHelper

  config.before do
    setup_sentry_test { _1.auto_session_tracking = true }
  end

  config.after { teardown_sentry_test }
end
```
```

**Comment by st0012** [comment]
Thanks for reporting the issue. It does look very puzzling as the only method that could nullify the session flusher is `Sentry.close`. Can you do a quick search to see if it's called anywhere in the test suite?

**Comment by MrSerth** [comment]
I am experiencing the same issue on production and hence reported #2378. Currently, the issue appears rather frequent for us, but not for each request:

Bildschirmfoto 2024-10-10 um 20 15 26

Is there any chance I could help to get this merged and released? Without a fix, each occurrence ends up with a 500 Internal Server Error and hides actually errors. I would rather miss a session tracing than having those errors.

In our app, we never call `Sentry.close` ourselves.

**Comment by ixti** [comment]
> Thanks for reporting the issue. It does look very puzzling as the only method that could nullify the session flusher is `Sentry.close`. Can you do a quick search to see if it's called anywhere in the test suite or codebase?

We don't call `Sentry.close` anywhere.

**Comment by MrSerth** [comment]
> Could you grab sentry from solnic/make-closing-thread-safe branch and see if the crash is gone?

I tried the branch, but since we modified the codebase (where this issue normally occurred) a lot recently, I wasn't able to confirm the fix is working. Sorry for not being able to help out here. Maybe it's still valid to merge this PR using the safe navigation operator?

**Comment by ixti** [comment]
We also took a workaround way by disabling session tracking completely. And even if we go back, the bug is pretty sporadic, to being able to tell if the patch in solnic/make-closing-thread-safe fixes the bug or no.

**Comment by solnic** [comment]
Thanks folks. Would it be possible for you to disable your workarounds after we release a new version and see if it helped. The proposed fix here comes with a risk that it will silence actual failures where the session_flusher should really be there. It's kinda hard to tell which option would be worse but given that the fix addresses an edge case by making a change that may result in breaking more common cases (an actual misconfiguration of Sentry), then it's probably better not to apply the fix.

**Comment by MrSerth** [comment]
As written above, I already tested the branch and partially reverted some of the code changes I considered to have the most impact on this issue.

While I understand your concern, this issue (and especially the one discussed in #2428) may hide actually bugs from our code base as Sentry customers -- that happened at least for us. We would normally trust in the Sentry error reporting, but when this bug occurs, we lack any notification and insights on actual defects of our software. Hence, from our perspective, it is far less critical to miss some sessions than to miss the actual bug that occurred.

**Comment by solnic** [comment]
> As written above, I already tested the branch and partially reverted some of the code changes I considered to have the most impact on this issue.

Gotcha, thank you!

> While I understand your concern, this issue (and especially the one discussed in #2428) may hide actually bugs from our code base as Sentry customers -- that happened at least for us. We would normally trust in the Sentry error reporting, but when this bug occurs, we lack any notification and insights on actual defects of our software. Hence, from our perspective, it is far less critical to miss some sessions than to miss the actual bug that occurred.

Alright, given that Hub is going away eventually (see https://github.com/getsentry/sentry-ruby/issues/1495#issuecomment-1881400829) we can just rely on the workarounds here and in your PR for the time being. Thanks for all your input and help with this btw!

File: sentry-ruby/lib/sentry/hub.rb [hunk 1]
```diff
@@ -248,7 +248,7 @@ def end_session
 
       return unless session
       session.close
-      Sentry.session_flusher.add_session(session)
+      Sentry.session_flusher&.add_session(session)
```
solnic, line 251: Could you add a comment pointing to the related issue so that it's clear why we're doing this? 🙏🏻

ixti:
> Would it be possible for you to disable your workarounds after we release a new version and see if it helped.

Sure, we will give it a try!

sentry-ruby/lib/sentry/hub.rb
ixti: Will do so shortly.

ixti: Updated

Vagab:
hey guys. So we've been running into `NoMethodError add_session for nil` for quite some time now, but the errors were intermittent.
Recently we were able to reproduce it with one `seed` and 3 specs. I haven't tried to reproduce it in a blank app/script yet, but the general problem I'm seeing is that if we're running `setup_sentry_test` **before** any `request` spec is ran then for every request spec that follows we'll get this error. If any request spec runs before the spec that does `setup_sentry_test` then we don't get the error at all. Which explains why it was intermittent for us. But it also tells me that the issue is not with threads, I think(cc @solnic), since it will always happen if the specs are ran in the order I mentioned above.
I have no idea if there's any connection between when sentry is initialized and request specs, or if the connection is unique to our project.
As a workaround, this:
```ruby
before do
  setup_sentry_test { |it| it.auto_session_tracking = false }
end
```
works, but meh.
I should also mention that the issue persists on `master` branch.
I will try to create a simple reproduction and/or a fix, but just wanted to provide some info in case it can help anyone 👍

Let me know if I can be of any further assistance here

solnic:
@vegab oh these are great insights, thank you! I'll look into this again

solnic (approved):
OK let's get this workaround merged. I could not reproduce this so we still don't really know what the culprit is. I've been running a sample test suite with auto_session_tracking for hours now and it just keeps passing 🙃 These type of issues will go away eventually once Hub situation is solved.

PR #2402, opened 2024-09-15
Prevent Hub from having nil scope and client

st0012:
`Hub#pop_scope` currently can cause a Hub to have a nil scope and client, which would highly likely lead of errors. For example, `Hub#configuration` would simply cause `NoMethodError`.

This commit prevents that from ever happening by making sure that `Hub#pop_scope` would at least leave a brand new scope and the old client on the stack.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-### 5.20.0
+## 5.20.0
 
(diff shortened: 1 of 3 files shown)


sl0thentr0py (approved):
thx!

Issue #2322, opened 2024-06-17
Queues tracking for ActiveJob?


nhorton:
Is there any ETA on Queues tracking for ActiveJob? We would love to try the feature, but I really don't want to write an integration when we know it is going to be replaced by something standard shortly thereafter.

codealchemy:
Looks like it may still be on hold -> https://github.com/getsentry/sentry-ruby/issues/2045

nhorton:
> @codealchemy that's the profiling feature, that's unrelated. @nhorton is talking about https://docs.sentry.io/product/insights/queue-monitoring/ which we will get to soon (like next 1-2 months).

Thanks for the update. Adding a Roadmap in Github, even if it does not have date commitments, would be helpful on stuff like this. Just knowing things like where this is relative to token count support for LLMs and such is useful context for customers.

frederikspang:
Would be very useful to include Sidekiq in this feature.

olivier-thatch:
@sl0thentr0py Are there any updates on this? I wrote my own pure-ActiveJob integration based on the [Sidekiq one](https://github.com/getsentry/sentry-ruby/blob/master/sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb), and AFAICT it works okay. Happy to submit a PR if that's helpful.

sl0thentr0py:
@olivier-thatch yes feel free to make a PR, that would be very appreciated!

PR #2403, opened 2024-09-16
Add child_spans for Sidekiq Queue instrumentation

frederikspang:
Implements #2322 for Sidekiq.

## Description
Adds support for Sentry Queues page (With predefined span `op` names)

https://docs.sentry.io/platforms/ruby/guides/sidekiq/tracing/instrumentation/custom-instrumentation/queues-module/

Screenshot 2024-09-16 at 15 45 15

### DIFF
--- a/sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
+++ b/sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
@@ -8,7 +8,7 @@ module Sentry
       OP_NAME = "queue.sidekiq"
       SPAN_ORIGIN = "auto.queue.sidekiq"
 
-      def call(_worker, job, queue)
+      def call(worker, job, queue)
         return yield unless Sentry.initialized?
 
         context_filter = Sentry::Sidekiq::ContextFilter.new(job)
@@ -26,8 +26,18 @@ module Sentry
         scope.set_span(transaction) if transaction
 
         begin
-          yield
-        rescue
+          Sentry.with_child_span(op: "queue.process", description: "Process #{worker.class.name}") do |span|
+            # Set span data
+            if span
+              span.set_data("messaging.message.id", job["jid"])
+              span.set_data("messaging.destination.name", queue)
+              span.set_data("messaging.message.receive.latency", ((Time.now.to_f - job["enqueued_at"]) * 1000).to_i)
+              span.set_data("messaging.message.retry.count", job["retry_count"] || 0)
+            end
+
+            yield
+          end
+        rescue => ex
           finish_transaction(transaction, 500)
           raise
         end
@@ -63,13 +73,22 @@ module Sentry
     end
 
     class SentryContextClientMiddleware
-      def call(_worker_class, job, _queue, _redis_pool)
+      def call(worker_class, job, queue, _redis_pool)
         return yield unless Sentry.initialized?
 
         user = Sentry.get_current_scope.user
         job["sentry_user"] = user unless user.empty?
         job["trace_propagation_headers"] ||= Sentry.get_trace_propagation_headers
-        yield
+
+        Sentry.with_child_span(op: "queue.publish", description: "Enqueue #{worker_class}") do |span|
+          # Set span data
+          if span
+            span.set_data("messaging.message.id", job["jid"])
+            span.set_data("messaging.destination.name", queue)
+          end
+
+          yield
+        end
       end
     end
   end
--- a/sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
+++ b/sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
@@ -37,7 +37,7 @@ module Sentry
 
             yield
           end
-        rescue => ex
+        rescue
           finish_transaction(transaction, 500)
           raise
         end
--- a/sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
+++ b/sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
@@ -4,7 +4,20 @@ require "sentry/sidekiq/context_filter"
 
 module Sentry
   module Sidekiq
+    module Helpers
+      def set_span_data(span, id:, queue:, latency: nil, retry_count: nil)
+        if span
+          span.set_data("messaging.message.id", id)
+          span.set_data("messaging.destination.name", queue)
+          span.set_data("messaging.message.receive.latency", latency) if latency
+          span.set_data("messaging.message.retry.count", retry_count) if retry_count
+        end
+      end
+    end
+
     class SentryContextServerMiddleware
+      include Sentry::Sidekiq::Helpers
+
       OP_NAME = "queue.process"
       SPAN_ORIGIN = "auto.queue.sidekiq"
 
@@ -27,10 +40,7 @@ module Sentry
         if transaction
           scope.set_span(transaction)
 
-          transaction.set_data("messaging.message.id", job["jid"])
-          transaction.set_data("messaging.destination.name", queue)
-          transaction.set_data("messaging.message.receive.latency", ((Time.now.to_f - job["enqueued_at"]) * 1000).to_i)
-          transaction.set_data("messaging.message.retry.count", job["retry_count"] || 0)
+          set_span_data(transaction, id: job["jid"], queue: queue, latency: ((Time.now.to_f - job["enqueued_at"]) * 1000).to_i, retry_count: job["retry_count"] || 0)
         end
 
         begin
@@ -71,6 +81,8 @@ module Sentry
     end
 
     class SentryContextClientMiddleware
+      include Sentry::Sidekiq::Helpers
+
       def call(worker_class, job, queue, _redis_pool)
         return yield unless Sentry.initialized?
 
@@ -78,12 +90,8 @@ module Sentry
         job["sentry_user"] = user unless user.empty?
         job["trace_propagation_headers"] ||= Sentry.get_trace_propagation_headers
 
-        Sentry.with_child_span(op: "queue.publish", description: "Enqueue #{worker_class}") do |span|
-          # Set span data
-          if span
-            span.set_data("messaging.message.id", job["jid"])
-            span.set_data("messaging.destination.name", queue)
-          end
+        Sentry.with_child_span(op: "queue.publish", description: worker_class.to_s) do |span|
+          set_span_data(span, id: job["jid"], queue: queue)
 
           yield
         end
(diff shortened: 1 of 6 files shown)


sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
```diff
@@ -26,8 +26,19 @@ def call(_worker, job, queue)
...
         begin
-          yield
-        rescue
+          Sentry.with_child_span(op: "queue.process", description: "Process #{worker.class.name}") do |span|
+            # Set span data
+            if span
+              span.set_data("messaging.message.id", job["jid"])
+              span.set_data("messaging.destination.name", queue)
+              span.set_data("messaging.message.body.size", job.to_json.size)
+              span.set_data("messaging.message.receive.latency", ((Time.now.to_f - job["created_at"]) * 1000).to_i)
```
nateberkopec, line 35: You should use enqueued_at instead. Using created_at will cause incorrect calculations for scheduled jobs.

sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
```diff
@@ -63,13 +74,23 @@ def finish_transaction(transaction, status)
...
         job["sentry_user"] = user unless user.empty?
         job["trace_propagation_headers"] ||= Sentry.get_trace_propagation_headers
-        yield
+
+        Sentry.with_child_span(op: "queue.publish", description: "Enqueue #{worker_class}") do |span|
+          # Set span data
+          if span
+            span.set_data("messaging.message.id", job["jid"])
+            span.set_data("messaging.destination.name", queue)
+            span.set_data("messaging.message.body.size", job.to_json.size)
```
nateberkopec, line 89: I wonder if there's a better way to do this. Dumping the job to json (an extra time) on every enqueue sounds not cheap.

nateberkopec (changes requested):

sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
frederikspang: I was wondering as well - I couldn't really figure out a good way, to give a good insight to size. Could be `job.args.size` but thats misleading; counting only the amount of arguments, not their actual size (Assuming the remaining payload is somewhat constant in size (retry, queue, class, created_at and enqueued_at keys).

nateberkopec: Having the accurate size of what's being pushed to Redis is cool, but IMO not worth the cost of encoding the entire thing twice.

frederikspang: I'm inclined to agree. I'll remove this one - I'm thinking core sentry team can make the final call.

nateberkopec (approved):

rwojsznis:
Oh wow, exactly what I was looking for 😅 🙌 

can we somehow help to get this shipped? 💪

solnic (commented):
Thanks for working on this! 🙇🏻 Would you be able to add some basic specs for this too?

frederikspang:
@solnic I've added two basic specs.
Latency is a bit difficult, without introducing some `timecop` kind of tool. Which we do have in sentry-ruby, so we could just introduce it here if we wanted to.

frederikspang:
@solnic After looking at these, I think queue.process may actually replace `op: queue.sidekiq` instead of being a child span.
They will always be redundant - But we should set origin or some other field to `sidekiq`, description or name maybe, to differentiate the processor. (ie. using another processor alongside Sidekiq - or doing queries in Sentry UI across applications)

image

nateberkopec:
Looks like it needs a changelog too

sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
```diff
@@ -26,8 +26,18 @@ def call(_worker, job, queue)
...
+            if span
+              span.set_data("messaging.message.id", job["jid"])
+              span.set_data("messaging.destination.name", queue)
+              span.set_data("messaging.message.receive.latency", ((Time.now.to_f - job["enqueued_at"]) * 1000).to_i)
+              span.set_data("messaging.message.retry.count", job["retry_count"] || 0)
+            end
+
+            yield
+          end
+        rescue => ex
```
nateberkopec, line 40: Not sure it matters, but you don't use this exception object, so you may want to not even bother with creating it?

frederikspang: You're right - Fixed.

solnic:
> Latency is a bit difficult, without introducing some `timecop` kind of tool. Which we do have in sentry-ruby, so we could just introduce it here if we wanted to.

Thanks for adding the specs. I think it would be good to use timecop there actually.

sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
```diff
@@ -23,7 +23,15 @@ def call(_worker, job, queue)
...
         scope.set_transaction_name(context_filter.transaction_name, source: :task)
         transaction = start_transaction(scope, job["trace_propagation_headers"])
-        scope.set_span(transaction) if transaction
+
+        if transaction
+          scope.set_span(transaction)
+
+          transaction.set_data("messaging.message.id", job["jid"])
+          transaction.set_data("messaging.destination.name", queue)
+          transaction.set_data("messaging.message.receive.latency", ((Time.now.to_f - job["enqueued_at"]) * 1000).to_i)
```
solnic, line 32: Are we sure that both time objects will use the same timezone?

sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
```diff
@@ -63,13 +71,22 @@ def finish_transaction(transaction, status)
...
-      def call(_worker_class, job, _queue, _redis_pool)
+      def call(worker_class, job, queue, _redis_pool)
         return yield unless Sentry.initialized?
 
         user = Sentry.get_current_scope.user
         job["sentry_user"] = user unless user.empty?
         job["trace_propagation_headers"] ||= Sentry.get_trace_propagation_headers
-        yield
+
+        Sentry.with_child_span(op: "queue.publish", description: "Enqueue #{worker_class}") do |span|
```
solnic, line 81: Just using worker_class would be better here, as it would make searching simpler.

sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
```diff
@@ -63,13 +71,22 @@ def finish_transaction(transaction, status)
...
+          # Set span data
+          if span
+            span.set_data("messaging.message.id", job["jid"])
+            span.set_data("messaging.destination.name", queue)
+          end
```
solnic, line 86: We could encapsulate this somehow as both classes need to do this.

solnic (changes requested):

frederikspang:
> > @solnic After looking at these, I think queue.process may actually replace `op: queue.sidekiq` instead of being a child span.
> > They will always be redundant
> > 
> > * But we should set origin or some other field to `sidekiq`, description or name maybe, to differentiate the processor. (ie. using another processor alongside Sidekiq - or doing queries in Sentry UI across applications)
> 
> I've been testing this out and I don't see queue.sidekiq/queue.process. This is what I see instead:
> 
> Screenshot 2024-10-25 at 14 51 35

You're right! I changed that after my comment, before your review here: 

https://github.com/getsentry/sentry-ruby/pull/2403/commits/cd3a1df20eb357498f91bf4217c1667e4d4839aa

I'll have a look at the other comments as soon as possible.

sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
frederikspang: Should be - enqueued_at is generated with `Time.now.to_f` as well in Sidekiq.

frederikspang:
@solnic I have updated PR with feedback adjusted!

frederikspang:
@solnic The specs failing seems related to https://github.com/sidekiq/sidekiq/pull/6430 - Was able to reproduce locally, after clearing Lock file and bundle install again.

Somewhat unrelated to this PR, although it's noisy.

solnic (approved):
Great job thank you!

solnic:
> @solnic The specs failing seems related to [sidekiq/sidekiq#6430](https://github.com/sidekiq/sidekiq/pull/6430) - Was able to reproduce locally, after clearing Lock file and bundle install again.
> 
> Somewhat unrelated to this PR, although it's noisy.

Please update to the latest master. I just checked locally and it passes. Once CI is green I'll merge it in.

frederikspang:
@solnic Done!

nateberkopec:
🙌  Thanks everyone! This is such a great feature.

sentry-sidekiq/spec/sentry/sidekiq_spec.rb
```diff
@@ -223,7 +223,7 @@ def retry_last_failed_job
       expect(transaction.contexts.dig(:trace, :trace_id)).to be_a(String)
       expect(transaction.contexts.dig(:trace, :span_id)).to be_a(String)
       expect(transaction.contexts.dig(:trace, :status)).to eq("ok")
-      expect(transaction.contexts.dig(:trace, :op)).to eq("queue.sidekiq")
```
Linuus, line 226: @frederikspang Did this break sampling? In the [docs](https://docs.sentry.io/platforms/ruby/guides/sidekiq/configuration/sampling/#setting-a-sampling-function) you show that you can do 

```rb
case op
when /sidekiq/
  0.01
[...]
```

But this is always set to `queue.process` now so we are sampling a lot more than we should which is costing us money...

Was this intentional or am I misunderstanding something?

frederikspang: That is very much not intentional. I will submit a PR for docs as well. I was unaware of this example being there.

Linuus: Yeah we've been running out of performance traces more often recently and were confused what had happened. And then I found that our sampler was not running correctly :-/

Linuus: Also, this is not all about the docs. It's like a public API and can't just be changed like this without proper deprecations etc. It must be costing your customers a lot of money in both extra Sentry charges as well as debugging time.

frederikspang: I get what you're saying - Except it's not _my_ customers. I've done this as OSS contribution to help the Sentry Ruby SDK conform to the guidelines and "predefined operations" as defined by the Sentry core system - ie making the `Queues` page work properly for Sidekiq jobs.

Deprecating and semantic versioning of what goes into what release version is very not much my job.

Linuus: Aha I thought you worked for Sentry :)
Sorry if my tone was harsh, I didn’t mean to offend. I know mistakes can be made.
My message was more directed to Sentry the business.

Also the feature is great and many thanks for implementing it! 🙏

PR #2417, opened 2024-09-26
fix: Unable to use sentry when URI contains non-ascii symbols on Net::HTTP traces.

### LINKED ISSUE #2418: Net::HTTP tracing doesn't support URIs with non ascii characters.
### Issue Description

When building traces for Net::HTTP sentry does not escape the string passed to `URI.parse` resulting in `URI::InvalidURIError` being raised if non-ascii characters are present in the path.

— https://github.com/getsentry/sentry-ruby/blob/master/sentry-ruby/lib/sentry/net/http.rb#L69

### Reproduction Steps

While using sentry, use Net::HTTP to send a request to a URI using non-ascii characters.

Sentry has no additional configuration:

```ruby
Sentry.init do |sentry_config|
  sentry_config.dsn = config.fetch(:dsn)
end
```

### Expected Behavior

When sentry tries to extract [the request information,](https://github.com/getsentry/sentry-ruby/blob/master/sentry-ruby/lib/sentry/net/http.rb#L69) there should not be errors when the path contains non-ascii characters.

### Actual Behavior

When sentry tries to extract the request information on `Net::HTTP` request traces, if the path contains non-ascii characters, the trace results in a `URI::InvalidURIError`.

### Ruby Version

2.7, 3.x

### SDK Version

5.19.0

- sentry-ruby
- sentry-rails
- sentry-delayed_job

### Sentry Config

```ruby
Sentry.init do |config|
  config.breadcrumbs_logger = [:active_support_logger, :http_logger, :sentry_logger]
  config.dsn = Rails.application.secrets.sentry_dsn
  config.enable_tracing = true
  config.excluded_exceptions += ['ActionController::RoutingError', 'ActiveRecord::RecordNotFound']
  config.delayed_job.report_after_job_retries = true
end
```

rascencio-kraft:
## Description

Fixes #2418

Using URI::DEFAULT_PARSER.escape to properly escape the constructed URLs, which prevents issues with special characters that may cause errors during URI parsing.

## Changelog

- Added URI::DEFAULT_PARSER.escape to handle special characters in URL construction and prevent parsing errors on Net::HTTP traces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 - Fix error events missing a DSC when there's an active span ([#2408](https://github.com/getsentry/sentry-ruby/pull/2408))
 - Verifies presence of client before adding a breadcrumb ([#2394](https://github.com/getsentry/sentry-ruby/pull/2394))
+- Fix `Net:HTTP` integration for non-ASCII URI's ([#2417](https://github.com/getsentry/sentry-ruby/pull/2417))
 
 ## 5.19.0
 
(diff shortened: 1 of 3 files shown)


sl0thentr0py (approved):
ty!

PR #2421, opened 2024-09-30
Generalize size limits based on envelope item type

### LINKED ISSUE #2413: Generalize size limits based on envelope item type
we need to generalize [this logic to have different limits](https://github.com/getsentry/sentry-ruby/blob/04e030f974558aa750e26afc458c48e663ddcfcc/sentry-ruby/lib/sentry/envelope.rb#L42-L56) according to different envelope `item.type` but let's make a separate issue/PR for that.

_Originally posted by @sl0thentr0py in https://github.com/getsentry/sentry-ruby/issues/2372#issuecomment-2368615966_

solnic:
Closes #2413


sentry-ruby/lib/sentry/envelope/item.rb
```diff
@@ -0,0 +1,88 @@
+# frozen_string_literal: true
+
+module Sentry
+  # @api private
+  class Envelope::Item
+    STACKTRACE_FRAME_LIMIT_ON_OVERSIZED_PAYLOAD = 500
+    MAX_SERIALIZED_PAYLOAD_SIZE = 1024 * 1000
+
+    SIZE_LIMITS = Hash.new(MAX_SERIALIZED_PAYLOAD_SIZE).update(
+      "profile" => 1024 * 1000 * 10
```
solnic, line 10: @sl0thentr0py wasn't sure what it should be exactly, so please let me know what value we want here and I'll adjust it + update the spec accordingly

PR #2432, opened 2024-10-15
Fix bundle cache on CI

solnic:
We didn't set BUNDLE_GEMFILE correctly so caching bundles didn't work. This PR fixes it which should give us a significant speed bump when running this massive test suite.

#skip-changelog

### DIFF
--- a/.github/workflows/sentry_resque_test.yml
+++ b/.github/workflows/sentry_resque_test.yml
@@ -25,7 +25,12 @@ jobs:
         working-directory: sentry-resque
     name: Ruby ${{ matrix.ruby_version }}, options - ${{ toJson(matrix.options) }}
     runs-on: ubuntu-latest
+    env:
+      RUBYOPT: ${{ matrix.options.rubyopt }}
+      BUNDLE_GEMFILE: ${{ github.workspace }}/sentry-resque/Gemfile_with_rails.rb
+      BUNDLE_WITHOUT: rubocop
     strategy:
+      fail-fast: false
       matrix:
         ruby_version: ${{ fromJson(needs.ruby-versions.outputs.versions) }}
         include:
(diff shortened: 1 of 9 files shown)


.github/workflows/sentry_resque_test.yml
```diff
@@ -53,17 +58,10 @@ jobs:
...
+        run: bundle exec rake
 
       - name: Run specs with Rails
-        env:
-          BUNDLE_GEMFILE: Gemfile_with_rails.rb
-          RUBYOPT: ${{ matrix.options.rubyopt }}
-        run: |
-          bundle install --jobs 4 --retry 3
-          bundle exec rake
+        run: bundle exec rake
```
st0012, line 64: We can remove this step as the previous one would already run with Rails.

.github/workflows/sentry_resque_test.yml
```diff
@@ -25,7 +25,12 @@ jobs:
         working-directory: sentry-resque
     name: Ruby ${{ matrix.ruby_version }}, options - ${{ toJson(matrix.options) }}
     runs-on: ubuntu-latest
+    env:
+      RUBYOPT: ${{ matrix.options.rubyopt }}
+      BUNDLE_GEMFILE: ${{ github.workspace }}/sentry-resque/Gemfile_with_rails.rb
```
st0012, line 30: TODO: I think we can just move Rails related dependencies to `Gemfile` later

.github/workflows/sentry_sidekiq_test.yml
```diff
@@ -60,12 +66,7 @@ jobs:
           redis-version: ${{ matrix.sidekiq_version == '7.0' && 6 || 5 }}
 
       - name: Run specs with Sidekiq ${{ matrix.sidekiq_version }}
-        env:
-          SIDEKIQ_VERSION: ${{ matrix.sidekiq_version }}
-          RUBYOPT: ${{ matrix.options.rubyopt }}
-        run: |
-          bundle install --jobs 4 --retry 3
-          make test
+        run: make test
```
st0012, line 69: I think we can move `make test`'s definitions here and remove it as other Makefiles don't seem to have it anymore either.

.github/workflows/sentry_resque_test.yml
solnic: @st0012 I made it so that it runs with and without Rails, as I reckon that was the original intention. WDYT?

.github/workflows/sentry_resque_test.yml
solnic: @st0012 I removed it already and added a rails group to the Gemfile, then added `BUNDLE_WITHOUT="rails" bundle exec rake` to run tests w/o rails loaded. Seems to work well.

Issue #2386, opened 2024-08-22
NoMethodError: undefined method `rails` for nil


MrSerth:
### Issue Description

In my server logs (but not in Sentry), I found another `NoMethodError`: This time, `rails` was called on `nil`.

<details><summary>Stack Trace of the Issue</summary>
<p>

```
NoMethodError (undefined method `rails' for nil):

sentry-rails (5.18.2) lib/sentry/rails/rescued_exception_interceptor.rb:20:in `report_rescued_exceptions?'
sentry-rails (5.18.2) lib/sentry/rails/rescued_exception_interceptor.rb:14:in `rescue in call'
sentry-rails (5.18.2) lib/sentry/rails/rescued_exception_interceptor.rb:11:in `call'
actionpack (7.1.3.4) lib/action_dispatch/middleware/debug_exceptions.rb:29:in `call'
sentry-ruby (5.18.2) lib/sentry/rack/capture_exceptions.rb:30:in `block (2 levels) in call'
sentry-ruby (5.18.2) lib/sentry/hub.rb:258:in `with_session_tracking'
sentry-ruby (5.18.2) lib/sentry-ruby.rb:404:in `with_session_tracking'
sentry-ruby (5.18.2) lib/sentry/rack/capture_exceptions.rb:21:in `block in call'
sentry-ruby (5.18.2) lib/sentry/hub.rb:59:in `with_scope'
sentry-ruby (5.18.2) lib/sentry-ruby.rb:384:in `with_scope'
sentry-ruby (5.18.2) lib/sentry/rack/capture_exceptions.rb:20:in `call'
actionpack (7.1.3.4) lib/action_dispatch/middleware/show_exceptions.rb:31:in `call'
railties (7.1.3.4) lib/rails/rack/logger.rb:37:in `call_app'
railties (7.1.3.4) lib/rails/rack/logger.rb:24:in `block in call'
activesupport (7.1.3.4) lib/active_support/tagged_logging.rb:135:in `block in tagged'
activesupport (7.1.3.4) lib/active_support/tagged_logging.rb:39:in `tagged'
activesupport (7.1.3.4) lib/active_support/tagged_logging.rb:135:in `tagged'
activesupport (7.1.3.4) lib/active_support/broadcast_logger.rb:240:in `method_missing'
railties (7.1.3.4) lib/rails/rack/logger.rb:24:in `call'
[...]

PR #2436, opened 2024-10-18
Make Sentry.{close,get_main_hub} thread-safe

solnic:
Since we're dealing with a bunch of global vars in `Sentry.close`, it should be wrapped in a mutex sync because threads may be accessing these vars while Sentry is closing, and THAT *probably* causes #2386 and #2396.

#skip-changelog

### DIFF
--- a/sentry-ruby/lib/sentry-ruby.rb
+++ b/sentry-ruby/lib/sentry-ruby.rb
@@ -50,6 +50,8 @@ module Sentry
 
   THREAD_LOCAL = :sentry_hub
 
+  MUTEX = Mutex.new
+
   class << self
     # @!visibility private
     def exception_locals_tp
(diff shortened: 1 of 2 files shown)


solnic:
`Sentry.initialized?` should use a mutex too, given that it relies on `@main_hub` that may get reset to `nil` by `Sentry.close`

sentry-ruby/lib/sentry-ruby.rb
```diff
@@ -248,35 +251,37 @@ def init(&block)
...
+          end
         end
-      end
 
-      @background_worker.shutdown
+        @background_worker.shutdown
 
-      @main_hub = nil
-      Thread.current.thread_variable_set(THREAD_LOCAL, nil)
+        @main_hub = nil
```
sl0thentr0py, line 282: let's just wrap lines 282 and 283 in the mutex, the other stuff is fine.
And I agree that `get_main_hub` should also be within a mutex sync

sl0thentr0py (changes requested):

sl0thentr0py (approved):

st0012 (approved):

PR #2439, opened 2024-10-19
Enforce latest rubygems on CI

solnic:
This makes sure that we use latest rubygems on CI. Without this we're seeing random failures caused by bugs in old Rubygems when bundling may crash with ArgumentError.

I discovered this through #2436 failing like that: https://github.com/getsentry/sentry-ruby/actions/runs/11411361838/job/31767045829?pr=2436#step:4:123

#skip-changelog

### DIFF
--- a/.github/workflows/sentry_sidekiq_test.yml
+++ b/.github/workflows/sentry_sidekiq_test.yml
@@ -36,7 +36,7 @@ jobs:
           - { ruby_version: 2.6, sidekiq_version: 5.0 }
           - { ruby_version: 2.6, sidekiq_version: 6.0 }
           - { ruby_version: jruby, sidekiq_version: 5.0 }
-          - { ruby_version: jruby, sidekiq_version: 6.0 }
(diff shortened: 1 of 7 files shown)


.github/workflows/sentry_sidekiq_test.yml
```diff
@@ -36,7 +36,7 @@ jobs:
           - { ruby_version: 2.6, sidekiq_version: 5.0 }
           - { ruby_version: 2.6, sidekiq_version: 6.0 }
           - { ruby_version: jruby, sidekiq_version: 5.0 }
-          - { ruby_version: jruby, sidekiq_version: 6.0 }
+          - { ruby_version: jruby, sidekiq_version: 6.5 }
```
solnic, line 39: With latest rubygems under jruby, bundling sidekiq using 6.0 as the version spec would install sidekiq 6.0.x - and that version actually crashes with `Sidekiq[]=` not being defined error.

st0012 (approved):

PR #2444, opened 2024-10-26
Test sentry-rails against Rails 8.0.0

st0012:
This PR:
- Includes a commit from #2445, which can be merged separately or together
- Re-organizes Gemfile's Ruby/Rails conditions
- Updates Rails 8 specific tests
- Adds Rails 8.0.0.rc1 as test target (which can easily be updated to 8.0)

#skip-changelog

PR #2445, opened 2024-10-26
Simplify ActiveJob specs setup

st0012:
We don't need to use Sidekiq for ActiveJob specs, we can use Rails' default adapters instead.

#skip-changelog


solnic (commented):
Just like in the other PR - we probably want to remove sidekiq from Gemfile now

st0012:
Merged as part of #2444

PR #2452, opened 2024-10-31
Don't gate request query strings and form data behind send_default_pii

sl0thentr0py:


sl0thentr0py:
The failing test seems to be about new behavior rails 7.2 and will be fixed separately

cleptric:
Query strings should not be guarded by `send_default_pii`, but it should still apply for the request body. https://develop.sentry.dev/sdk/expected-features/data-handling/#sensitive-data

sentry-ruby/lib/sentry/utils/http_tracing.rb
```diff
@@ -20,7 +20,7 @@ def record_sentry_breadcrumb(request_info, response_status)
           level: :info,
           category: self.class::BREADCRUMB_CATEGORY,
           type: :info,
-          data: { status: response_status, **request_info }
+          data: { status: response_status, **request_info.compact }
```
st0012, line 23: IMO, this should be called at the caller's `extract_request_info` helpers.

st0012 (commented):
I also think `send_default_pii` should be applied to body.

PR #2464, opened 2024-11-14
[rails] support string errors in error reporter

### LINKED ISSUE #2460: Using `ActiveSupport::ErrorReporter` with `capture_message`
I'm sorry if it already has been discussed, I couldn't find any info on this topic. `sentry-rails` has built-in integration with ActiveSupport's `ErrorReporter` which allows to report exceptions with service-agnostic `Rails.error` API. Reporting exceptions works fine, but surprisingly it does not work (AFAIK) with sentry's message capturing feature.

It seems like there's no restriction on reported object's type from AS' side; `Rails.error.report("foobar")` propagates message to `capture_exception` where it fails with "ArgumentError: expect the argument to be a Exception, got String" exception.

I'm curious if it's possible to extend sentry's subscriber to propagate strings to `capture_message`? Is there a reason it's not implemented yet?

solnic:
Closes #2460

### DIFF
--- a/sentry-rails/lib/sentry/rails/error_subscriber.rb
+++ b/sentry-rails/lib/sentry/rails/error_subscriber.rb
@@ -35,8 +35,12 @@ module Sentry
         when Exception
           Sentry::Rails.capture_exception(error, **options)
         else
-          raise ArgumentError, "Expected an Exception or a String, got: #{error.inspect}"
+          log_debug("Expected an Exception or a String, got: #{error.inspect}")
         end
+      end
+
+      def log_debug(message)
+        Sentry.configuration.logger.debug(message)
       end
     end
   end
--- a/sentry-rails/lib/sentry/rails/error_subscriber.rb
+++ b/sentry-rails/lib/sentry/rails/error_subscriber.rb
@@ -27,7 +27,16 @@ module Sentry
           hint.merge!(context.delete(:hint))
         end
 
-        Sentry::Rails.capture_exception(error, level: severity, contexts: { "rails.error" => context }, tags: tags, hint: hint)
+        options = { level: severity, contexts: { "rails.error" => context }, tags: tags, hint: hint }
+
+        case error
+        when String
+          Sentry::Rails.capture_message(error, **options)
+        when Exception
+          Sentry::Rails.capture_exception(error, **options)
+        else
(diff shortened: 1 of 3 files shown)


sentry-rails/lib/sentry/rails/error_subscriber.rb
```diff
@@ -27,7 +27,16 @@ def report(error, handled:, severity:, context:, source: nil)
...
-        Sentry::Rails.capture_exception(error, level: severity, contexts: { "rails.error" => context }, tags: tags, hint: hint)
+        options = { level: severity, contexts: { "rails.error" => context }, tags: tags, hint: hint }
+
+        case error
+        when String
+          Sentry::Rails.capture_message(error, **options)
+        when Exception
+          Sentry::Rails.capture_exception(error, **options)
+        else
+          raise ArgumentError, "Expected an Exception or a String, got: #{error.inspect}"
```
sl0thentr0py, line 38: we should not raise inside the SDK, let's use `log_debug` here

sl0thentr0py (changes requested):

sentry-rails/lib/sentry/rails/error_subscriber.rb
solnic: @sl0thentr0py hmm, what's the best way of accessing logger though? I can see that the helper is defined in the LoggingHelper, but not sure which object I should use for logging, or if I should include that helper in the ErrorSubscriber?

sl0thentr0py: yes you can just include it

solnic: @sl0thentr0py sorry, but could you tell me how I should configure the logger so that I could do something like this in the spec:

```ruby
expect {
          Rails.error.report(312, severity: :info, context: { foo: "bar" })
        }.to output(/Expected an Exception or a String, got: 312/).to_stdout
```

? Including the helper is not enough as it expects `@logger` which is not set on the ErrorReporter.

sl0thentr0py: yes for that you need the instance var, but you can also use `Sentry.configuration.logger` directly

PR #2471, opened 2024-11-22
Improve the accuracy of duration calculations in cron jobs monitoring

krzysztofjablonski:
This pull request includes changes to the `perform` method in the `sentry-ruby/lib/sentry/cron/monitor_check_ins.rb` file to improve the accuracy of duration calculations.

* Changed the duration calculation to use floating-point precision instead of integer precision to improve accuracy. [[1]](diffhunk://#diff-b5003612709a99eda69c66d664ad3dfed9ba11297720f5e94a231b19f5f084d0L22-R22) [[2]](diffhunk://#diff-b5003612709a99eda69c66d664ad3dfed9ba11297720f5e94a231b19f5f084d0L32-R32)

Before:

image

After:

image


whitemerry (approved):

solnic (commented):
Thank you for the PR! Actually, we should probably consider adding a helper for the duration and use the approach described here: https://blog.dnsimple.com/2018/03/elapsed-time-with-ruby-the-right-way/

This should give the best results.

whitemerry (approved):

solnic (approved):

PR #2473, opened 2024-11-23
Default to internal_error rather than unknown_error

cyberdelia:
When using `sentry-opentelemetry` to trace code that isn't related to an HTTP call, the `SpanProcessor` will set the status of a `Transaction` to be `"unknown_error"`, which is benign, but within Sentry UI this status is not interpreted as an error at all. 

The result is that all the impacted transactions appears to have 0% failure rate, etc. Changing it to `"internal_error"` prevents this, and also is in line with [other sentry SDKs](https://github.com/getsentry/sentry-python/blob/a7c2d704a90b374bc3c6a69365922614710d95f7/sentry_sdk/integrations/opentelemetry/span_processor.py#L304) behaviour.

Sidebar:  The `SpanEvent` metadata around exceptions seems to be dropped in the conversion process to `Transaction` , which is too bad.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Safe-navigate to session flusher [#2396](https://github.com/getsentry/sentry-ruby/pull/2396)
 - Fix latency related nil error for Sidekiq Queues Module span data [#2486](https://github.com/getsentry/sentry-ruby/pull/2486)
   - Fixes [#2485](https://github.com/getsentry/sentry-ruby/issues/2485)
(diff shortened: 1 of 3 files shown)


solnic (changes requested):
This makes sense, thanks. Could you also add a simple spec that covers this change to span_processor_spec.rb?

cyberdelia:
@solnic Done.

solnic (approved):
Looks good, thank you! Please update the top-level CHANGELOG.md too, and I'll be able to merge it in 🙂

cyberdelia:
@solnic Done!

PR #2474, opened 2024-11-25
Don't let `backtrace_cleanup_callback` affect `abs_path` and separate filename handling

### LINKED ISSUE #2472: Fix Source Code Extraction to Stack Traces (regression via rails)
**Describe the idea**
Currently, the Ruby Sentry SDK provides stack traces that include the file name, line number, and method name where an error occurred. However, it does not include the actual lines of source code. This feature is crucial for debugging, especially when dealing with third-party code, as it provides immediate context around the error without requiring developers to manually locate and open the files.

The feature request is to add functionality to the Ruby Sentry SDK to extract and include source code snippets in stack traces. This can be achieved by implementing a mechanism similar to the CodeExtractor class in the Bugsnag SDK, which reads the relevant files and attaches the surrounding lines of code to the stack trace.

1. Improved Debugging: Including source code snippets in stack traces allows developers to quickly understand the context of an error, making it easier to diagnose and fix issues.

2. Third-Party Code: When errors occur in third-party libraries, developers may not be familiar with the code. Providing the surrounding lines of code helps them understand what might be going wrong without needing to dive into the third-party codebase.

3. Efficiency: This feature saves time by reducing the need to manually locate and open files to see the relevant lines of code. It streamlines the debugging process and helps developers focus on solving the problem.

sl0thentr0py:
At some point, rails added a [default gem filter to their `BacktraceCleaner`](https://github.com/rails/rails/blob/a8709e6ea26eca73a652af4fdd0a9f7db5352af4/activesupport/lib/active_support/backtrace_cleaner.rb#L118-L125) which messed up our linecache/context lines parsing logic since we get paths like
```
activesupport (7.1.2) lib/active_support/callbacks.rb
```
instead of raw paths.

This PR cleans up handling this case by making a clear distinction between `abs_path` and `filename` as was intended in the original relay protocol and we were never populating these properly.

Fixes #2472

### DIFF
--- a/sentry-ruby/lib/sentry/backtrace.rb
+++ b/sentry-ruby/lib/sentry/backtrace.rb
@@ -18,6 +18,9 @@ module Sentry
       # org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:170)
       JAVA_INPUT_FORMAT = /^(.+)\.([^\.]+)\(([^\:]+)\:(\d+)\)$/
 
+      # The absolute path in the backtrace
+      attr_reader :abs_path
+
       # The file portion of the line (such as app/models/user.rb)
       attr_reader :file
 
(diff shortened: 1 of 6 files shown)


sl0thentr0py:
NOTE: since this affects the payload and grouping, I explicitly tested that the old and new payloads end up in the same issue group and it seems fine since we are only changing the underlying `abs_path` and simply making it more accurate.

sentry-ruby/lib/sentry/backtrace.rb
```diff
@@ -35,25 +38,33 @@ class Line
...
       # @return [Line] The parsed backtrace line
-      def self.parse(unparsed_line, in_app_pattern = nil)
-        ruby_match = unparsed_line.match(RUBY_INPUT_FORMAT)
-        if ruby_match
-          _, file, number, _, method = ruby_match.to_a
-          file.sub!(/\.class$/, RB_EXTENSION)
-          module_name = nil
-        else
+      def self.parse(unparsed_line, in_app_pattern = nil, cleaned_line = nil)
+        if RUBY_PLATFORM == "java"
```
sl0thentr0py, line 42: we only need to run this branch if we are in java land so we're saving some useless cycles now

sl0thentr0py:
testing 

all frames have context now

grafik


grafik


grouping is the same

grafik

st0012:
Please hold off merging this, I think the cause may be at other places and thus require a different change.

sl0thentr0py:
closed in favor of #2475

PR #2475, opened 2024-11-27
Fix sentry-rails' backtrace cleaner issues

### LINKED ISSUE #2472: Fix Source Code Extraction to Stack Traces (regression via rails)

st0012:
## Clear Rails' default backtrace filters from Sentry's backtrace cleaner

In Rails 7.2, Rails's backtrace cleaner, which sentry-rails' backtrace cleaner inherits from, starts shortening gem's paths in backtraces. This will prevent sentry-ruby's `Sentry::Backtrace` from handling them correctly, which will result in issues like #2472.

This commit avoids the issue by clearing the default filters that Sentry's backtrace cleaner inherits.

## Remove premature path-removal filter

This filter removes the project root part from the stacktrace, which prevents
sentry-ruby from retaining the correct absolute path of it.

Generally speaking, if stacktrace is messed up between Rails versions, but not Ruby versions, then it's usually caused by `Sentry::Rails::BacktraceCleaner` and/or the `backtrace_cleanup_callback` [`sentry-ruby` registers](https://github.com/getsentry/sentry-ruby/blob/a9b3687d3ce5d00d5586ef3fe76ae8233d8405f1/sentry-rails/lib/sentry/rails/railtie.rb#L114) (less likely). But if the backtrace is messed up between Ruby versions, it'd be on `sentry-ruby`'s backtrace handling instead.

Closes #2472

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@
 - Fix `RescuedExceptionInterceptor` to handle an empty configuration [#2428](https://github.com/getsentry/sentry-ruby/pull/2428)
 - Add mutex sync to `SessionFlusher` aggregates [#2469](https://github.com/getsentry/sentry-ruby/pull/2469)
   - Fixes [#2468](https://github.com/getsentry/sentry-ruby/issues/2468)
+- Fix sentry-rails' backtrace cleaner issues ([#2475](https://github.com/getsentry/sentry-ruby/pull/2475))
+  - Fixes [#2472](https://github.com/getsentry/sentry-ruby/issues/2472)
 
 ## 5.21.0
 
--- a/sentry-rails/lib/sentry/rails/backtrace_cleaner.rb
+++ b/sentry-rails/lib/sentry/rails/backtrace_cleaner.rb
@@ -11,8 +11,10 @@ module Sentry
 
       def initialize
         super
-        # we don't want any default silencers because they're too aggressive
+        # We don't want any default silencers because they're too aggressive
         remove_silencers!
+        # We don't want any default filters because Rails 7.2 starts shortening the paths. See #2472
+        remove_filters!
 
         @root = "#{Sentry.configuration.project_root}/"
         add_filter do |line|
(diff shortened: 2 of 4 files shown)


sl0thentr0py:
I'm fine with this change (after I test that grouping doesn't break this way either).

But we still need at least the `abs_path` change in https://github.com/getsentry/sentry-ruby/pull/2474 too because the current handling of `abs_path` after running backtrace cleaner is not really what we should be sending because it is sometimes already the stripped version. The whole path should be in `abs_path` and the processed/stripped path should be in `filename`.

sl0thentr0py (approved):

sl0thentr0py:
ok I tested grouping so this is fine to merge.

so the only problem now that I'd ideally like to fix (if possible in a simpler way that #2474) is that this block
https://github.com/getsentry/sentry-ruby/blob/50716c0ae26e23070f2eaee8e76c618c115cc4d0/sentry-rails/lib/sentry/rails/backtrace_cleaner.rb#L20-L22

does not preserve the `abs_path` for in_app frames as seen here
image

That logic is either way repeated here
https://github.com/getsentry/sentry-ruby/blob/50716c0ae26e23070f2eaee8e76c618c115cc4d0/sentry-ruby/lib/sentry/interfaces/stacktrace.rb#L50-L59

so are you ok with just removing that filter then?

st0012:
@sl0thentr0py ok I included that part of the fix too, and updated the PR description accordingly.

sl0thentr0py (approved):

PR #2487, opened 2024-12-07
Fix ActiveJob tests for Rails 7.2 and 8.0

olivier-thatch:
## Description

This PR fixes the test suite so that the ActiveJob related tests pass with Rails 7.2 and 8.0.

While working on the ActiveJob integration, I noticed that one of the tests was failing with Rails 7.2 and 8.0, specifically this one: https://github.com/getsentry/sentry-ruby/blob/1eb011b2ae31459bd3e2060dae7925e3b2629572/sentry-rails/spec/sentry/rails/activejob_spec.rb#L310-L324

After investigating, I realized that with Rails >= 7.2, all the test jobs were using `ActiveJob::QueueAdapters::AsyncAdapter` as their queue adapter instead of `ActiveJob::QueueAdapters::TestAdapter`. This is likely because of [this change](https://github.com/rails/rails/pull/48585) that was introduced in Rails 7.2.

The fix is easy enough: we simply need to make sure `ActiveJob::TestHelper` is included before the test suite runs. After including the helper module, all test jobs correctly use `ActiveJob::QueueAdapters::TestAdapter` as their queue adapter.

What I don't understand is why this happens locally but not in CI 🤔

#skip-changelog

### DIFF
--- a/sentry-rails/Gemfile
+++ b/sentry-rails/Gemfile
@@ -22,7 +22,7 @@ gem "rails", "~> #{rails_version}"
 
 if rails_version >= Gem::Version.new("8.0.0")
   gem "rspec-rails"
-  gem "sqlite3", platform: :ruby
+  gem "sqlite3", "~> 2.1.1", platform: :ruby
 elsif rails_version >= Gem::Version.new("7.1.0")
   gem "rspec-rails"
   gem "sqlite3", "~> 1.7.3", platform: :ruby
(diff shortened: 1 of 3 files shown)


solnic (approved):
Thanks! Could you please rebase this on top of latest master before we merge it?

olivier-thatch:
@solnic Done!

solnic:
> What I don't understand is why this happens locally but not in CI 🤔

@olivier-thatch on CI we run all the specs, and the spec fails locally only if you run it separately (at least for me). This means that there's some state that changes during whole spec suite run that causes this spec to pass, but it's different when we run it in isolation.

solnic (approved):
Thanks, this indeed fixes the problem :)

PR #2500, opened 2024-12-18
add report_after_job_retries support for ActiveJob

### LINKED ISSUE #2499: implement report_after_job_retries in ActiveJob adapter
we are using `sentry-rails` to report errors from `ActiveJob` (which is running on `sidekiq`). we want the ability to only report on errors when retries are exhausted. there doesn't seem to be a way to do this directly in the `ActiveJob` integration (`sentry-sidekiq`, `sentry-delayed_job`, and `sentry-resque` all support this configuration option).

is it possible to implement this directly in the `ActiveJob` adapter?

modosc:
## Description
this pr adds `report_after_job_retries` support to `ActiveJob` (configured via `config.rails.active_job_report_after_job_retries`). 

the easiest way to do this would have been to use [retry_on](https://edgeapi.rubyonrails.org/classes/ActiveJob/Exceptions/ClassMethods.html#method-i-retry_on) in the base `ActiveJob` class (similar to [here](https://github.com/getsentry/sentry-ruby/blob/master/sentry-raven/lib/raven/integrations/rails/active_job.rb#L9-L19)) but that's difficult to do properly which is (i think) why the current `ActiveJob` integration doesn't do it.

instead we can take advantage of the [`retry_stopped.active_job` notification](https://github.com/rails/rails/blob/035ff715b3db0c63d0990673671981fe1b9412b5/activejob/lib/active_job/exceptions.rb#L68-L77) that's published when a job runs out of retries. this makes our setup a single call to `ActiveSupport::Notifications.subscribe`. 

fixes #2499


sl0thentr0py:
@solnic could you review and get this merged if you have time?

solnic:
@sl0thentr0py looking into it today

modosc:
@solnic which version of `rails` / `ruby` are you testing against? in the automated tests i see seemingly random counts in the failures:

* [64](https://github.com/getsentry/sentry-ruby/actions/runs/12752068667/job/35540654670?pr=2500)
* [127](https://github.com/getsentry/sentry-ruby/actions/runs/12752068667/job/35540653392?pr=2500)
* [111](https://github.com/getsentry/sentry-ruby/actions/runs/12752068667/job/35540637748?pr=2500)
* [60](https://github.com/getsentry/sentry-ruby/actions/runs/12752068667/job/35540632146?pr=2500)
* [33](https://github.com/getsentry/sentry-ruby/actions/runs/12752068667/job/35540625001?pr=2500)

note that i only used `expect(Sentry::Rails::ActiveJobExtensions::SentryReporter)` because i got similar results using `expect(transport.events.size).to eq` - i just assumed i was doing something else wrong. 

it's also possible that the event publishing changed in between rails versions? if that's the case then i think we'd have to use the [other approach](https://github.com/getsentry/sentry-ruby/blob/master/sentry-raven/lib/raven/integrations/rails/active_job.rb#L9-L19) i mentioned. 

also fwiw i can confirm that subscribing to this event and sending an error to sentry does perform as expected in rails 8.0.1 / ruby 3.4.1 (we're using this now in production at my workplace) so it's quite possible the issue lies in the test tooling (maybe `assert_performed_jobs` runs in a separate thread or something)?

solnic:
> also fwiw i can confirm that subscribing to this event and sending an error to sentry does perform as expected in rails 8.0.1 / ruby 3.4.1 (we're using this now in production at my workplace) so it's quite possible the issue lies in the test tooling (maybe `assert_performed_jobs` runs in a separate thread or something)?

It fails when I run both scenarios, passes when I run individually:

```
     reports 3 exceptions
    when active_job_report_after_job_retries is true
      reports 1 exception (FAILED - 1)

Failures:

  1) ActiveJob integration active_job_report_after_job_retries when active_job_report_after_job_retries is true reports 1 exception
     Failure/Error:
       expect(Sentry::Rails::ActiveJobExtensions::SentryReporter)
         .to have_received(:capture_exception)
         .exactly(1).times
     
       (Sentry::Rails::ActiveJobExtensions::SentryReporter (class)).capture_exception(*(any args))
           expected: 1 time with any arguments
           received: 2 times with any arguments
     # ./spec/sentry/rails/activejob_spec.rb:425:in `block (4 levels) in <top (required)>'
     # /usr/local/rvm/gems/default/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:124:in `block in run'
     # /usr/local/rvm/gems/default/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:110:in `run'
     # /usr/local/rvm/gems/default/gems/rspec-retry-0.6.2/lib/rspec_ext/rspec_ext.rb:12:in `run_with_retry'
     # /usr/local/rvm/gems/default/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:37:in `block (2 levels) in setup'

Finished in 0.28003 seconds (files took 0.76855 seconds to load)
2 examples, 1 failure

Failed examples:

rspec ./spec/sentry/rails/activejob_spec.rb:420 # ActiveJob integration active_job_report_after_job_retries when active_job_report_after_job_retries is true reports 1 exception
```

This is latest rails and latest ruby.

solnic:
This is SO strange. When I run all tests from that file I get this:

```
       (Sentry::Rails::ActiveJobExtensions::SentryReporter (class)).capture_exception(*(any args))
           expected: 1 time with any arguments
           received: 23 times with any arguments
```

So it's like we get rspec's mocked calls count from all tests, and they just leak between the spec runs 🙈

solnic:
Closing this in favor of #2573 where I fixed issues with specs and handling of AS notification

Issue #1632, opened 2021-11-30
The number of attempts before notifications will be sent.


manme:
**Describe the idea**

I would like to have the option: `attempt_threshold` - the number of attempts before notifications will be sent. 
The option: report_after_job_retries - is a boolean and does not allow to send a notifications after N retry. 

**Why do you think it's beneficial to most of the users**

The user will receive a notification after the third retry because it could be solved automatically (external service was not available or similar), no more spam notifications. 

**Possible implementation**

Similar to https://github.com/honeybadger-io/honeybadger-ruby/blob/master/lib/honeybadger/plugins/sidekiq.rb#L43

st0012:
https://github.com/getsentry/sentry-ruby/issues/1570 looks pretty similar to what you propose here but with more detail. Would you say it matches your expectation?

manme:
Thanks for your response. As I understand, #1570 does not solve my problem: I want to get a notification in Sentry from a job that has failed N times (not on a first attempt) where N < sidekiq retry. For example, I get a notification after the third attempt, and the job will stay in Retries (with a default retry value of 25). Then as a developer, I have time to fix the bug and won't receive notification from jobs that failed because of network problems as it's not a bug, just a temporary issue.

st0012:
@manme I see. What'll happen after the first report then? In your example, the SDK should send the event after the 4th retry failed, right?. After that, should it send events for the 5th, 6th...etc. failures?

renchap:
I agree with the initial author, this would be very useful, especially for jobs depending on an external service and/or unreliable network, where it is not a problem if the job is retried for a few times.

I would expect all failures after the threshold to be sent to Sentry.

Sam-Killgallon:
:+1: I was just looking for this exact feature as well with the same use case. 

We integrate with a third party and a small percentage of requests fail and then succeed on the second retry due to things outside our control. I'd ideally like to be able to set a threshold on the `job` level, so that each job could be configured separately.

benhutton:
yes! I was about to submit a request for the exact same feature. I want the first one or two or N retries to NOT send anything to sentry, because that means that the error is a network error and is recovered quickly. I want all of the rest of the errors sent.

bekasovoleg:
👍🏻 I really need this feature

benhutton:
hey Sentry, Would y'all be open to a PR that implements this?

PR #2503, opened 2024-12-23
Use attempt_threshold to skip reporting on first N attempts

benhutton:
This is an implementation for #1632.

I am bringing over the concept of `attempt_threshold` from [Honeybadger](https://github.com/honeybadger-io/honeybadger-ruby/blob/master/lib/honeybadger/plugins/sidekiq.rb#L103).

Retry counting is a bit tricky, and so I included a comment [based on a GH comment from Honeybadger](https://github.com/honeybadger-io/honeybadger-ruby/pull/347/files#r390555981).

There are two relatively arbitrary decisions I made here that you should validate:
1. I am 1-indexing attempts, not 0-indexing
2. I am skipping until the number of attempts _equals_ the `attempt_threshold`. Alternatively, we could skip until number of attempts is _greater than_ `attempt_threshold`.

Thanks for your consideration!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Features
 
 - Improve the accuracy of duration calculations in cron jobs monitoring ([#2471](https://github.com/getsentry/sentry-ruby/pull/2471))
+- Use attempt_threshold to skip reporting on first N attempts ([#2503](https://github.com/getsentry/sentry-ruby/pull/2503))
 
 ### Bug fixes
 
--- a/sentry-sidekiq/lib/sentry/sidekiq/error_handler.rb
+++ b/sentry-sidekiq/lib/sentry/sidekiq/error_handler.rb
@@ -31,6 +31,19 @@ module Sentry
           end
         end
 
+        # Check if the retry count is below the attempt_threshold
+        attempt_threshold = context.dig(:job, "attempt_threshold")
+        if attempt_threshold && retryable?(context)
+          attempt_threshold = attempt_threshold.to_i
+          retry_count = context.dig(:job, "retry_count")
+          # attempt 1 - retry_count is nil
+          # attempt 2 - this is your first retry so retry_count is 0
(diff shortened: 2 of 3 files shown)


solnic (commented):
Thank you, this LGTM. Could you update the CHANGELOG.md (top level one) too so that we can get a green build?

benhutton:
@solnic sorry for the delay. Updated!

solnic (approved):

simoleone:
What's the intended interaction with `report_after_job_retries`? At the moment `report_after_job_retries` takes precedence and reports aren't sent until retries are exhausted regardless of `attempt_threshold`. 

I think it would be useful if when `report_after_job_retries: true` then `attempt_threshold` could be used to selectively lower the threshold for some jobs (so `attempt_threshold` would take precedence).

What do you think? It's just a matter of flipping the order of the conditionals.

benhutton:
personally I have no objection to that... I think my use case would be supported either way!

Linuus:
Any reasons this is not documented anywhere?

solnic:
> Any reasons this is not documented anywhere?

@Linuus good point, we should definitely document this and related features as it's kinda tricky 😅 I'll make sure we get it done!

Linuus:
> > Any reasons this is not documented anywhere?
> 
> @Linuus good point, we should definitely document this and related features as it's kinda tricky 😅 I'll make sure we get it done!

Thanks :) I was mostly curious if it was some beta/internal thing. Just to make sure we don't depend on something we shouldn't.

solnic:
@Linuus oh, sorry about the confusion. Let me clarify then - it is a first-class feature and you can rely on this 🙂

hibachrach:
One issue with this solution is it means jobs that are out of the developer's control (e.g. `Sidekiq::Batch::Callback`) cannot be configured with `sidekiq_options`. Instead, you need to set via some Sidekiq middleware.

PR #2506, opened 2025-01-01
Test Ruby 3.4; Upgrade actions workflows versions

frederikspang:
## Description
Describe your changes:

Upgrades tons of actions/checkout@v1 to v4, add 3.4 to build matrix.

Other questions: Ruby 2.4 is supported and tested, although it's EOL since 2020. Same goes for all versions through 3.0 actually, in yearly increments.

I'm all for keeping the test suite including some EOL versions, but 2.4 seems excessive.

### DIFF
--- a/.github/workflows/build_batch_release.yml
+++ b/.github/workflows/build_batch_release.yml
@@ -9,11 +9,11 @@ jobs:
     name: Build gems
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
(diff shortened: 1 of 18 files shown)


solnic (approved):
Thank you! We need to support ancient Rubies as long as Sentry customers use them 😅

frederikspang:
To be fair, as long as it's not an excessive work, it's fine!

We have a 2.7.x app running as well😄

PR #2507, opened 2025-01-02
Fix syntax in workflow files

solnic:
This fixes "unexpected end of the stream within a flow collection" error that we started seeing. Not sure what changed on GH though.

#skip-changelog

### DIFF
--- a/.github/workflows/sentry_delayed_job_test.yml
+++ b/.github/workflows/sentry_delayed_job_test.yml
@@ -34,14 +34,12 @@ jobs:
       matrix:
         ruby_version: ${{ fromJson(needs.ruby-versions.outputs.versions) }}
         include:
-          - {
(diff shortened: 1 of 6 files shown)


st0012 (approved):
While I do prefer this style better, I've never seen the said error before. Is there any example build that failed with it?

solnic:
> While I do prefer this style better, I've never seen the said error before. Is there any example build that failed with it?

So, here's a funny story - after I merged a PR that added 3.4 to the matrix last week, builds stopped working. I looked at the workflow yml in GH UI and it showed me the error, so I thought something changed on GH and it breaks now due to this syntax error...and then it just magically started working again.

We could merge this in regardless, it's simpler syntax after all.

sl0thentr0py (approved):
thx

PR #2509, opened 2025-01-13
Use RFC2396_PARSER explicitly to avoid the warning.

eGreshkov:

**Other Notes**

## Description
Describe your changes:

The URI library recently changed the `URI::DEFAULT_PARSER` to `URI::RFC3986_PARSER`.
https://github.com/ruby/uri/blob/master/lib/uri/common.rb#L22

Additionally, when using the `URI::RFC3986_PARSER.escape` method a warning is logged, telling that the method is obsolete and that `URI::RFC2396_PARSER.escape` should be used explicitly instead.
https://github.com/ruby/uri/blob/master/lib/uri/rfc3986_parser.rb#L155-L159

This results in a lot of warnings logged stderr. This simple fix should do the trick :)

### DIFF
--- a/sentry-ruby/lib/sentry/net/http.rb
+++ b/sentry-ruby/lib/sentry/net/http.rb
@@ -66,7 +66,7 @@ module Sentry
         # IPv6 url could look like '::1/path', and that won't parse without
         # wrapping it in square brackets.
         hostname = address =~ Resolv::IPv6::Regex ? "[#{address}]" : address
-        uri = req.uri || URI.parse(URI::DEFAULT_PARSER.escape("#{use_ssl? ? 'https' : 'http'}://#{hostname}#{req.path}"))
+        uri = req.uri || URI.parse(URI::RFC2396_PARSER.escape("#{use_ssl? ? 'https' : 'http'}://#{hostname}#{req.path}"))
         url = "#{uri.scheme}://#{uri.host}#{uri.path}" rescue uri.to_s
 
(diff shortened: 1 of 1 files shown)


solnic (commented):
Thank you for the PR! Please update to latest master though as we had some issues with the spec suite that should be fixed on master now.

solnic (changes requested):
Looks like in older rubies the constant does not exist:

```
uninitialized constant URI::RFC2396_PARSER
```

This is from 2.5 tests. There are more failing test suites with this error unfortunately, so we gotta have a fallback mechanism to make this work for old rubies too.

eGreshkov:
Thanks for checking the tests so quickly, great catch!

sentry-ruby/lib/sentry/net/http.rb
```diff
@@ -78,6 +78,12 @@ def extract_request_info(req)
 
         result
       end
+
+      def uri_parser
+        URI::RFC2396_PARSER
+      rescue NameError
+        URI::DEFAULT_PARSER
+      end
```
solnic, line 86: What do you think about defining our own constant conditionally instead of adding another method to the callstack? ie we could have this:

```ruby
URI_PARSER = URI.const_defined?(RFC2396_PARSER) ? URI::RFC2396_PARSER : URI::DEFAULT_PARSER
```

Then just use it in the extract_request_info `URI_PARSER.escape(...)`

solnic (approved):

PR #2510, opened 2025-01-16
Prevent crashes from get_main_hub when in trap context

### LINKED ISSUE #2495: ThreadError   resque:no_fork resque:work  New Level: Error Unhandled can't be called from trap context (ThreadError)
### Issue Description

When trying to upgrade our sentry-ruby version from 5.21 to the latest 5.22, I noticed the following error:
```
ThreadError resque:no_fork resque:work
Unhandled
can't be called from trap context (ThreadError)
```

The trace indicates the error is coming from sentry-ruby/lib/sentry-ruby.rb at line 310 from inside get_main_hub. 
[get_main_hub was indeed updated in the last release](https://github.com/getsentry/sentry-ruby/compare/5.21.0...5.22.0#diff-22b275750d1e90117f2c54a2b886b774f2eee7a8a96b2c3cd029b4ca182a12d5R310), but I can not say why is generating the error above.

### Reproduction Steps

I don't know what triggers it.

### Expected Behavior

The error mentioned above should not be present.

### Actual Behavior

The error mentioned above is present.

### Ruby Version

3.2.6

### SDK Version

5.21 - 5.22

Rails 7.0.8.5

### Sentry Config

```ruby
Exceptions.configure do |config|
  config.sentry.with(
    enabled: ENV.fetch('SENTRY_ENABLED', 'false') == 'true',
    dsn:  ENV.fetch('SENTRY_DSN', nil)
  )
  config.generate_fingerprint = false
end
```

solnic:
Fixes #2495

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - Default to `internal_error` error type for OpenTelemetry spans [#2473](https://github.com/getsentry/sentry-ruby/pull/2473)
 - Improve `before_send` and `before_send_transaction`'s return value handling ([#2504](https://github.com/getsentry/sentry-ruby/pull/2504))
+- Fix a crash when calling `Sentry.get_main_hub` in a trap context ([#2510](https://github.com/getsentry/sentry-ruby/pull/2510))
 
 ### Internal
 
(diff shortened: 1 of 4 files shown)


sl0thentr0py (approved):

solnic:
@sl0thentr0py I set the spec to run only under 3.4+ because it was randomly failing on older ones and jruby has no `fork`. This is such an edge case and hub's going away so it makes no sense to pay the price for running a complex repro test on all rubies.

PR #2504, opened 2024-12-29
Improve `before_send` and `before_send_transaction`'s return value handling

st0012:
The 1st commit fixes a testing setup issue and the 2nd commit adds additional checks to the `before_send` and `before_send_transaction`'s callback values. I explained the changes with more details in the commit messages.

Issue #2524, opened 2025-01-25
[rails] Documentation example on filtering needs to be updated : santize : extra, context, user, etc...


chrishough:
### Issue Description

This issue was closed https://github.com/getsentry/sentry-ruby/issues/1140#issuecomment-2614098758 but I posted the updated example there and would love to cut a PR to update the documentation, but I am not sure how you build that in the company...

```
# Sanitize sensitive data before sending
  config.before_send = lambda do |event, _hint|
    # Sanitize extra data
    if event.extra
      event.extra = filter.filter(event.extra)
    end

    # Sanitize user data
    if event.user
      event.user = filter.filter(event.user)
    end

    # Sanitize context data (if present)
    if event.contexts
      event.contexts = filter.filter(event.contexts)
    end

    # Return the sanitized event object
    event
  end

### Reproduction Steps

Described https://github.com/getsentry/sentry-ruby/issues/1140#issuecomment-2613462799

### Expected Behavior

<img width="575" alt="Image" src="https://github.com/user-attachments/assets/64c40472-614b-4946-a282-faebe1766850" />

### Actual Behavior

It will error out as notated in that post or it will not filter extra at all

### Ruby Version

3.4.1

### SDK Version

5.22.2

### Integration and Its Version

_No response_

### Sentry Config

_No response_

sl0thentr0py:
@st0012 same as #2526 
imo we should remove the `ActiveSupport::ParameterFilter` from the `before_send` docs because we don't really want to support a `Hash` event anymore.

chrishough:
great point @st0012 @sl0thentr0py

PR #2529, opened 2025-01-28
Accept Hash in `before_send*` callbacks again

### LINKED ISSUE #2526: Event not sent when scrubbing sensitive data in `before_send` on sentry-ruby 5.22.2
### Issue Description

- We have some tests checking our filtering before sending data to Sentry.
- When upgrading to 5.22.2 we no longer get any events in our tests

Our filtering matches the documentation
https://docs.sentry.io/platforms/ruby/guides/rails/data-management/sensitive-data/#scrubbing-data

I think this [change is the culprit](https://github.com/getsentry/sentry-ruby/compare/5.22.1...5.22.2#diff-6d22597786dbe5be19941aeb28ba2ba67b74c7a186642f902dc6c9e567319aeeR199).

### Reproduction Steps

config/initializers/sentry.rb
```ruby
SENTRY_FILTER = %i[
  identification_number
  ...
]

Sentry.init do |config|
  # https://docs.sentry.io/platforms/ruby/guides/rails/data-management/sensitive-data/#scrubbing-data
  filter = ActiveSupport::ParameterFilter.new(
    SENTRY_FILTER
  )

    filter.filter(event.to_hash)
  end
end

```

```ruby
require "test_helper"
require "sentry/test_helper"

class SentryTest < ActiveSupport::TestCase
  include Sentry::TestHelper

  setup do
    setup_sentry_test
  end

  test "exception sends to Sentry" do
    exception = RuntimeError.new("This is a test")
    Sentry.capture_exception(exception)

    event = last_sentry_event # this returns nil on 5.22.2
    message = event[:exception][:values][0][:value]

    assert_match(/This is a test/, message)
  end
end
```

### Expected Behavior

`before_send` filter should still be valid (or another way to write it)

### Actual Behavior

`before_send` returns hash and therefore it will not send

### Ruby Version

3.3.7

### SDK Version

5.22.2

Rails 7.1, Sidekiq 7.3.8

### Sentry Config

Minimal example that reproduces the error
```ruby
SENTRY_FILTER = %i[
  body
  identification_number
]

Sentry.init do |config|
  # https://docs.sentry.io/platforms/ruby/guides/rails/data-management/sensitive-data/#scrubbing-data
  filter = ActiveSupport::ParameterFilter.new(
    SENTRY_FILTER
  )

    filter.filter(event.to_hash)
  end
end

```

st0012:
It was unintentionally dropped in #2504. But because we do want to stop accepting Hash value in the future, I made it print deprecation messages when a Hash is passed as well.

Closes #2526

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Unreleased
+
+### Bug Fixes
+
+- Accept Hash in `before_send*` callbacks again ([#2529](https://github.com/getsentry/sentry-ruby/pull/2529))
+  - Fixes [#2526](https://github.com/getsentry/sentry-ruby/issues/2526)
+
 ## 5.22.2
 
 ### Features
@@ -6,7 +13,7 @@
 - Use attempt_threshold to skip reporting on first N attempts ([#2503](https://github.com/getsentry/sentry-ruby/pull/2503))
 - Support `code.namespace` for Ruby 3.4+ stacktraces ([#2506](https://github.com/getsentry/sentry-ruby/pull/2506))
 
-### Bug fixes
+### Bug Fixes
 
 - Default to `internal_error` error type for OpenTelemetry spans [#2473](https://github.com/getsentry/sentry-ruby/pull/2473)
 - Improve `before_send` and `before_send_transaction`'s return value handling ([#2504](https://github.com/getsentry/sentry-ruby/pull/2504))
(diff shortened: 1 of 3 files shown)


st0012:
https://github.com/getsentry/sentry-ruby/issues/2524 is what we'll use to track the documentation update and the author provided a good example to work with the event types for now.

sl0thentr0py (approved):

PR #2539, opened 2025-02-04
Deprecate config.metrics and Sentry::Metrics

sl0thentr0py:


st0012 (commented):
I don't have much context on this so this is an unblocking opinion: we probably should include this directly in the 6.0 release and move the release of it forward. Removing APIs' functionality technically is still breaking change so we'd still be breaking semver if we release it before 6.0.

sl0thentr0py:
> Removing APIs' functionality technically is still breaking change 

the server ignores these payloads completely anyway, so in that way it already 'broke'

solnic:
> but yeah, I also want to do the major soon but I do not have bandwidth to prepare it in the coming month or so

@sl0thentr0py happy to help with this

sl0thentr0py:
closing in favor of #2726

PR #2548, opened 2025-02-06
Add a spec checking range serialization in ActiveJob

### LINKED ISSUE #2545: ActiveJob TypeError: can't iterate from ActiveSupport::TimeWithZone
### Issue Description

Rails supports ActiveJob arguments to be a range of `ActiveSupport::TimeWithZone` objects, like `(5.days.ago...1.day.ago)`.

However, Sentry tries to iterate it, which is not allowed.

```
TypeError: can't iterate from ActiveSupport::TimeWithZone
    /Users/max/.asdf/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/activesupport-8.0.1/lib/active_support/core_ext/range/each.rb:19:in 'ActiveSupport::EachTimeWithZone#ensure_iteration_allowed'
    /Users/max/.asdf/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/activesupport-8.0.1/lib/active_support/core_ext/range/each.rb:8:in 'ActiveSupport::EachTimeWithZone#each'
    /Users/max/.asdf/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/sentry-rails-5.22.3/lib/sentry/rails/active_job.rb:85:in 'Enumerable#map'
    /Users/max/.asdf/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/sentry-rails-5.22.3/lib/sentry/rails/active_job.rb:85:in 'Sentry::Rails::ActiveJobExtensions::SentryReporter.sentry_serialize_arguments'
    /Users/max/.asdf/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/sentry-rails-5.22.3/lib/sentry/rails/active_job.rb:85:in 'block in Sentry::Rails::ActiveJobExtensions::SentryReporter.sentry_serialize_arguments'
    /Users/max/.asdf/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/sentry-rails-5.22.3/lib/sentry/rails/active_job.rb:85:in 'Array#map'
    /Users/max/.asdf/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/sentry-rails-5.22.3/lib/sentry/rails/active_job.rb:85:in 'Sentry::Rails::ActiveJobExtensions::SentryReporter.sentry_serialize_arguments'
    /Users/max/.asdf/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/sentry-rails-5.22.3/lib/sentry/rails/active_job.rb:72:in 'Sentry::Rails::ActiveJobExtensions::SentryReporter.sentry_context'
    /Users/max/.asdf/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/sentry-rails-5.22.3/lib/sentry/rails/active_job.rb:51:in 'block in Sentry::Rails::ActiveJobExtensions::SentryReporter.record'
    /Users/max/.asdf/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/sentry-ruby-5.22.3/lib/sentry/hub.rb:59:in 'Sentry::Hub#with_scope'
    /Users/max/.asdf/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/sentry-ruby-5.22.3/lib/sentry-ruby.rb:399:in 'Sentry.with_scope'
    /Users/max/.asdf/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/sentry-rails-5.22.3/lib/sentry/rails/active_job.rb:26:in 'Sentry::Rails::ActiveJobExtensions::SentryReporter.record'
    /Users/max/.asdf/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/sentry-rails-5.22.3/lib/sentry/rails/active_job.rb:10:in 'Sentry::Rails::ActiveJobExtensions#perform_now'
    /Users/max/.asdf/installs/ruby/3.4.1/lib/ruby/gems/3.4.0/gems/activejob-8.0.1/lib/active_job/execution.rb:23:in 'ActiveJob::Execution::ClassMethods#perform_now'
```

### Reproduction Steps

Run any ActiveJob passing a range like `5.days.ago...1.day.ago` as an argument.

### Expected Behavior

Sentry should be able to serialize a range of `ActiveSupport::TimeWithZone`.

### Actual Behavior

Sentry errors-out when trying to serialize a range of `ActiveSupport::TimeWithZone`.

### Ruby Version

3.4.1

### SDK Version

5.22.3

Rails 8.0.1

### Sentry Config

_No response_

solnic:
WIP

Closes #2545


sentry-rails/spec/sentry/rails/activejob_spec.rb
```diff
@@ -139,6 +139,49 @@ def post.to_global_id
...
+
+      event = transport.events.last.to_json_compatible
+      expect(event.dig("extra", "arguments")).to eq(
+        [
+          "foo",
+          { "bar" => "Sentry" },
+          {
+            "integer" => 1,
+            "post" => "gid://rails-test-app/Post/#{post.id}",
+            "range" => [1, 2, 3]
```
solnic, line 158: it's probably not a good idea to do this, but in a way this is public API so I didn't change it.

sentry-rails/lib/sentry/rails/active_job.rb
```diff
@@ -79,6 +79,12 @@ def sentry_context(job)
 
           def sentry_serialize_arguments(argument)
             case argument
+            when Range
+              if argument.first.is_a?(ActiveSupport::TimeWithZone)
```
maxim, line 83: This would still not support ranges of floats for example. (E.g. `(1.1...1.5)`) Is that intentional, just to keep this fix small?

maxim: @solnic Also, beware that `Range#first` raises an exception if it's a beginless range (..1).

```irb
(..1).first
'Range#first': cannot get the first element of beginless range (RangeError)
```

Unlike `begin`, which would return nil.

solnic: @maxim I just left original behavior untouched, but eventually I'd vote for changing it to always dump ranges to a string "{left}..{right}"

solnic: > Also, beware that `Range#first` raises an exception if it's a beginless range (..1).

@maxim ah great catch, thanks I addressed it via 6d8f774c

sentry-rails/lib/sentry/rails/active_job.rb
```diff
@@ -79,6 +79,12 @@ def sentry_context(job)
 
           def sentry_serialize_arguments(argument)
             case argument
+            when Range
+              if (argument.begin || argument.end).is_a?(ActiveSupport::TimeWithZone)
+                argument.to_s
+              else
+                argument.map { |v| sentry_serialize_arguments(v) }
```
st0012, line 86: Do we need to serialize individual components in the range? If the range is `(1..10000)` then it'd probably be pretty slow AND it'd allocate a new array with 10k elements in it.
Maybe in this case we just leave the `argument` until we find another element type that requires special treatments?

solnic: @st0012 we don't but it's the original behavior - is it OK to change it in a patch/bugfix release?

solnic: @st0012 yes exactly :)

st0012 (approved):

solnic:
This PR introduced a new challenge where we have specs that now have min Ruby version requirements due to syntax that's used (end-less and begin-less ranges in this case).

To solve this, I introduced a new rake task called `spec:versioned` that allows you to define `spec/versioned/x.y/**/*_spec.rb` specs and they will be run when ruby meets the min version requirements specified via directory name. See 5609d89f95073a72a71277a354a552c902be3260

@sl0thentr0py @st0012 are you OK with such a solution? One remaining bit is to figure out what to do with Rubocop that has Lint/Syntax and target ruby version set to 2.6, so now it fails.

sl0thentr0py:
let's just not use fancy syntax please then

solnic:
> let's just not use fancy syntax please then

oh, it's not that - we just need verify that such ranges work with activejob context param serialization 😅

sl0thentr0py:
I assume a simple `if RUBY_VERSION` guard around the spec doesn't work? If so, fine with this.

solnic:
> I assume a simple `if RUBY_VERSION` guard around the spec doesn't work? If so, fine with this.

hah yeah that was my instinct but then I realized that ruby parses entire file first, it won't skip anything due to conditionals so we were ending up with syntax errors.

Alright, gonna merge this in, FWIW I think it's gonna be useful for other stuff too.

PR #2549, opened 2025-02-06
Add correct breadcrumb levels for 4xx/5xx response codes

sl0thentr0py:
see https://github.com/getsentry/team-sdks/issues/100?reload=1?reload=1

### DIFF
--- a/sentry-ruby/lib/sentry/utils/http_tracing.rb
+++ b/sentry-ruby/lib/sentry/utils/http_tracing.rb
@@ -17,7 +17,7 @@ module Sentry
 
       def record_sentry_breadcrumb(request_info, response_status)
         crumb = Sentry::Breadcrumb.new(
-          level: :info,
+          level: get_level(response_status),
           category: self.class::BREADCRUMB_CATEGORY,
           type: "info",
           data: { status: response_status, **request_info }
@@ -54,6 +54,14 @@ module Sentry
           raise ArgumentError, "value must be a Hash" if prefix.nil?
           "#{URI.encode_www_form_component(prefix)}=#{URI.encode_www_form_component(value)}"
         end
+      end
+
+      private
+
+      def get_level(status)
+        return :info unless status && status.is_a?(Integer)
+
+        status >= 500 ? :error : (status >= 400 ? :warning : :info)
       end
     end
   end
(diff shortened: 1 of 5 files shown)


sentry-ruby/lib/sentry/utils/http_tracing.rb
```diff
@@ -55,6 +55,14 @@ def build_nested_query(value, prefix = nil)
           "#{URI.encode_www_form_component(prefix)}=#{URI.encode_www_form_component(value)}"
         end
       end
+
+      private
+
+      def get_level(status)
+        return :info unless status && status.is_a?(Integer)
+
+        status >= 500 ? :error : (status >= 400 ? :warning : :info)
```
solnic, line 64:
```suggestion
if status >= 500
  :error
elsif status >= 400
  :warning
else
  :info
end
```

nit but 🙏🏻

sentry-ruby/lib/sentry/utils/http_tracing.rb
```diff
@@ -55,6 +55,14 @@ def build_nested_query(value, prefix = nil)
           "#{URI.encode_www_form_component(prefix)}=#{URI.encode_www_form_component(value)}"
         end
       end
+
+      private
+
+      def get_level(status)
+        return :info unless status && status.is_a?(Integer)
```
solnic, line 62: Is it really OK to default to :info when we don't recognize the status? Shouldn't this `log_warn("Unexpected HTTP status for breadcrumb: #{status.inspect}")`?

sentry-ruby/spec/sentry/breadcrumb/http_logger_spec.rb
```diff
@@ -56,6 +56,17 @@
         { status: 200, method: "POST", url: "http://example.com/path", query: "foo=bar", body: 'quz=qux' }
       )
     end
+
+    { 200 => :info, 400 => :warning, 500 => :error }.each do |status, level|
```
solnic, line 60:
```suggestion
    { 200 => :info, 302 => :info, 400 => :warning, 404 => :warning, 500 => :error, 502 => error }.each do |status, level|
```

solnic (approved):
Left some minor suggestions, otherwise LGTM

sentry-ruby/lib/sentry/utils/http_tracing.rb
sl0thentr0py: yes
https://github.com/getsentry/relay/blob/bc70b9e3401bccb69e362acf16b9e6bf630aa3a2/relay-event-schema/src/protocol/breadcrumb.rs#L90-L95

PR #2553, opened 2025-02-07
Update dependabot config

solnic:
#skip-changelog

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,74 +1,14 @@
 version: 2
-
 updates:
-  - directory: "sentry-delayed_job"
-    package-ecosystem: "bundler"
+  - package-ecosystem: "bundler"
     schedule:
       interval: "daily"
     allow:
       - dependency-type: "direct"
-    ignore:
-      - dependency-type: "all"
-        update-types: ["version-update:semver-patch"]
-      - patterns:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,74 +1,14 @@
...
-        update-types: ["version-update:semver-patch"]
-      - patterns:
-        - "**/examples/**"
+    directories:
+      - "sentry-delayed_job"
+      - "sentry-opentelemetry"
+      - "sentry-rails"
+      - "sentry-ruby"
+      - "sentry-resque"
+      - "sentry-sidekiq"
```
sl0thentr0py, line 14: what about the ignores?

solnic: @sl0thentr0py turned out it doesn't work like that :( you can't specify dirs/patterns to ignore

sl0thentr0py: then we don't need a config at all, just remove it again

solnic: @sl0thentr0py no, it will help because it will at least not bother us with npm deps. The config tells dependabot to only consider direct gem dependencies from the provided directories.

sl0thentr0py: ah right, thx

Issue #2522, opened 2025-01-24
Fix flaky specs


solnic:
### Issue Description

This issue gathers random failures we get from some flaky tests.

## Failure 01

- Ruby 3.3

```
Failures:

  1) Sentry::Client integrated error handling testing with HTTPTransport #capture_event when sending events in background causes error captures client report for transaction event with span counts
     Failure/Error: expect(client.transport).to have_recorded_lost_event(:network_error, 'transaction')
       expected #<Sentry::HTTPTransport:0x00007fad0197b5e8 @logger=#<Logger:0x00007fad0197b868 @level=0, @progname=ni...nt_reports=true, @discarded_events={}, @last_client_report_sent=2025-01-24 10:01:40.480909663 +0000> to have recorded lost event :network_error and "transaction"
     # ./spec/sentry/client/event_sending_spec.rb:492:in `block (5 levels) in <top (required)>'
     # ./spec/sentry/client/event_sending_spec.rb:359:in `block (4 levels) in <top (required)>'
     # /home/runner/work/sentry-ruby/sentry-ruby/vendor/bundle/ruby/3.3.0/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:124:in `block in run'
     # /home/runner/work/sentry-ruby/sentry-ruby/vendor/bundle/ruby/3.3.0/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:110:in `run'
     # /home/runner/work/sentry-ruby/sentry-ruby/vendor/bundle/ruby/3.3.0/gems/rspec-retry-0.6.2/lib/rspec_ext/rspec_ext.rb:12:in `run_with_retry'
     # /home/runner/work/sentry-ruby/sentry-ruby/vendor/bundle/ruby/3.3.0/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:37:in `block (2 levels) in setup'

  2) Sentry::Client integrated error handling testing with HTTPTransport #capture_event when sending events in background causes error captures client report for error event
     Failure/Error: expect(client.transport).to have_recorded_lost_event(:network_error, 'error')
       expected #<Sentry::HTTPTransport:0x00007fad00bd46b0 @logger=#<Logger:0x00007fad00bd48e0 @level=0, @progname=ni...nt_reports=true, @discarded_events={}, @last_client_report_sent=2025-01-24 10:01:41.297596915 +0000> to have recorded lost event :network_error and "error"
[...]

PR #2561, opened 2025-02-14
Fix flaky specs

solnic:
Our spec suite has become very unstable. I've managed to track it down to issues related to specs that set up HTTPTransport and then having request stubbing not really working. This PR adds Webmock that guarantees that we don't send HTTP requests unintentionally and keeps using our customized Net::HTTP stubbing context untouched and used the same way in specs where it is needed.

One *especially* problematic crash was a transport failure *after* spec suite run. I tracked it down to one specific spec - it lacked some configuration so I moved it to the context where that configuration is present, and the error went away, see 0c1476bdd7b63af30194afc4cc753aff360d85f7.

It fixes *most* of the problems with flakiness, issue #2522 still lists some random failures that should be addressed so I'm keeping it open.

I think it's worth merging this in already.

Refs #2522 

#skip-changelog


sentry-ruby/spec/sentry/transport_spec.rb
```diff
@@ -531,11 +531,14 @@
         end
       end
 
-      context "with Faraday::Error" do
+      context "with an HTTP exception" do
+        before do
+          stub_request(:post, "http://sentry.localdomain:80/sentry/api/42/envelope/").
+          to_raise(Timeout::Error)
+        end
```
solnic, line 538: I do not know what the intention of this spec was originally since Faraday was NOT configured here and the spec was passing by an accident since there was a socket timeout raised under the hood.

I can add a faraday spec too. Just lemme know if it'd make sense.

sl0thentr0py (approved):
merci!

sentry-ruby/spec/sentry/transport_spec.rb
sl0thentr0py: its a leftover from the legacy Raven SDK, ignore
https://github.com/getsentry/sentry-ruby/blob/cc1dc53c0d60a21db7ad95cac05590ec23c82910/sentry-raven/lib/raven/transports/http.rb#L28

st0012: Yes we may just remove the entire test case.

sentry-ruby/spec/contexts/with_request_mock.rb
```diff
@@ -12,7 +12,13 @@ def setsockopt(*args); end
...
 
-  def stub_request(fake_response, &block)
+  around do |example|
+    WebMock.disable!
+    example.run
+    WebMock.enable!
+  end
+
+  def sentry_stub_request(fake_response, &block)
     allow_any_instance_of(Net::HTTP).to receive(:connect)
```
st0012, line 22: Do we still need these stubs with webmock?

sentry-ruby/spec/spec_helper.rb
```diff
@@ -69,6 +70,12 @@
     skip("Skipping because one or more guards `#{guards.inspect}` returned false") if skip_examples
   end
 
+  config.around(:each, webmock: false) do |example|
+    WebMock.disable!
+    example.run
+    WebMock.enable!
+  end
```
st0012, line 77: Should this live in `sentry-ruby/spec/contexts/with_request_mock.rb` too?

sentry-ruby/spec/sentry/transport_spec.rb
solnic: I'm gonna leave it. It does verify specific error handling - HTTPTransport defines a list of exceptions that are rescued in send_data and now this spec checks that it works as expected that the exception is wrapped in Sentry::ExternalError.

st0012 (approved):

sentry-ruby/spec/contexts/with_request_mock.rb
solnic: @st0012 this is plain RSpec mocking, not WebMock

sentry-ruby/spec/spec_helper.rb
solnic: @st0012 no because this context is cherry-picked only by a handful of specs and we want to be able to easily toggle webmock on or off on a per-spec basis.

I do want to say that the whole setup could be done nicer, but it's a nice to have at this point I think :) I can get back to it on another occasion.

sentry-ruby/spec/contexts/with_request_mock.rb
st0012: Ah we're not replacing the current stubbing completely. Sorry that I didn't go through the descriptions thoroughly:

> and keeps using our customized Net::HTTP stubbing context untouched and used the same way in specs where it is needed.

sentry-ruby/spec/spec_helper.rb
st0012: Yeah that's not a problem I'm just curious. And sorry that it's mostly the mess I left 😬 
I'll take a look at the setup this week as well. Thanks for fixing it.

PR #2564, opened 2025-02-19
TypeError when payload class is already a constant

jgarber-cisco:
## Description
When `Resque.inline = true`, which is often used in an application's specs when testing a Resque job's business logic, the `payload["class"]` that the Sentry hook gets is an actual class, not a string! I've added a spec for this case and fixed it by avoiding constantization if it's already a class.

Note: Reliance on [ActiveSupport::Inflector#constantize](https://github.com/rails/rails/blob/cf6ff17e9a3c6c1139040b519a341f55f0be16cf/activesupport/lib/active_support/inflector/methods.rb#L289) can be removed if/when `sentry-ruby` no longer supports Ruby 2.4-2.6. According to [this Rails commit](https://github.com/rails/rails/commit/7057ccf6565c1cb5354c1906880119276a9d15c0), `Object.const_get` works properly from Rails 2.7 on.

### DIFF
--- a/sentry-resque/lib/sentry/resque.rb
+++ b/sentry-resque/lib/sentry/resque.rb
@@ -40,8 +40,11 @@ module Sentry
 
               finish_transaction(transaction, 200)
             rescue Exception => exception
-              klass = payload["class"]
-              klass = Object.const_get(klass) unless klass.is_a?(Class)
+              klass = if payload["class"].respond_to?(:constantize)
+                payload["class"].constantize
+              else
+                Object.const_get(payload["class"].to_s)
+              end
 
               raise if Sentry.configuration.resque.report_after_job_retries &&
                        defined?(::Resque::Plugins::Retry) == "constant" &&
--- a/sentry-resque/lib/sentry/resque.rb
+++ b/sentry-resque/lib/sentry/resque.rb
@@ -40,16 +40,7 @@ module Sentry
 
               finish_transaction(transaction, 200)
             rescue Exception => exception
-              klass = if payload["class"].respond_to?(:constantize)
-                payload["class"].constantize
-              else
-                Object.const_get(payload["class"].to_s)
-              end
-
-              raise if Sentry.configuration.resque.report_after_job_retries &&
-                       defined?(::Resque::Plugins::Retry) == "constant" &&
-                       klass.is_a?(::Resque::Plugins::Retry) &&
-                       !klass.retry_limit_reached?
+              raise_if_retries_remain(payload["class"], exception)
 
               ::Sentry::Resque.capture_exception(exception, hint: { background: false })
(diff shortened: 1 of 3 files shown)


sentry-resque/spec/sentry/resque_spec.rb
```diff
@@ -208,6 +208,25 @@ def self.perform(msg)
...
+        expect do
+          Resque::Job.create(:default, FailedRetriableJob)
+          process_job(worker)
+        end.not_to raise_error(TypeError)
+
+        event = transport.events.last.to_hash
+        expect(event.dig(:exception, :values, 0, :type)).to eq("ZeroDivisionError")
+      end
+    end
+
```
solnic, line 229: redundant empty line:

```
sentry-resque/spec/sentry/resque_spec.rb:229:1: C: [Correctable] Layout/EmptyLinesAroundBlockBody: Extra empty line detected at block body end.
```

sentry-resque/lib/sentry/resque.rb
```diff
@@ -43,7 +43,7 @@ def record(queue, worker, payload, &block)
               klass = if payload["class"].respond_to?(:constantize)
                 payload["class"].constantize
               else
-                Object.const_get(payload["class"])
+                Object.const_get(payload["class"].to_s)
```
solnic, line 46: It would be faster and safer to check the type:
```suggestion
                (klass = payload["class"]).is_a?(Class) ? klass : Object.const_get(klass)
```

solnic (changes requested):
Thank you! I've left a couple of comments 🙂

sentry-resque/lib/sentry/resque.rb
jgarber-cisco: Now checking type. Also removed `constantize` logic because it was a lot of gymnastics to basically use an ActiveSupport alias for `Object.const_get`.

solnic (changes requested):

sentry-resque/lib/sentry/resque.rb
```diff
@@ -40,11 +40,8 @@ def record(queue, worker, payload, &block)
 
               finish_transaction(transaction, 200)
             rescue Exception => exception
-              klass = if payload["class"].respond_to?(:constantize)
-                payload["class"].constantize
-              else
-                Object.const_get(payload["class"])
-              end
```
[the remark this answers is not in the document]
jgarber-cisco: You're right, [it used to be more complex before Rails 7.0](https://github.com/rails/rails/commit/7057ccf6565c1cb5354c1906880119276a9d15c0) to support Ruby <2.7. Since Sentry supposedly supports Ruby >= 2.4, I'll keep `constantize` and add a code comment.

solnic (approved):
TY :)

PR #2565, opened 2025-02-24
Flaky spec fixes

solnic:
I reckon this fixes the most problematic flaky specs that we have now. I could not reproduce the jruby failures locally no matter what I did, so these are still pending. Given that the build passed 3 times in a row here, I feel like this is already very beneficial to merge in.

Refs #2522 

#skip-changelog


sentry-ruby/Gemfile
```diff
@@ -29,7 +29,7 @@ gem "benchmark_driver"
 gem "benchmark-ipsa"
 gem "benchmark-memory"
 
-gem "yard", github: "lsegal/yard"
+gem "yard", github: "lsegal/yard", branch: "main"
```
st0012, line 32: I used yard's GH source because we were waiting for a fix (https://github.com/getsentry/sentry-ruby/pull/1994), which has been released now. So I think we can do this instead:

```suggestion
gem "yard"
```

sentry-ruby/lib/sentry/test_helper.rb
```diff
@@ -83,5 +83,18 @@ def last_sentry_event
     def extract_sentry_exceptions(event)
       event&.exception&.values || []
     end
+
+    def reset_sentry_globals!
+      # Don't check initialized? because sometimes we stub it in tests
+      if Sentry.instance_variable_defined?(:@main_hub)
```
st0012, line 89: Should we move this into mutex as well?

st0012 (approved):

PR #2568, opened 2025-02-27
Multi ruby devcontainer setup

solnic:
I struggled a lot when switching between rubies to test things out / fix specs etc. so I ended up with a devcontainer setup that allows me to easily switch to a specific ruby version in VS Code.

I figured I could share this here because it may turn out to be useful for others.

The reason why this is based on devcontainers is simple - we support too many rubies that have conflicting dependencies. I could not use rvm or any other Ruby version manager to easily switch between super old crubies or jruby.

Furthermore, I think we could have a nice setup with relay running as a docker service which can *really help* with figuring things out with envelopes etc.

#skip-changelog

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,6 +22,18 @@ And if you have any questions, please feel free to reach out on [Discord].
 If you use editors that support [VS Code-style multi-root workspaces](https://code.visualstudio.com/docs/editor/multi-root-workspaces),
 such as VS Code, Cursor...etc., opening the editor with `sentry-ruby.code-workspace` file will provide a better development experience.
 
+## Working in a devcontainer
+
+If you use [Dev Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) extension, you can open the project with the devcontainer by running `Remote-Containers: Reopen in Container` command.
+
+The devcontainer is configured with `.devcontainer/.env` file, that you need to create:
+
+```bash
+cp .devcontainer/.env.example .devcontainer/.env
+```
+
+This file defines which specific image and Ruby version will be used to run the code. Edit it whenever you need to use a different image or Ruby version.
+
 ## Contribute To Individual Gems
 
 - Install the dependencies of a specific gem by running `bundle` in it's subdirectory. I.e:
--- a/.devcontainer/.env.example
+++ b/.devcontainer/.env.example
@@ -1,0 +1,7 @@
+IMAGE=bitnami/ruby
+
+# Adjust as needed
+TAG=3.4
+
+# IMAGE=jruby
+# TAG=latest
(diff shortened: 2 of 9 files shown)


st0012 (commented):
I think this is an interesting idea. Can you also describe the process you use to switch between Rubies with this setup? I think it's probably worth documented in `CONTRIBUTING.md`?

solnic:
> I think this is an interesting idea. Can you also describe the process you use to switch between Rubies with this setup? I think it's probably worth documented in `CONTRIBUTING.md`?

@st0012 I updated the CONTRIBUTING file - would you mind checking out this branch and seeing if you can get up and running with the devcontainer?

CONTRIBUTING.md
```diff
@@ -22,6 +22,18 @@ And if you have any questions, please feel free to reach out on [Discord].
...
 such as VS Code, Cursor...etc., opening the editor with `sentry-ruby.code-workspace` file will provide a better development experience.
 
+## Working in a devcontainer
+
+If you use [Dev Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) extension, you can open the project with the devcontainer by running `Remote-Containers: Reopen in Container` command.
+
+The devcontainer is configured with `.devcontainer/.env` file, that you need to create:
+
+```bash
+cp .devcontainer/.env.example .devcontainer/.env
```
sl0thentr0py, line 32: do we need to commit this `.env.example` file?

sl0thentr0py (approved):
I don't use vscode but totally okay with merging in a better tooling experience.
I'll also try to see if I can use these without vscode.

CONTRIBUTING.md
solnic: @sl0thentr0py ohhh good catch TY :) just added it

st0012:
Thanks for the instructions. Generally speaking it works but I have a few questions as a devcontainer noob:

1. You mentioned that the VS Code command is called `Remote-Containers: Reopen in Container`, but I did find it and ended up using this instead. I wonder what's causing the difference 🤔 
      Screenshot 2025-03-08 at 22 55 35
2. How did you pull your local shell configs...etc. into the container? I currently don't see a way to do it and it makes the terminal in the container hard to use 😢 
3. I noticed that the top-level folder is `sentry` instead of `sentry-ruby`, which fails some tests (see below). Do you want to update the config to use `sentry-ruby` as the root instead?
    https://github.com/getsentry/sentry-ruby/blob/f11a7aa64b02305ca01cb6cb6ad3ce920360f3ef/sentry-ruby/spec/sentry/profiler_spec.rb#L244
4. `Ruby LSP` doesn't run with the current setting as the container doesn't use any Ruby version manager and the devcontainer setting doesn't specify sufficient settings as an alternative. How do you make it work?

PR #2572, opened 2025-03-04
Specify versions for dev deps

solnic:
#skip-changelog


solnic:
For the record, this was supposed to address dependabot alerts but the new versions of simplecov and friends require Ruby 2.5+.

PR #2574, opened 2025-03-06
Stop byteslicing empty strings in breadcrumbs

nirebu:

**Other Notes**

## Description
Describe your changes:

If we don't pass a message to the breadcrumb message, we default to an empty string. However we can avoid byteslicing it, which allocates a new empty string every time regardless.

This is a pretty hot path, and the tests pass. Only difference I can think of is returning a frozen empty string, which could cause issues downstream. Let me know if it should be dup-ed!

```rb

require 'benchmark-memory'

MAX_MESSAGE_SIZE_IN_BYTES = 1024 * 8

def current(message)
  (message || "").byteslice(0..MAX_MESSAGE_SIZE_IN_BYTES)
end

def updated(message)
  message ? message.byteslice(0..MAX_MESSAGE_SIZE_IN_BYTES) : ""
end

Benchmark.memory do |x|
  x.report("current") do
    current(nil)
  end

  x.report("updated") do
    updated(nil)
  end

  x.compare!
end
```

Calculating -------------------------------------
             current    80.000  memsize (     0.000  retained)
                         2.000  objects (     0.000  retained)
                         1.000  strings (     0.000  retained)
             updated     0.000  memsize (     0.000  retained)
                         0.000  objects (     0.000  retained)
                         0.000  strings (     0.000  retained)

Comparison:
             updated:          0 allocated
             current:         80 allocated - Infx more

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 - Use `File.open` in `LineCache` ([#2566](https://github.com/getsentry/sentry-ruby/pull/2566))
 - Update java backtrace regexp ([#2567](https://github.com/getsentry/sentry-ruby/pull/2567))
+- Stop byteslicing empty strings in breadcrumbs ([#2574](https://github.com/getsentry/sentry-ruby/pull/2574))
 
 ### Miscellaneous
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- Stop byteslicing empty strings in breadcrumbs ([#2574](https://github.com/getsentry/sentry-ruby/pull/2574))
+
 ### Features
```
solnic, line 5: Could you please move this under `### Internal` section? 🙏🏻

Issue #2397, opened 2024-09-06
SystemStackError when serializing breadcrumbs.  Potentially causing serious issues in Sidekiq when uses in conjunction with background workers


drj17:
### Issue Description

[Discord thread](https://discord.com/channels/621778831602221064/1278773125881794701)

Continuing this report from a discord thread I created a few days ago.   We have been investigating our Sidekiq workers becoming unresponsive and think we traced the issue to a combination of two things.

First, a SystemStackError when serializing the breadcrumbs - here's some relevant logging
```
17653784398360371212024-08-28 09:10:412024-08-28 09:10:4113994936442blaze-ai-rails34.225.253.144Local7Infoapp/almanacworker.3/app/vendor/bundle/ruby/3.1.0/gems/activesupport-7.0.5/lib/active_support/core_ext/object/instance_variables.rb:15: warning: Exception in finalizer #<Proc:0x00007fe503ce5a28 (lambda)>
17653784398360371222024-08-28 09:10:412024-08-28 09:10:4113994936442blaze-ai-rails34.225.253.144Local7Infoapp/almanacworker.3/app/vendor/bundle/ruby/3.1.0/gems/activesupport-7.0.5/lib/active_support/core_ext/object/instance_variables.rb:15:in `[]': SystemStackError
17653784398360371232024-08-28 09:10:412024-08-28 09:10:4113994936442blaze-ai-rails34.225.253.144Local7Infoapp/almanacworker.3\tfrom /app/vendor/bundle/ruby/3.1.0/gems/activesupport-7.0.5/lib/active_support/core_ext/object/instance_variables.rb:15:in `instance_values'
17653784398360371242024-08-28 09:10:412024-08-28 09:10:4113994936442blaze-ai-rails34.225.253.144Local7Infoapp/almanacworker.3\tfrom /app/vendor/bundle/ruby/3.1.0/gems/activesupport-7.0.5/lib/active_support/core_ext/object/json.rb:63:in `as_json'
17653784398360371252024-08-28 09:10:412024-08-28 09:10:4113994936442blaze-ai-rails34.225.253.144Local7Infoapp/almanacworker.3\tfrom /app/vendor/bundle/ruby/3.1.0/gems/activesupport-7.0.5/lib/active_support/core_ext/object/json.rb:180:in `block in as_json'
[...]

PR #2583, opened 2025-03-17
Add max_nesting of 10 to breadcrumbs data serialization

### LINKED ISSUE #2393: Restrict breadcrumb JSON generate/parse to depth 5
Use `max_nesting`.
Relay already trims this so this data is useless anyway and sometimes causes recursion errors.
https://github.com/getsentry/relay/blob/bcbfa7e7db6e4a8f3f0383270fbb2b5cfe668770/relay-event-schema/src/protocol/breadcrumb.rs#L111

sl0thentr0py:
This is [trimmed in relay](https://github.com/getsentry/relay/blob/bcbfa7e7db6e4a8f3f0383270fbb2b5cfe668770/relay-event-schema/src/protocol/breadcrumb.rs#L111) at a depth of 5.
We allow upto 10 to have some leeway but if we have a really large data object we just drop it to avoid `SystemStackError`.

Closes #2393 and #2397

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Features
 
 - Add new sidekiq config `report_only_dead_jobs` ([#2581](https://github.com/getsentry/sentry-ruby/pull/2581))
+- Add `max_nesting` of 10 to breadcrumbs data serialization ([#2583](https://github.com/getsentry/sentry-ruby/pull/2583))
 
 ### Bug Fixes
 
--- a/sentry-ruby/lib/sentry/breadcrumb.rb
+++ b/sentry-ruby/lib/sentry/breadcrumb.rb
@@ -2,6 +2,7 @@
 
 module Sentry
   class Breadcrumb
+    MAX_NESTING = 10
     DATA_SERIALIZATION_ERROR_MESSAGE = "[data were removed due to serialization issues]"
 
     # @return [String, nil]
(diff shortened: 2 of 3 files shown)


szokeasaurusrex (commented):
If I understand correctly, this change causes the entire breadcrumb data to be deleted if the nesting is at least 5. Why not just strip the portions which are too deeply nested? 

I suppose if Relay is throwing away the entire data when there is too much nesting, this is okay, but if relay only removes the part that is too deeply nested, the SDK should have the same behavior.

sl0thentr0py:
valid, but that will require generalized databag trimming which is not implemented in ruby.
getting this in is more valuable since we're doing arbitrary recursion, we can revisit this separately.

sl0thentr0py:
alright, let's do this, I'll make the depth 10, so we allow some leeway instead of dropping the whole thing.

szokeasaurusrex (approved):
Nice! So if I understand correctly, the limit of 10 is so that if the depth of 5 is only slightly exceeded, we allow Relay's logic to trim down to a depth of 5 (preserving the portion that is less than 5 deep), but if we exceed the limit of 10, we drop the whole thing to avoid the error in Ruby?

I think that is reasonable. Although, if we can avoid the recursion error altogether, that would be the best solution (i.e. by replacing the recursive function which causes the error with an iterative one). Would this be feasible?

sl0thentr0py:
yes with 10 we will drop the whole thing
we will likely not replicate the python serializer mess here so there is no way to avoid this behavior for now, which is fine.

Issue #2476, opened 2024-11-27
Allow controlling trace injection behaviour in queue gems


sl0thentr0py:
> @Linuus doing this in the clockwork process **only** should work as a workaround if you need it urgently.
> 
> ```ruby
> Sidekiq.configure_client do |config|
>   config.client_middleware do |chain|
>     chain.remove Sentry::Sidekiq::SentryContextClientMiddleware
>   end
> end
> ```
> 
> This will remove the trace header injection but only in the clockwork process. Make sure you don't put it in a general configuration.
> 
> For the longer term, I think I will add a configuration parameter to the sidekiq gem which lets you control the trace header injection behaviour. 

 _Originally posted by @sl0thentr0py in [#2391](https://github.com/getsentry/sentry-ruby/issues/2391#issuecomment-2504004985)_

cobyeastwood183:
https://sentry.zendesk.com/agent/tickets/163096

PR #2588, opened 2025-03-19
Add sidekiq config `propagate_traces` to control trace header injection

### LINKED ISSUE #2531: Unrelated Sidekiq job runs being grouped in same trace
### Issue Description

I'm seeing across multiple unrelated apps/projects this behavior where Sidekiq jobs are being grouped into the same parent trace, even though the job runs are completely unrelated:

Project A:

Image

Project B:

Image

You end up getting traces that are _days_ long:

Image

I think this really messes with sampling, as now I can end up with massive traces from Sidekiq using tons and tons of spans, which means I have to decrease my sample rate overall to compensate, meaning I get less spans from web requests. 

### Reproduction Steps

Unsure at this point. I don't know how this behavior works in Sentry so I don't know how to repro.

I think this only happens when a Sidekiq job is run and a web request was _not_ the parent (i.e. a job triggered by a cron). This

### Expected Behavior

Unrelated job runs should not be grouped into the same transaction.

### Actual Behavior

They are.

### Ruby Version

Multiple

### SDK Version

Multiple, but includes `5.22.1`.

_No response_

### Sentry Config

Multiple, no common thread in the config AFAICT

sl0thentr0py:
closes #2531
part of #2476

### DIFF
--- a/sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
+++ b/sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
@@ -95,7 +95,10 @@ module Sentry
 
         user = Sentry.get_current_scope.user
         job["sentry_user"] = user unless user.empty?
-        job["trace_propagation_headers"] ||= Sentry.get_trace_propagation_headers
+
+        if Sentry.configuration.sidekiq.propagate_traces
+          job["trace_propagation_headers"] ||= Sentry.get_trace_propagation_headers
+        end
 
         Sentry.with_child_span(op: "queue.publish", description: worker_class.to_s) do |span|
           set_span_data(span, id: job["jid"], queue: queue)
--- a/sentry-sidekiq/spec/sentry/sidekiq/configuration_spec.rb
+++ b/sentry-sidekiq/spec/sentry/sidekiq/configuration_spec.rb
@@ -26,4 +26,16 @@ RSpec.describe Sentry::Sidekiq::Configuration do
       expect(subject.report_after_job_retries).to eq(false)
     end
   end
+
+  describe "#report_only_dead_jobs" do
+    it "has correct default value" do
+      expect(subject.report_only_dead_jobs).to eq(false)
+    end
+  end
+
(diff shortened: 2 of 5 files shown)


sentry-sidekiq/spec/sentry/sidekiq/configuration_spec.rb
```diff
@@ -26,4 +26,16 @@
       expect(subject.report_after_job_retries).to eq(false)
     end
   end
+
+  describe "#report_only_dead_jobs" do
+    it "has correct default value" do
+      expect(subject.report_only_dead_jobs).to eq(false)
+    end
+  end
```
solnic, line 34: This seems unrelated?

sl0thentr0py: was missing so just added it

sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
```diff
@@ -95,7 +95,10 @@ def call(worker_class, job, queue, _redis_pool)
 
         user = Sentry.get_current_scope.user
         job["sentry_user"] = user unless user.empty?
-        job["trace_propagation_headers"] ||= Sentry.get_trace_propagation_headers
+
+        if Sentry.configuration.sidekiq.propagate_traces
+          job["trace_propagation_headers"] ||= Sentry.get_trace_propagation_headers
+        end
```
solnic, line 101: Isn't this a breaking change in a way? Previously it would do it unconditionally. Now you have to explicitly enable it. I understand that the original behavior was problematic though so maybe it's fine?

solnic (commented):
Left a question

sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb
sl0thentr0py: it defaults to true

st0012 (approved):

solnic (approved):

PR #2590, opened 2025-03-23
Improve test setup

st0012:
I tried several ways to simplify out test setup but I think it's already very lean. So these are all the things I could find that aren't needed anymore.

#skip-changelog


solnic (approved):
Ahh that's great, thanks this probably fixes some flakyness too

PR #2598, opened 2025-04-14
[active_job] report on each retry failure

### LINKED ISSUE #2597: ActiveJob missing error reporting on retries
As described [here](https://github.com/getsentry/sentry-ruby/pull/2573#issuecomment-2788161525) hooking into `perform_now` is not enough to capture all exceptions that happen during execution of a job because retry logic is implemented as a rescue handler, which simply bypasses entire code flow from the `perform_now`.

~It turned out we need to patch `retry_job` in order to be able to report exceptions *on each retry* but this is needed *only up until Rails 8.0.2 (included)*. Luckily latest Rails' version of ActiveJob uses `ActiveSupport.error.reporter` to report an exception before retrying a job, which means we can provide our own error subscriber and no patching will be needed for Rails > 8.0.2 (not released in the moment of writing this but it's in main).~

It turned out there's `enqueue_retry.active_job` that we can simply rely on for this.

solnic:
### ⚠️ this will increase error reporting rates - we may consider setting `active_job_report_after_job_retries` to `true` by default

Closes #2597 

#skip-changelog


solnic:
Danger is still not seeing my changelog entry :(

sl0thentr0py (approved):

Issue #2600, opened 2025-04-22
Sentry Structured Logging for Ruby


AbhiPrasad:
https://develop.sentry.dev/sdk/telemetry/logs/

Sentry is adding support for [structured logging](https://github.com/getsentry/sentry/discussions/86804). Let's add it to the Ruby SDK!

1. Define the [logs protocol](https://develop.sentry.dev/sdk/telemetry/logs/#logs-protocol) and log envelope item in the SDK
1. Add the Public API (SDK options and methods) as per the [docs](https://develop.sentry.dev/sdk/telemetry/logs/#public-api)
1. Make sure the the SDK follow the [documented behavior](https://develop.sentry.dev/sdk/telemetry/logs/#sdk-behavior)
1. Attach [default attributes](https://develop.sentry.dev/sdk/telemetry/logs/#default-attributes) to the SDK as per docs
1. Instrument popular logging libraries to send logs to Sentry. SDK maintainers can best decide what they should support, but we should aim to get std lib `Logger` (and I guess `ActiveSupport::Logger`) automatically instrumented.
1. Create a GH discussion that contains instructions for setting up the SDK in your repo. Eventually this will be moved into the primary docs. See the [JS SDK's GH discussion](https://github.com/getsentry/sentry-javascript/discussions/15916) for inspiration.

Issue #2604, opened 2025-04-24
Introduce a new Sentry.logger interface


solnic:
When `config._experiments[:enable_logger]` is `true`, the default `Sentry.logger` interface should support logging log events.

This is meant to implement the API as described here https://develop.sentry.dev/sdk/telemetry/logs/#public-api

More requirements described here: https://develop.sentry.dev/sdk/telemetry/logs/#implementation

Initial questions:

- Should we separate an internal logger for logging errors/crashes when things don't work in Sentry SDK? It seems like that's what `Sentry.logger` is used for now?
- Should we support multiple logging handlers, ie logging both to std output and via log events? For example should calling `Sentry.logger.info("Hello")` support sending a log event AND logging to other destinations?
- How should we set `trace_id` when it's not provided explicitly?

sl0thentr0py:
> Should we separate an internal logger for logging errors/crashes when things don't work in Sentry SDK? It seems like that's what Sentry.logger is used for now?

The current `Sentry.logger` should be renamed to `internal_logger` or `sdk_logger` or something, that is just debug logging for the SDK and should be completely separate from capturing log events that go over the network.

> Should we support multiple logging handlers, ie logging both to std output and via log events? For example should calling Sentry.logger.info("Hello") support sending a log event AND logging to other destinations?

For now not, since we're still iterating. Note that what we're offering is **Structured Logging** so it is not the same as logging something locally with a string template. Ideally we want people to pass in the template AND the arguments separately that leads to nicer UX in the product.

> How should we set trace_id when it's not provided explicitly?

we pull it from the currently active transaction/span/propagation_context, see
https://github.com/getsentry/sentry-python/blob/bbdf789902e3d8ee7940d7b7442934b0d6b8b30d/sentry_sdk/client.py#L917-L923

PR #2606, opened 2025-04-25
Introduce Sentry.capture_log

### LINKED ISSUE #2603: Introduce LogEvent
The new `LogEvent` type should be supported by Transport and there should be a new interface `Sentry.capture_log` that creates log events and passes them down to the client just like we do with other event types.

solnic:
First step towards Structured Logging #2600 

This adds `Sentry.capture_log` interface with a new underlying `LogEvent` event type.

Closes #2603

### DIFF
--- a/sentry-ruby/lib/sentry/log_event.rb
+++ b/sentry-ruby/lib/sentry/log_event.rb
@@ -13,16 +13,15 @@ module Sentry
       timestamp
       trace_id
       attributes
-      release
-      sdk
-      platform
-      environment
-      server_name
-      trace_id
     ]
 
     SENTRY_ATTRIBUTES = {
-      "sentry.trace.parent_span_id" => :parent_span_id
+      "sentry.trace.parent_span_id" => :parent_span_id,
+      "sentry.environment" => :environment,
+      "sentry.release" => :release,
+      "sentry.server_name" => :server_name,
+      "sentry.sdk.name" => :sdk_name,
+      "sentry.sdk.version" => :sdk_version
     }
 
     LEVELS = %i[trace debug info warn error fatal].freeze
@@ -60,6 +59,14 @@ module Sentry
       level.to_s
     end
 
+    def serialize_sdk_name
+      Sentry.sdk_meta[:name]
+    end
+
+    def serialize_sdk_version
+      Sentry.sdk_meta[:version]
+    end
+
     def serialize_timestamp
       Time.parse(timestamp).to_f
     end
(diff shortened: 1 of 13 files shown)


sentry-ruby/lib/sentry/log_event.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    private
+
+    def serialize_attributes
+      result = {}
+
+      @attributes.each do |key, value|
+        result[key.to_s] = {
+          value: value.to_s,
```
AbhiPrasad, line 38: We don't want to convert booleans, integers or floats to strings, we can keep them as their original values when serializing to JSON to send via the envelope.

AbhiPrasad (commented):
Looks solid! Just one comment about the client method, and then we are good to ✅ 

Before we move on to the logger interface though (https://github.com/getsentry/sentry-ruby/issues/2604), can we spec out a buffering implementation for logs? We can do it after this PR and create another issue to track it so that we contain the scope of this particular set of changes.

https://develop.sentry.dev/sdk/telemetry/logs/#implementation

It would be nice to get a lightweight logs buffer so that we don't overwhelm sentry with requests, logs were designed so you can send multiple log payloads in a single log envelope.

solnic:
> Before we move on to the logger interface though (#2604), can we spec out a buffering implementation for logs?

@AbhiPrasad yes this makes sense, I reported #2613 :)

solnic:
@AbhiPrasad I added remaining pieces of info to the attributes, please take another look 😄

AbhiPrasad (approved):

sentry-ruby/lib/sentry/log_event.rb
```diff
@@ -0,0 +1,113 @@
...
+      timestamp
+      trace_id
+      attributes
+    ]
+
+    SENTRY_ATTRIBUTES = {
+      "sentry.trace.parent_span_id" => :parent_span_id,
+      "sentry.environment" => :environment,
+      "sentry.release" => :release,
+      "sentry.server_name" => :server_name,
```
AbhiPrasad, line 22:
```suggestion
      "server.address" => :server_name,
```

This follows the OpenTelemetry semantic conventions for server, which we want to follow: https://opentelemetry.io/docs/specs/semconv/attributes-registry/server/#server-attributes

AbhiPrasad (approved):
One change with `server_name`, but approving to unblock! Thanks for the quick pace on this.

PR #2619, opened 2025-05-06
Introduce log event buffering

### LINKED ISSUE #2613: Log buffering
Capturing logs should buffer them instead of sending them right away.

Refs #2606

solnic:
Next step in our logging story - this adds a log event buffer, so that log events are buffered until they hit the limit set as a new config option called `max_log_events`, and then they are all packaged under the same envelope and sent right away.

Closes #2613

I tested it out and it works:

Screenshot 2025-05-06 at 15 56 11

#skip-changelog

### DIFF
--- a/sentry-ruby/lib/sentry/log_event_buffer.rb
+++ b/sentry-ruby/lib/sentry/log_event_buffer.rb
@@ -1,0 +1,85 @@
+# frozen_string_literal: true
+
+require "sentry/threaded_periodic_worker"
+
+module Sentry
+  class LogEventBuffer < ThreadedPeriodicWorker
+    FLUSH_INTERVAL = 5 # seconds
+    DEFAULT_MAX_EVENTS = 100
+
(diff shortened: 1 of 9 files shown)


AbhiPrasad (approved):
nice!

sentry-ruby/lib/sentry/log_event_buffer.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+
+    def start
+      ensure_thread
+      self
+    end
+
+    def flush
+      @mutex.synchronize do
+        return unless size >= @max_events
```
sl0thentr0py, line 29: in the case where someone adds only say 5 logs, these will never get sent, even at the exit flush

sl0thentr0py: what you want instead is 
* a periodic flush all the time
* a force flush when the size exceeds while adding

PR #2621, opened 2025-05-07
Replace logger with sdk_logger

solnic:
This is in preparation for #2604 - we have an internal `Sentry.logger` used for warn/error/debug logging with the SDK but with Structured Logging feature we're introducing a public-facing `Sentry.logger` API, so our internal one needs to be renamed.

### DIFF
--- a/sentry-ruby/lib/sentry-ruby.rb
+++ b/sentry-ruby/lib/sentry-ruby.rb
@@ -617,6 +617,16 @@ module Sentry
     ##### Helpers #####
 
     # @!visibility private
+    def logger
+      warn <<~STR
+        [sentry] `Sentry.logger` will no longer be used as internal SDK logger when `enable_logs` feature is turned on.
+                 Use Sentry.configuration.sdk_logger for SDK-specific logging needs."
+      STR
+
+      configuration.sdk_logger
+    end
+
+    # @!visibility private
     def sys_command(command)
       result = `#{command} 2>&1` rescue nil
       return if result.nil? || result.empty? || ($CHILD_STATUS && $CHILD_STATUS.exitstatus != 0)
--- a/sentry-ruby/lib/sentry-ruby.rb
+++ b/sentry-ruby/lib/sentry-ruby.rb
@@ -94,6 +94,11 @@ module Sentry
     #   @return [Metrics::Aggregator, nil]
     attr_reader :metrics_aggregator
 
+    # @!attribute [r] logger
+    #   @return [Logger]
+    # @!visibility private
+    attr_reader :sdk_logger
+
     ##### Patch Registration #####
 
     # @!visibility private
@@ -138,7 +143,7 @@ module Sentry
     # @param version [String] version of the integration
     def register_integration(name, version)
       if initialized?
-        logger.warn(LOGGER_PROGNAME) do
+        sdk_logger.warn(LOGGER_PROGNAME) do
           <<~MSG
             Integration '#{name}' is loaded after the SDK is initialized, which can cause unexpected behavior.  Please make sure all integrations are loaded before SDK initialization.
           MSG
@@ -238,6 +243,10 @@ module Sentry
     def init(&block)
       config = Configuration.new
       yield(config) if block_given?
+
+      # Internal SDK logger
+      @sdk_logger = config.sdk_logger
+
       config.detect_release
       apply_patches(config)
       config.validate
@@ -613,11 +622,6 @@ module Sentry
       return if result.nil? || result.empty? || ($CHILD_STATUS && $CHILD_STATUS.exitstatus != 0)
 
       result.strip
-    end
-
-    # @!visibility private
-    def logger
-      configuration.logger
     end
 
     # @!visibility private
--- a/sentry-ruby/lib/sentry/client.rb
+++ b/sentry-ruby/lib/sentry/client.rb
@@ -19,8 +19,6 @@ module Sentry
     # @!macro configuration
     attr_reader :configuration
 
-    attr_reader :log_event_buffer
-
     # @param configuration [Configuration]
     def initialize(configuration)
       @configuration = configuration
--- a/sentry-ruby/lib/sentry/client.rb
+++ b/sentry-ruby/lib/sentry/client.rb
@@ -22,7 +22,7 @@ module Sentry
     # @param configuration [Configuration]
     def initialize(configuration)
       @configuration = configuration
-      @logger = configuration.sdk_logger
+      @sdk_logger = configuration.sdk_logger
 
       if transport_class = configuration.transport.transport_class
         @transport = transport_class.new(configuration)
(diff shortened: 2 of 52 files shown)


AbhiPrasad (approved):

sentry-ruby/lib/sentry-ruby.rb
```diff
@@ -615,11 +624,6 @@ def sys_command(command)
       result.strip
     end
 
-    # @!visibility private
-    def logger
```
st0012, line 619: Not blocking: Although we marked it as private in docs, it's still very accessible to users. So I'd suggest keeping it like:

```rb
def logger
  print_deprecation_message_to_ask_users_not_to_use_it
  configuration.sdk_logger
end
```

Then at least some users of it can have time to move off of it.

sentry-ruby/lib/sentry/client.rb
```diff
@@ -19,13 +19,12 @@ class Client
     # @!macro configuration
     attr_reader :configuration
 
-    # @deprecated Use Sentry.logger to retrieve the current logger instead.
-    attr_reader :logger
+    attr_reader :log_event_buffer
```
st0012, line 22: Why do we need this change?

sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -184,7 +184,9 @@ def capture_exception_frame_locals=(value)
     # Logger used by Sentry. In Rails, this is the Rails logger, otherwise
     # Sentry provides its own Sentry::Logger.
     # @return [Logger]
-    attr_accessor :logger
+    attr_accessor :sdk_logger
+    alias_method :logger, :sdk_logger
```
st0012, line 188: Let's print warning messages when users try to access these `logger` accessors.

sentry-ruby/lib/sentry/log_event_buffer.rb
```diff
@@ -7,16 +7,19 @@ class LogEventBuffer < ThreadedPeriodicWorker
     FLUSH_INTERVAL = 5 # seconds
     DEFAULT_MAX_EVENTS = 100
 
+    attr_reader :pending_events
```
st0012, line 10: Is this needed for the logger change?

sentry-ruby/lib/sentry/utils/logging_helper.rb
```diff
@@ -2,21 +2,23 @@
 
 module Sentry
   module LoggingHelper
+    attr_reader :logger
```
st0012, line 5: Maybe we should name this `sdk_logger` too to standardize how we access internal logging. It'll make grepping the codebase easier.

sentry-ruby/lib/sentry-ruby.rb
solnic: @st0012 so, this is tricky - in #2600 we're adding `Sentry.logger` interface, so it is not going to be deprecated, but it will be enabled conditionally based on `config._experiments[:enable_logs]`. What we could do is printing deprecation message *only if* enabled_logs is not `true` - WDYT? The message would say something like "Starting with 5.{version_where_we_enable_logging_by_default} Sentry.logger will become public API for structured logging feature"

sentry-ruby/lib/sentry/client.rb
solnic: Good catch!

sentry-ruby/lib/sentry-ruby.rb
solnic: @sl0thentr0py I think there's no way around that - IF there are people who use Sentry.logger, then we gotta let them know via deprecation warning that this logger is changing its purpose. I'm not sure how changing to Sentry::Logger class methods would help?

sl0thentr0py: ok let me be clearer,

about the internal SDK logging:
* `config.logger` stays but deprecated
* `config.sdk_logger` comes and replaces `config.logger`

Now, everything about the **new Sentry Logging Feature** goes under a completely new namespace `Sentry::Logging.do_something` (not `Sentry::Logger` since that's also taken) and **NOT** `Sentry.logger.do_something`

solnic: > Now, everything about the new Sentry Logging Feature goes under a completely new namespace Sentry::Logging.do_something (not Sentry::Logger since that's also taken) and NOT Sentry.logger.do_something

I don't think it's a good idea as it would diverge from the naming convention from other SDKs where we already have `Sentry.logger` - also, it's just much nicer and intuitive vs `Sentry::Logging`.

What would be the benefit of going with `Sentry::Logging`? The way I see it is that we just need to deal with the deprecation cycle of `config.logger` / `Sentry.logger` being used as an internal logging facility and eventually replace it with the new logging stuff.

Please remember that initially we are going to release new logging with `_experiments[:enable_logger]` config option, so the users will have to make a deliberate choice to enable it and we can simply add docs/release notes + maybe even extra warning on init, that say that when this is enabled, then `config.logger` / `Sentry.logger` is no longer the same thing and if the user relies on it, then they should basically stop doing that as it was always meant to be used solely as an internal SDK logger.

sl0thentr0py: resolved in slack, we will show deprecation and do the switch based on the config

solnic: @st0012 @sl0thentr0py I re-added `Sentry.logger` method via cb555446cd4e4bd3fdea757413a7effb662acc3c and it has a deprecation warning now:

Screenshot 2025-05-12 at 15 03 36

sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -184,7 +184,19 @@ def capture_exception_frame_locals=(value)
...
+    # @deprecated Use {#sdk_logger=} instead.
+    def logger=(logger)
+      warn "[sentry] `config.logger=` is deprecated. Please use `config.sdk_logger=` instead."
+      self.sdk_logger = logger
+    end
+
+    # @deprecated Use {#sdk_logger} instead.
+    def logger
+      warn "[sentry] `config.logger` is deprecated. Please use `config.sdk_logger` instead."
+      self.sdk_logger = logger
```
sl0thentr0py, line 198:
```suggestion
      self.sdk_logger
```

sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -312,6 +324,10 @@ def capture_exception_frame_locals=(value)
     # @return [Integer]
     attr_accessor :max_log_events
 
+    # Experimental features configuration
+    # @return [Hash]
+    attr_accessor :_experiments
```
sl0thentr0py, line 329: remove this as discussed

sl0thentr0py (changes requested):

sentry-ruby/lib/sentry/configuration.rb
solnic: @sl0thentr0py fixed it via 5f0788f83878d18b2f6ba3202977d52b292e61ba + added specs

sl0thentr0py (approved):

PR #2629, opened 2025-05-15
Fix coverage by making sure isolated specs have SimpleCov.command_name

sl0thentr0py:
also
* make sure only spec files are run by the `:isolated_specs` task (and not the helper ruby init file)
* upload coverage for all delayed_job runs

see https://github.com/simplecov-ruby/simplecov?tab=readme-ov-file#merging-results

#skip-changelog


solnic (approved):
🙇🏻

PR #2634, opened 2025-05-19
Support for before send log

### LINKED ISSUE #2633: Support for `before_send_log`
### Problem Statement

As described https://develop.sentry.dev/sdk/telemetry/logs/#initialization-options

### Solution Brainstorm

_No response_

solnic:
This adds support for a new configuration option called `before_send_log`.

Closes #2633

### DIFF
--- a/sentry-ruby/lib/sentry/client.rb
+++ b/sentry-ruby/lib/sentry/client.rb
@@ -272,6 +272,36 @@ module Sentry
       raise
     end
 
+    # Send an envelope with batched logs
+    # @param log_events [Array<LogEvent>] the log events to be sent
+    # @api private
+    # @return [void]
+    def send_logs(log_events)
+      envelope = Envelope.new(
+        event_id: SecureRandom.uuid.delete("-"),
+        sent_at: Sentry.utc_now.iso8601,
+        dsn: configuration.dsn,
+        sdk: Sentry.sdk_meta
+      )
+
+      items = if configuration.before_send_log
+        log_events.map { |log_event| configuration.before_send_log.call(log_event) }
+      else
+        log_events
+      end
+
+      envelope.add_item(
+        {
+          type: "log",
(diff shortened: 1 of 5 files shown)


solnic:
@sl0thentr0py do we need to support `hint` argument here too? Also, any...hints how to handle recording lost events in case of logs? 🙂

sl0thentr0py:
checked python (empty placeholder hint) and JS (no hint)

https://github.com/getsentry/sentry-python/blob/8be519faa2d7df2d1882fa2f57cbca13cc899328/sentry_sdk/client.py#L937-L941
https://github.com/getsentry/sentry-javascript/blob/61940fc22052e9005043f77c7355db1742d1d78c/packages/core/src/logs/exports.ts#L132

let's just go with no hint since I can't see any additional info we can put in there anyway

the client report you can use 
```
record_lost_event(:before_send, "log_item")
```

sentry-ruby/lib/sentry/client.rb
```diff
@@ -272,6 +273,53 @@ def send_event(event, hint = nil)
...
+    # @return [void]
+    def send_logs(log_events)
+      envelope = Envelope.new(
+        event_id: Sentry::Utils.uuid,
+        sent_at: Sentry.utc_now.iso8601,
+        dsn: configuration.dsn,
+        sdk: Sentry.sdk_meta
+      )
+
+      discarded_logs = []
```
sl0thentr0py, line 288: let's just keep a count since we don't need the discarded events to be slightly more memory efficient

sl0thentr0py (changes requested):

sentry-ruby/lib/sentry/client.rb
solnic: @sl0thentr0py done

sl0thentr0py (approved):

PR #2643, opened 2025-06-06
Optimize LogEvent to reduce memory consumption

### LINKED ISSUE #2641: Increase Memory usage in Logging
### Issue Description

We're trying out logging with Sentry - Firstly as a partial log-sink for Rails logs. Rails is not supported officially yet, but our initial attempts has shown somewhat highly increased memory in preprod environments. Is this something you're aware of, or something that we/I should dig into?

### Reproduction Steps

Add "mirrored" logging to for config.logger in Rails initializer:

```ruby
dual_logger = Class.new(Logger) do
    LOGGER_TO_SENTRY = {
      Logger::Severity::DEBUG => :debug,
      Logger::Severity::INFO => :info,
      Logger::Severity::WARN => :warning,
      Logger::Severity::ERROR => :error,
      Logger::Severity::FATAL => :fatal,
      Logger::Severity::UNKNOWN => :fatal
    }
    def initialize(input)
      super(input)

      @stdout_logger = ActiveSupport::TaggedLogging.new(ActiveSupport::Logger.new(input))
      @stdout_logger.level = Logger::DEBUG
    end

    def add(severity, message = nil, progname = nil)
      message ||= yield if block_given?

      return true if message.nil? || message.empty?

      @stdout_logger.add(severity, message, progname)
      Sentry.logger.log(LOGGER_TO_SENTRY[severity], message, parameters: []) if severity >= Logger::Severity::INFO
    end
  end

  Rails.logger = dual_logger.new($stdout)
```

This is not exactly production-ready, or pretty; But solves the issue of logging to default STDOUT logger, as well as `Sentry.logger`.

And while this DOES work - Memory increases massively (~40% in our test-env on our application until OOM in Kubernetes).

I have not determined if it's leaking, or if it's just fragmenting a lot more, and leaving huge empty pages etc.

### Expected Behavior

Memory can increase, but should not increase this much for logging.

### Actual Behavior

Memory increases until OOM

### Ruby Version

3.4.3

### SDK Version

5.24.0

ruby, rails, sidekiq

### Sentry Config

```ruby
  require 'p2b/sentry/active_record_instantiation_subscriber'

  Sentry.init do |config|
    config.dsn = 'https://onprem.sentry/...'
    config.enabled_environments = %w[production staging]
    config.skip_rake_integration = true
    config.send_modules = false

    config.traces_sample_rate = Rails.env.production? ? 0.4 : 1
    config.profiles_sample_rate = Rails.env.production? ? 0.2 : 1
    if Rails.env.staging?
      config.debug = true
      config.enable_logs = true
      config.sdk_logger.level = Logger::Severity::DEBUG
    else
      config.sdk_logger.level = Logger::Severity::WARN
    end

    config.enabled_patches += [:sidekiq_cron, :excon, :redis, :graphql]
    # Hold that thought. Would be nice to have this as nested spans. How though.
    config.rails.tracing_subscribers << P2b::Sentry::ActiveRecordInstantiationSubscriber
      if hint[:exception].is_a?(ActiveRecord::RecordInvalid)
        begin
          event.set_tags(e.record.errors.to_hash.to_h { ["errors.#{_1}", [*_2].join(';')] })
        rescue
          # This is ok. Don't error out if we can't set tags.
        end
      end

      event
    end
  end
```

solnic:
## Current status

```
Git SHA      Branch     Iter     Allocated    Retained     Alloc Mem    Ret Mem     
------------------------------------------------------------------------------------------
213558f3     master     100,000  9,681,713    1,395,346    921.32 MB    217.9 MB    
cafd4cdc     2641-incre 100,000  2,920,177    1,598,060    445.48 MB    285.83 MB   

=== Performance Trends ===

Comparing latest (cafd4cdc) vs baseline (213558f3):

Object Allocation: 📉 -69.84% (decrease)
Object Retention: 📈 +14.53% (increase)
Memory Usage: 📉 -51.65% (decrease)

Memory per iteration:
  Baseline: 9.43 KB
  Latest:   4.56 KB
  Change:   📉 -51.65% (decrease)

=== Recent Commits ===

213558f3 (2025-06-09 12:15)
  [sentry-ruby] fix and improve spec suite setup (#2640)
  Memory: 921.32 MB allocated, 217.9 MB retained

cafd4cdc (2025-06-09 12:15)
  Update CHANGELOG
  Memory: 445.48 MB allocated, 285.83 MB retained
```

Closes #2641

### DIFF
--- a/sentry-ruby/lib/sentry/log_event.rb
+++ b/sentry-ruby/lib/sentry/log_event.rb
@@ -9,7 +9,6 @@ module Sentry
 
     DEFAULT_PARAMETERS = [].freeze
     DEFAULT_ATTRIBUTES = {}.freeze
-    DEFAULT_CONTEXT = {}.freeze
 
     SERIALIZEABLE_ATTRIBUTES = %i[
       level
--- a/sentry-ruby/lib/sentry/log_event.rb
+++ b/sentry-ruby/lib/sentry/log_event.rb
@@ -39,6 +39,17 @@ module Sentry
 
     attr_reader :configuration, *SERIALIZEABLE_ATTRIBUTES
 
+    SERIALIZERS = %i[
+      attributes
+      body
+      level
+      parent_span_id
+      sdk_name
+      sdk_version
+      timestamp
+      trace_id
+    ].map { |name| [name, :"serialize_#{name}"] }.to_h
+
     def initialize(configuration: Sentry.configuration, **options)
       @configuration = configuration
       @type = TYPE
@@ -62,9 +73,9 @@ module Sentry
     private
 
     def serialize(name)
-      serializer = :"serialize_#{name}"
+      serializer = SERIALIZERS[name]
 
-      if respond_to?(serializer, true)
+      if serializer
         __send__(serializer)
       else
         public_send(name)
--- a/sentry-ruby/lib/sentry/log_event.rb
+++ b/sentry-ruby/lib/sentry/log_event.rb
@@ -50,6 +50,13 @@ module Sentry
       trace_id
     ].map { |name| [name, :"serialize_#{name}"] }.to_h
 
+    VALUE_TYPES = Hash.new("string").merge!({
+      TrueClass => "boolean",
+      FalseClass => "boolean",
+      Integer => "integer",
+      Float => "double"
+    }).freeze
+
     def initialize(configuration: Sentry.configuration, **options)
       @configuration = configuration
       @type = TYPE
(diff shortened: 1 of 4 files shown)


sentry-ruby/lib/sentry/log_event.rb
```diff
@@ -123,17 +123,15 @@ def attribute_hash(value)
       { value: value, type: value_type(value) }
     end
 
+    VALUE_TYPES = Hash.new("string").merge!({
```
dentarg: Gotta be careful when using the [default value](https://docs.ruby-lang.org/en/3.4/Hash.html#class-Hash-label-Default+Values)
```ruby
$ irb
irb(main):001> a = Hash.new("string")
=> {}
irb(main):002> a["foo"]
=> "string"
irb(main):003> a["bar"] << "bar"
=> "stringbar"
irb(main):004> a["foo"]
=> "stringbar"
```

sl0thentr0py: @solnic 🙏

solnic: @dentarg @sl0thentr0py I believe we're good since there's frozen string literals magic comment at the top of this file.

Issue #2644, opened 2025-06-09
Logger support


solnic:
### Description

As part of the Logging feature we want to provide an opt-in integration with the stdlib `Logger` so that any usage of any logger instance would result in sending logs to Sentry too.

Eazybright:
This feature came to my mind as well. 
@solnic i would love to contribute to the development of this feature.

AbhiPrasad:
@solnic could we get docs for this? (apologies if this is tracked somewhere else, couldn't find it)

you can look at https://docs.sentry.io/platforms/go/logs/#integrations or https://docs.sentry.io/platforms/javascript/logs/#integrations for reference on what this should look like.

solnic:
@AbhiPrasad oh sure thing!

PR #2653, opened 2025-06-18
implement stblib logger support

Eazybright:
## Description
Describe your changes:
This is a fun way of adding integration for stdlib Logger so that any usage of any logger instance would result in sending logs to Sentry too. [issue](https://github.com/getsentry/sentry-ruby/issues/2644)

### Sample Config
```ruby
require "sentry-ruby"

# Initialize Sentry with experimental logging feature enabled
Sentry.init do |config|
  config.dsn = "YOUR_SENTRY_DSN"
  config.environment = "development"

  # Set this to enable structured logging via Sentry.logger
  config.enable_logs = true

  # set this to send any instance of stdlib logger to sentry
  config.send_stdlib_logs = true
end

logger = Logger.new(STDOUT)
logger.info('send_stdlib_log') { "send_stdlib_log_enable_logs: Testing 789..." }
logger.warn("ruby logger warn  - - 9")

```

### Sample Result on the UI
Screenshot 2025-06-18 at 11 34 29 PM


sentry-ruby/lib/sentry/std_lib_logger.rb
```diff
@@ -0,0 +1,45 @@
...
+		  0 => :debug,
+		  1 => :info,
+		  2 => :warn,
+		  3 => :error,
+		  4 => :fatal
+		}.freeze
+
+		def add(severity, message = nil, progname = nil, &block)
+
+			return unless Sentry.initialized? && Sentry.get_current_hub
```
sl0thentr0py: we need a `super` here and below since the original logger should not be affected

sl0thentr0py (changes requested):
left a small comment, will leave a proper review to @solnic

sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -479,6 +485,7 @@ def initialize
       self.traces_sampler = nil
       self.enable_tracing = nil
       self.enable_logs = false
+      self.send_stdlib_logs = false
```
solnic: This is not needed, there's a config option called `enabled_patches` that should be used instead. The Logger patch should be registered as `:logger` and then the users will be able to configure it to be enabled by default.

You can see how `:http` patch is done for reference.

sentry-ruby/lib/sentry/std_lib_logger.rb
```diff
@@ -0,0 +1,45 @@
...
+
+			config = Sentry.configuration
+			return unless config.enable_logs && config.send_stdlib_logs
```
solnic: This guard won't be needed when you add patching based on our `enabled_patches` option like I mentioned in the other comment. When registering your patch via `register_patch` you'll be able to do something like:

```ruby
Sentry.register_patch(:logger, Sentry::StdLibLogger, ::Logger) do |config|
  if config.enable_logs
    ::Logger.prepend(Sentry::StdLibLogger)
  else
    Sentry.sdk_logger.warn(":logger patch enabled but `enable_logs` is turned off - skipping applying patch")
  end
end
```

sentry-ruby/lib/sentry/std_lib_logger.rb
```diff
@@ -0,0 +1,45 @@
...
+	      else
+	        message = progname
+	      end
+
+	      message = message.to_s.strip
+
+	      if !message.nil? && message != Sentry::Logger::PROGNAME && method = SEVERITY_MAP[severity]
+					Sentry.logger.send(method, message)
+	      end
+	    end
```
solnic: I think this method should always call `super` and logs to Sentry as an additional side effect.

@sl0thentr0py wouldn't you agree?

solnic (changes requested):
Thank you for the PR! I've left some initial comments. Please also ensure that `rubocop` passes against the files you added and changed because there are issues with formatting 🙂

sentry-ruby/lib/sentry/std_lib_logger.rb
solnic: Oh wait nevermind @sl0thentr0py - I just noticed that you pointed it out as well above, sorry for the noise!

Eazybright:
> Thank you for the PR! I've left some initial comments. Please also ensure that `rubocop` passes against the files you added and changed because there are issues with formatting 🙂

all comments have been addressed and PR updated. @solnic  @sl0thentr0py

solnic:
Thanks for the PR and for following up on the reviews! 🙇🏻

I've opened a new PR that builds on top of yours because of time constraints, as we need to finish this feature this week and we may need more back and worth in the review process.

See #2657 for more progress!

Gonna close this one because of this, I hope that's OK! All your commits are in my branch of course :)

Eazybright:
> Thanks for the PR and for following up on the reviews! 🙇🏻
> 
> I've opened a new PR that builds on top of yours because of time constraints, as we need to finish this feature this week and we may need more back and worth in the review process.
> 
> See #2657 for more progress!
> 
> Gonna close this one because of this, I hope that's OK! All your commits are in my branch of course :)

Totally fine!

solnic:
@Eazybright thanks again for helping with this. I just merged my PR with your work included. We're gonna ship it in the next release 🙂

Eazybright:
> @Eazybright thanks again for helping with this. I just merged my PR with your work included. We're gonna ship it in the next release 🙂

Great work @solnic 

Thanks for the review, i learnt from it 🙏

sl0thentr0py:
thanks for the contribution @Eazybright! if you want a free Sentry t-shirt, send me an email on `neel dot shah at sentry.io` and we'll ship it to you :)

Eazybright:
> thanks for the contribution @Eazybright! if you want a free Sentry t-shirt, send me an email on `neel dot shah at sentry.io` and we'll ship it to you :)

Will do, thanks!

PR #2657, opened 2025-06-24
Logger support

solnic:
This builds on #2653.

I had to isolate specs because we can't patch logger while running the whole spec suite.

#skip-changelog


solnic:
OK I need to rework this because there are issues with the spec suite as testing is not isolated and patching obviously changes runtime. Right now it gets to a state that would normally be very unlikely to have where the patch is applied but `enable_logs` is disabled, causing `Sentry.logger` to be no longer `StructuredLogger` but the sdk logger which inherits from `Logger`...leading to stack call errors. Fun! :)

solnic:
@sl0thentr0py ignore head failures, something is messed up with bundling rn 🤷🏻

sentry-ruby/spec/isolated/sentry_logger_spec.rb
sl0thentr0py, line 1: also needs a `SimpleCov.command_name` since it is now isolated

sl0thentr0py (approved):

PR #2659, opened 2025-06-26
Tests for Rails.logger when :logger patch is enabled

solnic:
This adds tests for `Rails.logger` when `:logger` patch is enabled. I also add a unified way of defining spec tasks for our projects so that I can easily add an isolated spec to sentry-rails.

Refs #2644 

#skip-changelog

### DIFF
--- a/sentry-rails/spec/isolated/rails_logger_patch_spec.rb
+++ b/sentry-rails/spec/isolated/rails_logger_patch_spec.rb
@@ -1,0 +1,157 @@
+# frozen_string_literal: true
+
+begin
+  require "simplecov"
+  SimpleCov.command_name "RailsLoggerPatch"
+rescue LoadError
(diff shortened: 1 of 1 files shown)


solnic:
@sl0thentr0py no idea what's going on with ruby head yet so pls ignore it again 🙃

sentry-ruby/lib/sentry/test/rake_tasks.rb
sl0thentr0py, line 1: these are only spec related and should not live in `lib` which is the actual library, you can have it as a shared task definition in the main project root maybe.

sl0thentr0py (changes requested):

solnic:
@sl0thentr0py OK I moved the rake tasks stuff to the toplevel

PR #2664, opened 2025-07-28
Add Sentry::DebugTransport for testing/debugging

solnic:
Long story short: I think this is something very useful to have. I missed something like this in the past and when I worked on e2e test setup, I figured it will be a good moment to finally add it because it helped with automated e2e testing too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Internal
 
 - Factor out do_request in HTTP transport ([#2662](https://github.com/getsentry/sentry-ruby/pull/2662))
+- Add `Sentry::DebugTransport` that captures events and stores them as JSON for debugging purposes ([#2664](https://github.com/getsentry/sentry-ruby/pull/2664))
 
 ## 5.26.0
 


sentry-ruby/lib/sentry/test_helper.rb
```diff
@@ -46,6 +46,11 @@ def setup_sentry_test(&block)
     def teardown_sentry_test
       return unless Sentry.initialized?
 
+      transport = Sentry.get_current_client&.transport
+      if transport.is_a?(Sentry::DebugTransport)
+        transport.clear_events
+      end
```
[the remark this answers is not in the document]
sl0thentr0py: you need to fix this I guess?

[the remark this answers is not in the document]
solnic: Yeah I must've messed up rebasing.

solnic:
@sl0thentr0py mind taking another look? I made a couple of extra improvements and refactored DebugTransport into a delegator.

sentry-ruby/spec/sentry_spec.rb
```diff
@@ -390,7 +390,8 @@
 
       Sentry.get_current_client.flush
 
-      expect(sentry_envelopes.size).to be(2)
+      # 3 envelopes: log, transaction, and client_report about dropped profile
```
sl0thentr0py, line 393: we actually shouldn't be sending the dropped profile client report here since we don't have `profiles_sample_rate` enabled, we should fix this separately

sentry-ruby/spec/spec_helper.rb
```diff
@@ -79,6 +79,14 @@
   end
 
   config.after(:each) do
+    if Sentry.initialized?
+      transport = Sentry.get_current_client&.transport
+
+      if transport.is_a?(Sentry::DebugTransport)
```
sl0thentr0py, line 85: we aren't actually using the `DebugTransport` anywhere in the tests so why is this needed?

sentry-ruby/spec/sentry/rack/capture_exceptions_spec.rb
```diff
@@ -667,11 +667,15 @@ def will_be_sampled_by_sdk
 
           Sentry.session_flusher.flush
 
-          expect(sentry_envelopes.count).to eq(1)
-          envelope = sentry_envelopes.first
+          expect(sentry_envelopes.count).to eq(3)
```
sl0thentr0py, line 670: why did these values change?

sentry-ruby/spec/spec_helper.rb
solnic: @sl0thentr0py another PR will actually use this in a couple of e2e test

sentry-ruby/spec/sentry/rack/capture_exceptions_spec.rb
solnic: @sl0thentr0py because the test actually sends 2 errors and flushes session which sends another event, so total is 3 envelopes, previously dummy transport wouldn't catch 2 other envelopes

sl0thentr0py (approved):

sentry-ruby/spec/sentry_spec.rb
solnic: @sl0thentr0py yeah it seemed suspicious, I reported an issue about this https://github.com/getsentry/sentry-ruby/issues/2669

PR #2666, opened 2025-07-29
Fix flaky faraday spec

solnic:
#skip-changelog


solnic:
@sl0thentr0py jruby is broken again, I'm trying to address it in #2665 separately :(

PR #2667, opened 2025-07-29
Fix flaky net/http spec

solnic:
Fixes this:

```
  1) Sentry::Net::HTTP with tracing enabled when there're multiple requests doesn't mess different requests' data together when making multiple requests with Net::HTTP.start
     Failure/Error:
               Net::HTTP.start("example.com") do |http|
                 stub_normal_response(code: "200")
                 request = Net::HTTP::Get.new("/path?foo=bar")
                 response = http.request(request)
                 expect(response.code).to eq("200")
       
                 stub_normal_response(code: "404")
                 request = Net::HTTP::Get.new("/path?foo=bar")
                 response = http.request(request)
                 expect(response.code).to eq("404")

     Net::OpenTimeout:
       Failed to open TCP connection to example.com:80 (execution expired)
```

#skip-changelog

### DIFF
--- a/sentry-ruby/spec/contexts/with_request_mock.rb
+++ b/sentry-ruby/spec/contexts/with_request_mock.rb
@@ -36,4 +36,14 @@ RSpec.shared_context "with request mock" do
   def stub_normal_response(code: "200", &block)
     sentry_stub_request(build_fake_response(code), &block)
   end
+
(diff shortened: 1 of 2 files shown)


solnic:
@sl0thentr0py just like previously, ignore jruby failures as I fixed them in #2665

PR #2670, opened 2025-07-31
More flaky spec fixes

solnic:
2 more flaky specs addressed, hopefully


sl0thentr0py (approved):
i'm fine with merging because it's ultimately tests, but this is fixing the symptom by adding more non-determinism and not the cause of the original non-determinism.

solnic:
@sl0thentr0py I agree, this is more about reducing the risk of failing under jruby. I'd like to revisit this once I'm done with trace propagation / sample rand stuff.

PR #2682, opened 2025-08-23
Silence `_perform` method redefinition warning

viralpraxis:
I'd like to get rid of this noisy warning:

```
gems/sentry-rails-5.26.0/lib/sentry/rails/background_worker.rb:5: warning: method redefined; discarding old _perform
gems/sentry-ruby-5.26.0/lib/sentry/background_worker.rb:74: warning: previous definition of _perform was here
```

#skip-changelog

### DIFF
--- a/sentry-rails/lib/sentry/rails/background_worker.rb
+++ b/sentry-rails/lib/sentry/rails/background_worker.rb
@@ -2,14 +2,18 @@
 
 module Sentry
   class BackgroundWorker
-    def _perform(&block)
-      block.call
-    ensure
-      # some applications have partial or even no AR connection
-      if ActiveRecord::Base.connected?
-        # make sure the background worker returns AR connection if it accidentally acquire one during serialization
-        ActiveRecord::Base.connection_pool.release_connection
+    module ActiveRecordConnectionPatch
+      def _perform(&block)
+        super(&block)
+      ensure
+        # some applications have partial or even no AR connection
+        if ActiveRecord::Base.connected?
+          # make sure the background worker returns AR connection if it accidentally acquire one during serialization
+          ActiveRecord::Base.connection_pool.release_connection
+        end
       end
     end
   end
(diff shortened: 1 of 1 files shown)


sentry-rails/lib/sentry/rails/background_worker.rb
```diff
@@ -2,6 +2,7 @@
 
 module Sentry
   class BackgroundWorker
+    alias :_perform :_perform # Silence method redefinition warning.
```
solnic: How about using a module with prepend instead?

solnic (commented):
Thanks for the PR! We definitely do not want to have such warnings :) I left one comment.

sentry-rails/lib/sentry/rails/background_worker.rb
viralpraxis: Yep, makes sense 👍🏻

sentry-rails/lib/sentry/rails/background_worker.rb
```diff
@@ -2,14 +2,16 @@
...
+        super(&block)
+      ensure
+        # some applications have partial or even no AR connection
+        if ActiveRecord::Base.connected?
+          # make sure the background worker returns AR connection if it accidentally acquire one during serialization
+          ActiveRecord::Base.connection_pool.release_connection
+        end
       end
-    end
+    end)
```
solnic: Please define a module explicitly, it's better than anonymous modules. You can call it `Sentry::BackgroundWorker::ActiveRecordConnectionPatch`.

solnic (changes requested):
One more small adjustment needed 🙏🏻

sentry-rails/lib/sentry/rails/background_worker.rb
viralpraxis: Done.

PR #2683, opened 2025-08-25
feat: include otel as custom sampling contex

saviogl:
Sentry's OpenTelemetry (OTel) adapter does not do the greatest job in adapting the OTel span to a semantically correct sentry span, nor does it expose the OTel span in the context for sampling decisions. This PR adds a OTel span as an additional custom sampling context so it can be used for sampling decisions. This gives clients more opportunities to use the entire context of the OTel span for sampling decisions, and allows replicating sampling mechanisms as originally suggested by the Sentry team via the "Sampling Function".

https://docs.sentry.io/platforms/ruby/configuration/sampling/#setting-a-sampling-function

**Other Notes**

## Description
Describe your changes:

### DIFF
--- a/sentry-opentelemetry/spec/sentry/opentelemetry/span_processor_spec.rb
+++ b/sentry-opentelemetry/spec/sentry/opentelemetry/span_processor_spec.rb
@@ -154,6 +154,20 @@ RSpec.describe Sentry::OpenTelemetry::SpanProcessor do
 
     it 'starts a sentry transaction on otel root span' do
       expect(Sentry).to receive(:start_transaction).and_call_original
+      expect_any_instance_of(Sentry::Transaction).to receive(:set_initial_sample_decision).with(
+        sampling_context: a_hash_including(
+          otel: {
+            attributes: root_span.attributes,
+            resource: root_span.resource.attribute_enumerator.to_h,
+            kind: root_span.kind,
+            instrumentation_scope: {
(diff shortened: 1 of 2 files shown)


sentry-opentelemetry/spec/sentry/opentelemetry/span_processor_spec.rb
```diff
@@ -154,6 +154,20 @@
...
+            attributes: root_span.attributes,
+            resource: root_span.resource.attribute_enumerator.to_h,
+            kind: root_span.kind,
+            instrumentation_scope: {
+              name: root_span.instrumentation_scope.name,
+              version: root_span.instrumentation_scope.version
+            }
+          }
+        )
+      )
```
solnic: It would be good to not rely on mocking here, especially that it touches internals. Would you be able to improve this spec, or maybe cover this change in another spec where mocks are not needed?

solnic (changes requested):

sentry-opentelemetry/spec/sentry/opentelemetry/span_processor_spec.rb
saviogl: Addressed [here](https://github.com/getsentry/sentry-ruby/pull/2683/commits/7c8ab68175b91edaab8db9ddeab5293445f6c367)

PR #2687, opened 2025-08-26
Rails test runner

solnic:
This adds `bin/test` CLI for running Rails test suite locally against different rails versions. It's just for convenience and I've found it very helpful (FWIW).

The other updates in gemfiles fixed bundling issues that I experienced while running tests (CI didn't catch it because we don't run all of possible combinations).

Depends on #2688 

#skip-changelog

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ node_modules
 
 .devcontainer/.env
 vendor/gems
+sentry-rails/Gemfile-*.lock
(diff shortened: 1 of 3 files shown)


sl0thentr0py (approved):
fine since this is just a helper script, if we want to actually make CI consistent with local dev, we should look at something like
https://github.com/thoughtbot/appraisal

solnic:
> fine since this is just a helper script, if we want to actually make CI consistent with local dev, we should look at something like
> https://github.com/thoughtbot/appraisal

@sl0thentr0py happy to evaluate it, I used it a bunch of times a long time ago but had some issues with it (can't remember what exactly though). I added the script mostly because it manages dedicate lock files, which helped with bundling and made it faster.

PR #2690, opened 2025-08-27
Rails active support log subscribers

### LINKED ISSUE #2605: Support for ActiveSupport instrumentation in Sentry::Logging
This needs #2604.

Add support for the most common instrumentation events:

- [ ] ActiveRecord queries
- [ ] Rails requests
- [ ] ActiveJob events
- [ ] ???

solnic:
Adds log subscribers for common Rails components.

## Screenshots

Screenshot 2025-08-25 at 10 57 03

Screenshot 2025-08-25 at 10 56 58

Screenshot 2025-08-25 at 10 56 51

Closes #2605


sentry-ruby/lib/sentry/test_helper.rb
```diff
@@ -2,6 +2,8 @@
 
 module Sentry
   module TestHelper
+    module_function
```
sl0thentr0py, line 5: what is this for

sentry-ruby/lib/sentry/test_helper.rb
```diff
@@ -63,6 +62,14 @@ def teardown_sentry_test
       Sentry::Scope.global_event_processors.clear
     end
 
+    def clear_sentry_events
+      if Sentry.initialized? && Sentry.configuration.enable_logs
+        [Sentry.get_current_client.transport, Sentry.logger].each do |obj|
```
sl0thentr0py: this will only clear the transport if `enable_logs` is on

sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -201,6 +201,18 @@ def capture_exception_frame_locals=(value)
     # @return [String, nil]
     attr_accessor :sdk_debug_transport_log_file
 
+    # File path for DebugStructuredLogger to log events to. If not set, defaults to a temporary file.
+    # This is useful for debugging and testing structured logging.
+    # @return [String, nil]
+    attr_accessor :sdk_debug_structured_logger_log_file
```
sl0thentr0py, line 207: I do not like that we are polluting the main `configuration` namespace with this debug stuff, would prefer they be passed as init arguments to `DebugStructuredLogger.new` and instead of `structured_logger_class` we just have `structured_logger` where you can pass in your own instance.

sentry-rails/lib/sentry/rails/log_subscribers/parameter_filter.rb
```diff
@@ -0,0 +1,52 @@
...
+      #         filtered_params = filter_sensitive_params(event.payload[:params])
+      #         attributes[:params] = filtered_params unless filtered_params.empty?
+      #       end
+      #     end
+      #   end
+      module ParameterFilter
+        EMPTY_HASH = {}.freeze
+
+        if ::Rails.version.to_f >= 6.0
+          def self.backend
```
sl0thentr0py, line 28: these do not need to be methods, just make a class variable

sentry-rails/lib/sentry/rails/configuration.rb
```diff
@@ -159,6 +159,25 @@ class Configuration
       # Set this option to true if you want Sentry to capture each retry failure
       attr_accessor :active_job_report_on_retry_error
 
+      # Configuration for structured logging feature
+      # @return [StructuredLoggingConfiguration]
+      attr_reader :structured_logging
+
+      # Allow setting structured_logging as a boolean for convenience
+      # @param value [Boolean, StructuredLoggingConfiguration]
+      def structured_logging=(value)
```
sl0thentr0py: let's not have the boolean overloading, having the boolean `structured_logging.enabled` inside is enough

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ### Feature
 
 - Propagated sampling rates as specified in [Traces](https://develop.sentry.dev/sdk/telemetry/traces/#propagated-random-value) docs ([#2671](https://github.com/getsentry/sentry-ruby/pull/2671))
+- Support for Rails ActiveSupport log subscribers ([#2690](https://github.com/getsentry/sentry-ruby/pull/2690))
```
sl0thentr0py, line 6: need a detailed changelog entry for how to use this stuff

sentry-rails/lib/sentry/rails/configuration.rb
```diff
@@ -176,6 +195,22 @@ def initialize
...
+      # @return [Boolean]
+      attr_accessor :enabled
+
+      # Hash of components to subscriber classes for structured logging
+      # @return [Hash<Symbol, Class>]
+      attr_accessor :subscribers
+
+      def initialize
+        @enabled = false
+        @subscribers = {}
```
sl0thentr0py: when enabled is set to true, we should populate a set of sane subscribers by default

sentry-rails/lib/sentry/rails/log_subscribers/active_record_subscriber.rb
```diff
@@ -0,0 +1,144 @@
...
+            attributes[:db_name] = db_name
+          end
+
+          attributes[:server_address] = db_config[:host] if db_config[:host]
+          attributes[:server_port] = db_config[:port] if db_config[:port]
+          attributes[:server_socket_address] = db_config[:socket] if db_config[:socket]
+        end
+
+        if ::Rails.version.to_f >= 6.1
+          def extract_db_config_from_connection(connection)
```
sl0thentr0py, line 102: Logging is supposed to be _extremely_ low overhead because the volume is very high. If any of these branches are potentially expensive or incur an extra db connection, I would err on the side of leaving this data out in those cases instead of trying to be too smart to find it.

So basically, just take it simply if it exists otherwise let it not be there.

We already have a full fledged tracing product and if people want complete instrumentation, they are supposed to use Performance and not just Logging.

sl0thentr0py (changes requested):
mainly left comments on 
* overhead
* configuration api

some more high level feedback:
I would have much preferred this to be a simple proxy of just passing through to our logger whatever rails emits. Instead, now we have special loggers for each component. I will not block this if you are ok with maintaining all this extra code (I do not like maintaining a lot of code, my personal preference.)

Further, there is now a lot of duplication of logic between
* breadcrumbs
* logger
* tracing

To reiterate from a business point of view, Logging is mainly a high volume trace connected feed of events, whereas Tracing is a much more curated instrumentation experience and they both have their place. The way this PR is makes Logging very similar to Performance.

sentry-ruby/lib/sentry/configuration.rb
solnic: @sl0thentr0py are you OK with cleaning that up in a separate PR?

sentry-ruby/lib/sentry/test_helper.rb
solnic: @sl0thentr0py makes it possible to call methods as class methods too, or include them as RSpec example helpers. Very handy IMO.

sentry-ruby/lib/sentry/configuration.rb
sl0thentr0py: yes sure

sentry-rails/lib/sentry/rails/log_subscribers/parameter_filter.rb
solnic: actually that's not a good idea because it couples loading of this class with the presence of a specific module that may, or may not be loaded. Using methods is a safe-bet.

PR #2692, opened 2025-09-03
Clean up handling of debug/structured logger

solnic:
Follow-up after #2690


sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -201,28 +201,26 @@ def capture_exception_frame_locals=(value)
...
-    # @return [Class, nil]
-    attr_reader :structured_logger_class
-
-    # @deprecated Use {#sdk_logger=} instead.
+    # Set the logger instance for structured logging.
+    # @return [Sentry::StructuredLogger, Sentry::Logger]
     def logger=(logger)
-      warn "[sentry] `config.logger=` is deprecated. Please use `config.sdk_logger=` instead."
-      self.sdk_logger = logger
+      if enable_logs
```
sl0thentr0py: this is too confusing, we don't need a new setter on the config when `enable_logs` is on. It will cause unexpected behavior to existing `config.logger` setups if they simply turn on `enable_logs`.

solnic: @sl0thentr0py right :/ so I need to add a temporary `config.structured_logger` and we'll rename it to `logger` in the next major.

sl0thentr0py: no please nest all that you need under `config.structured_logging` not under the top level `config`

solnic:
@sl0thentr0py force-pushed because I rebuilt it after all, looks nice and clean :)

PR #2700, opened 2025-09-15
Skip sentry.message.template when there are no parameters

### LINKED ISSUE #2691: Make sure the Ruby SDK only set `sentry.message.template` on logs if `sentry.message.parameter` values are also defined
ref [https://github.com/getsentry/sentry-docs/pull/14759](https://github.com/getsentry/sentry-docs/pull/14759)

> If there are no `sentry.message.parameter.X` attributes included in the log, then the SDK MUST NOT attach a `sentry.message.template` attribute. This is important because it reduces duplicate content in the log and makes PII processing in Sentry much easier as only we don’t have to duplicate processing between the log message and the `sentry.message.template` attribute.

solnic:
Closes #2691 

#skip-changelog

### DIFF
--- a/sentry-ruby/lib/sentry/log_event.rb
+++ b/sentry-ruby/lib/sentry/log_event.rb
@@ -32,6 +32,8 @@ module Sentry
       "sentry.message.template" => :template
     }
 
+    PARAMETER_PREFIX = "sentry.message.parameter"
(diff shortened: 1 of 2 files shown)


dingsdax (approved):
👍

Issue #2698, opened 2025-09-11
Report Sidekiq errors via `Rails.error`


lucasmazza:
### Problem Statement

Our app includes both `sentry-rails` and `sentry-sidekiq`, and we moved a lot of our codebase to use the `Rails.error` API. 

We use `Rails.error.set_context` on a few places on our app (instead of `Sentry.with_scope`) to enrich reports with extra info, but when the exception is raised inside a background job it isn't included on the issue.

Would make sense to have some integration between those two gems to make `Sentry::Sidekiq.capture_exception` go through the Rails reporter instead of its default implementation?

dingsdax:
Thanks for raising this issue 👍 !

It makes sense imho to use a single way for error context across Rails and Sidekiq if the Rails.error API is used.
We'll add this to our roadmap and take a closer look.

PRs are always very welcome! If you'd like to tackle this 😊:
* take a look at [our contribution guide](https://github.com/getsentry/sentry-ruby/blob/master/CONTRIBUTING.md)
* I suggest to make it configurable option for starters

PR #2701, opened 2025-09-15
feat(sentry-sidekiq): add `report_errors_via_rails` config

lucasmazza:
## Description

Adds a `report_errors_via_rails` config to `sentry-sidekiq` to fix https://github.com/getsentry/sentry-ruby/issues/2698. When this is enabled, exceptions reported by the Sidekiq error handler will first go through `Rails.error` (and `Rails::Sentry::ErrorSubscriber`) before being sent to Sentry. 

This allow exceptions from Sidekiq jobs to use the same context from `Rails.error.set_context` or `Rails.error.add_middleware` (available on Rails 8.1+)

Fixes RUBY-92

### DIFF
--- a/sentry-sidekiq/lib/sentry/sidekiq/error_handler.rb
+++ b/sentry-sidekiq/lib/sentry/sidekiq/error_handler.rb
@@ -55,7 +55,7 @@ module Sentry
       private
 
       def report_error(ex, context, hint: {})
-        if Sentry.configuration.sidekiq.report_errors_via_rails
+        if Sentry.configuration.sidekiq.report_errors_via_rails && defined?(::Rails)
           mechanism = Sentry::Mechanism.new(type: "sidekiq", handled: false)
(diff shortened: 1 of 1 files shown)


lucasmazza:
I'm testing this with our Rails app but the solution isn't quite there yet - Rails reporter [is ignoring the exception](https://github.com/rails/rails/blob/0b5af58d2f0b8916f8211fa620c48da71d6da657/activesupport/lib/active_support/error_reporter.rb#L234) because the retry check has marked it as `@__rails_error_reported` somehow :/

solnic:
> Rails reporter [is ignoring the exception](https://github.com/rails/rails/blob/0b5af58d2f0b8916f8211fa620c48da71d6da657/activesupport/lib/active_support/error_reporter.rb#L234) because the retry check has marked it as `@__rails_error_reported` somehow :/

@lucasmazza did you manage to figure it out?

lucasmazza:
@solnic nope - I couldn't figure this out and changed our app code to stop using `Rails.error.set_context` and use `Sentry.set_context` directly.

PR #2709, opened 2025-09-19
2696 Respect log_level

### LINKED ISSUE #2696: Sentry logging doesn't respect log_level severity set in Rails
### Problem Statement

We recently enabled the new Sentry logger (using `config.enable_logs = true` and the `:logger` patch) for our Rails application.  We noticed that all logs are being forwarded to Sentry, regardless of the severity.  We'd love to use this feature but are not interested in forwarding the massive amount of debug and info logs.

With some digging, I noticed that the [patch introduced by the Sentry gem](https://github.com/getsentry/sentry-ruby/blob/master/sentry-ruby/lib/sentry/std_lib_logger.rb) intercepts the calls to the logger and simply forwards all logs without respect for the log level. Is there any interest in adding logic to respect the `log_level` configured? I would be happy dive into the code.

### Solution Brainstorm

By default Rails exposes a configuration option ([`log_level`](https://github.com/rails/rails/blob/b0c813bc7b61c71dd21ee3a6c6210f6d14030f71/railties/lib/rails/application/configuration.rb#L383)) that [sets the minimum severity on the `Rails.logger` which is created by default](https://github.com/rails/rails/blob/83aec03da6e5dcb0d64ac40c0a51ec3b9f06de29/railties/lib/rails/application/bootstrap.rb#L56-L63). I suggest that when the gem is used within a Rails application we respect this configuration.

solnic:
Fixes #2696

### DIFF
--- a/sentry-rails/spec/isolated/rails_logger_patch_spec.rb
+++ b/sentry-rails/spec/isolated/rails_logger_patch_spec.rb
@@ -15,20 +15,23 @@ require_relative "../dummy/test_rails_app/app"
 RSpec.describe "Rails.logger with :logger patch" do
   include Sentry::TestHelper
 
-  # Set up a real Rails app with logger
-  let(:log_output) { StringIO.new }
-  let(:app) do
-    make_basic_app do |config|
+  let!(:app) do
+    make_basic_app do |config, app|
       config.enable_logs = true
       config.enabled_patches = [:logger]
       config.max_log_events = 10
       config.sdk_logger = Logger.new(nil)
+
+      app.config.log_level = log_level
     end
   end
 
+  let(:log_level) { ::Logger::DEBUG }
+  let(:log_output) { StringIO.new }
+
   before do
-    app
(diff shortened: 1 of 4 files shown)


sentry-rails/spec/isolated/rails_logger_patch_spec.rb
```diff
@@ -15,20 +15,23 @@
...
     end
   end
 
+  let(:log_level) { ::Logger::DEBUG }
+  let(:log_output) { StringIO.new }
+
   before do
-    app
     Rails.logger = Logger.new(log_output)
+    Rails.logger.level = log_level
```
solnic, line 34: This is actually what Rails does under the hood, and uses configured `log_level`.

nbelzer:
Thank you @solnic, happy to report this is working as expected.

solnic:
> Thank you @solnic, happy to report this is working as expected.

I appreciate you letting me know 😄 🙇🏻

PR #2716, opened 2025-09-23
Default config.enabled_environments to nil instead of []

### LINKED ISSUE #2611: Sentry reports traces and errors even if enabled_environments is empty
### Issue Description

From the documentation and the name of the configuration option `enabled_environments` I was expecting that setting it to an empty array would disable sentry. Instead it gets enabled in all environments.

I worked around this by wrapping the whole `Sentry.init` block with my enabled-condition. Is this the expected way to disable sentry globally (for example via an environment variable)?

### Reproduction Steps

Configure Sentry with the configuration below. Start the Rails server and visit a page. 

### Expected Behavior

I expected it to not report in any environment. 

### Actual Behavior

However as the empty array is the default value of the configuration option, it reports in all environments.

### Ruby Version

3.4.3

### SDK Version

5.23.0

Rails (5.23.0)

### Sentry Config

```ruby
Sentry.init do |config|
  config.dsn = "..."
  config.enabled_environments = []
end
```

sl0thentr0py:
### Issues
* resolves #2611
* resolves RUBY-68

### DIFF
--- a/sentry-ruby/lib/sentry/configuration.rb
+++ b/sentry-ruby/lib/sentry/configuration.rb
@@ -149,7 +149,7 @@ module Sentry
     attr_reader :dsn
 
     # Whitelist of enabled_environments that will send notifications to Sentry. Array of Strings.
-    # @return [Array<String>]
+    # @return [Array<String>, nil]
     attr_accessor :enabled_environments
 
     # Logger 'progname's to exclude from breadcrumbs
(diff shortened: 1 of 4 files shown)


sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -647,7 +647,7 @@ def exception_class_allowed?(exc)
     end
 
     def enabled_in_current_env?
-      enabled_environments.empty? || enabled_environments.include?(environment)
+      enabled_environments.nil? || enabled_environments.include?(environment)
```
solnic, line 650: Just a remark for now but eventually it would be good to have a more explicit interface for checking if a setting is disabled, something like `config.enabled?(:enabled_environments)` or something like this, because nil checks are bug prone.

PR #2721, opened 2025-09-24
feat(logging): auto-enable structured_logging in Rails

### LINKED ISSUE #2715: Remove double opt-in for Logs in Rails
if  `enable_logs = true`  in Ruby, we need to enable Rails logging automatically

```
Sentry.init do |config|
  # ... your setup ...

  # Make sure structured logging is enabled
  config.enable_logs = true

  # Enable default Rails log subscribers (ActionController and ActiveRecord)
  config.rails.structured_logging.enabled = true
end
```

solnic:
When config.enable_logs is true, enable structured logging in Rails automatically too.

Closes #2715

### DIFF
--- a/sentry-rails/lib/sentry/rails/configuration.rb
+++ b/sentry-rails/lib/sentry/rails/configuration.rb
@@ -14,7 +14,7 @@ module Sentry
     attr_reader :rails
 
     after(:initialize) do
-      @rails = Sentry::Rails::Configuration.new(self)
+      @rails = Sentry::Rails::Configuration.new
       @excluded_exceptions = @excluded_exceptions.concat(Sentry::Rails::IGNORE_DEFAULT)
 
       if ::Rails.logger
@@ -170,7 +170,7 @@ module Sentry
       # @return [StructuredLoggingConfiguration]
       attr_reader :structured_logging
 
-      def initialize(parent_config = nil)
+      def initialize
         @register_error_subscriber = false
         @report_rescued_exceptions = true
         @skippable_job_adapters = []


solnic:
@sl0thentr0py feeling so-so about how I implemented it but it is the simplest thing that worked. The drawback is that I made `enabled` attribute reader private. A couple of alternatives that I tried involved a new callback in Configuration or overidding `enabled=` in Configuration, and both felt even worse 😅

sentry-rails/lib/sentry/rails/configuration.rb
```diff
@@ -14,7 +14,7 @@ class Configuration
     attr_reader :rails
 
     add_post_initialization_callback do
-      @rails = Sentry::Rails::Configuration.new
+      @rails = Sentry::Rails::Configuration.new(self)
```
sl0thentr0py: hmm this is too complicated just add in this callback
```
@rails.structured_logging.enabled = true if self.enable_logs
```

don't need to change the config implementation now.

solnic: @sl0thentr0py but this won't work, it was the first thing I tried actually. This callback runs in `Configuration#initialize` so it's *too early*. We could just introduce a new callback that runs after calling the block passed to `Sentry.init` that ie sets configuration default values.

sl0thentr0py: no that's actually how it's supposed to be, we set our defaults first, and then the user block runs at the end where they could then override whatever (including this one, if they want to disable)

solnic: @sl0thentr0py OK but `enable_logs` is disabled by default 🙃

sl0thentr0py: ok let's do other callback then

sentry-rails/lib/sentry/rails/configuration.rb
```diff
@@ -166,7 +170,7 @@ class Configuration
       # @return [StructuredLoggingConfiguration]
       attr_reader :structured_logging
 
-      def initialize
+      def initialize(parent_config = nil)
```
sl0thentr0py: don't need the `parent_config` anymore right?

sl0thentr0py (approved):
small nit/leftover, otherwise 🚢

PR #2723, opened 2025-09-25
[sidekiq] better and isolated rails specs

solnic:


sl0thentr0py:
We generally default enable integrations that are widely used. Sentry SDK design is not opt-in in most cases.

What errors specifically does the patch cause?

solnic:
> We generally default enable integrations that are widely used. Sentry SDK design is not opt-in in most cases.
> 
> What errors specifically does the patch cause?

@sl0thentr0py see recent commit. Basically after moving the conditional integration test to the isolated dir, for whatever reason, redis instrumentation started kicking in in specs where it previously did not, resulting in more spans in events.

sl0thentr0py (approved):
ty, makes sense!

PR #2725, opened 2025-09-25
Add `config.trace_ignore_status_codes` to control which status codes don't get traced

sl0thentr0py:
### Issues
* resolves: #2693, #2694 
* resolves: RUBY-89, RUBY-90

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ Gemfile.lock
 .idea
 *.rdb
 .rgignore
+.claude
 
 node_modules
 .vite
--- a/sentry-ruby/lib/sentry/configuration.rb
+++ b/sentry-ruby/lib/sentry/configuration.rb
@@ -328,6 +328,18 @@ module Sentry
     # @return [Array<String, Regexp>]
     attr_accessor :trace_propagation_targets
 
+    # Collection of HTTP status codes or inclusive ranges to ignore when tracing incoming requests.
+    # If a transaction's http.response.status_code matches one of these values,
+    # the transaction will be dropped and marked as not sampled.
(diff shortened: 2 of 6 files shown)


sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -328,6 +328,18 @@ def logger
...
     attr_accessor :trace_propagation_targets
 
+    # Collection of HTTP status codes or inclusive ranges to ignore when tracing incoming requests.
+    # If a transaction's http.response.status_code matches one of these values,
+    # the transaction will be dropped and marked as not sampled.
+    # Defaults to TRACE_IGNORE_STATUS_CODES_DEFAULT.
+    #
+    # @example
+    #   # ignore 404 and 502 <= status_code <= 511
+    #   config.trace_ignore_status_codes = [404, [502, 511]]
```
solnic: Why not use ranges? It'd be clearer.

.gitignore
```diff
@@ -13,6 +13,7 @@ Gemfile.lock
 .idea
 *.rdb
 .rgignore
+.claude
```
solnic, line 16: Not sure if we want this here?

solnic (approved):
Left a couple of minor questions. LGTM in general.

sentry-ruby/lib/sentry/configuration.rb
sl0thentr0py: good point, was just following the spec, will change!

.gitignore
sl0thentr0py: we do for now, later we might add repo-wide rules to git

PR #2739, opened 2025-10-07
Remove hub from Transaction#initialize

sl0thentr0py:


sentry-ruby/lib/sentry/utils/logging_helper.rb
```diff
@@ -3,27 +3,29 @@
...
     end
 
     # @!visibility private
     def log_warn(message)
-      sdk_logger.warn(LOGGER_PROGNAME) { message }
+      sdk_logger&.warn(LOGGER_PROGNAME) { message }
+    end
+
+    # @!visibility private
+    def sdk_logger
```
sl0thentr0py, line 27: generalized this to fallback to the global logger

sentry-ruby/lib/sentry/transaction.rb
```diff
@@ -141,7 +128,9 @@ def set_measurement(name, value, unit = "")
     # @param sampling_context [Hash] a context Hash that'll be passed to `traces_sampler` (if provided).
     # @return [void]
     def set_initial_sample_decision(sampling_context:)
-      unless @tracing_enabled
+      configuration = Sentry.configuration
+
+      unless configuration && configuration.tracing_enabled?
```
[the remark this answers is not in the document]
sl0thentr0py: this happens when the SDK is not init so intentional

sentry-ruby/lib/sentry/transaction.rb
```diff
@@ -207,24 +196,27 @@ def finish(end_timestamp: nil)
         @name = UNLABELD_NAME
       end
 
-      @hub.stop_profiler!(self)
+      hub = Sentry.get_current_hub
+      return unless hub
+
```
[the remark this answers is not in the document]
sl0thentr0py: intentional as above

sentry-ruby/lib/sentry/transaction.rb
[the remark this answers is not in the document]
sl0thentr0py: the SDK is off in this state so intentional

sentry-ruby/lib/sentry/utils/logging_helper.rb
```diff
@@ -3,27 +3,29 @@
...
-    # @!visibility private
-    attr_reader :sdk_logger
-
     # @!visibility private
     def log_error(message, exception, debug: false)
       message = "#{message}: #{exception.message}"
       message += "\n#{exception.backtrace.join("\n")}" if debug
 
-      sdk_logger.error(LOGGER_PROGNAME) do
+      sdk_logger&.error(LOGGER_PROGNAME) do
```
solnic, line 11: Why is the safe operator needed here? Sentry.sdk_logger is always set.

sentry-ruby/lib/sentry/transaction.rb
```diff
@@ -302,12 +305,13 @@ def populate_head_baggage
     end
 
     def ignore_status_code?
-      return false unless @trace_ignore_status_codes
+      trace_ignore_status_codes = Sentry.configuration&.trace_ignore_status_codes
```
solnic, line 308: How come configuration could be `nil`?

sl0thentr0py: https://github.com/getsentry/sentry-ruby/blob/7de03b3f6d1b93a9b1cbdefd3bcdd6295013feee/sentry-ruby/lib/sentry-ruby.rb#L165

sentry-ruby/lib/sentry/utils/logging_helper.rb
sl0thentr0py: if `configuration` is `nil`, that `Sentry.sdk_logger` will actually break right now, will also fix.

sl0thentr0py: will do this in a separate PR since I need to remove the old `logger` stuff too

PR #2742, opened 2025-10-08
Change default Stackprof profiler interval to 1000 microseconds

nateberkopec:
In using Sentry, I've found that the default profiler interval makes profiles useful only in very unusual situations. 10ms just isn't enough detail to figure out what's going on most of the time. In addition, it's so coarse-grained that it confuses newbies who don't understand how sampling profilers work and don't understand that each span may not _necessarily_ represent 10ms of execution time.

I have not seen StackProf generate more than 1-5% overhead most of the time, and profile sample rates on every Sentry app I've worked with are < 1% of total requests. I think the better tradeoff here is in favor of more useful profiles.

### DIFF
--- a/sentry-ruby/spec/sentry/profiler_spec.rb
+++ b/sentry-ruby/spec/sentry/profiler_spec.rb
@@ -55,7 +55,7 @@ RSpec.describe Sentry::Profiler, when: :stack_prof_installed? do
         subject.set_initial_sample_decision(true)
 
         expect(StackProf).to receive(:start).with(
-          interval: Profiler::DEFAULT_INTERVAL,
+          interval: 1000,
           mode: :wall,
           raw: true,
           aggregate: false


nateberkopec:
As far as I can tell, Vernier is running on the default interval, which is actually 500micros. I have much less experience with Vernier so I can't comment as to how much overhead that's adding, but in my conversations with John I think it's not greatly different than Stackprof.

sl0thentr0py:
thanks @nateberkopec, I'll have to verify with the ingestion folks if this is supported on the backend first.

As a side, the 101 is actually intentional to avoid 'lockstep sampling' as [mentioned in the note here](https://develop.sentry.dev/sdk/telemetry/profiles/sample-format-v1/#profile-data), so we'll have to do something similar for the new number too.

sl0thentr0py:
Alright, this is possible, but we'll make this an optional config that still defaults to 101, I'll ship a PR later this week for that. I'll need to verify that normal payloads aren't way too oversized otherwise the event will just be dropped by the server.

sl0thentr0py:
tracking in #2744

sl0thentr0py:
https://github.com/getsentry/sentry-ruby/pull/2745

Issue #2744, opened 2025-10-09
Make sampling frequency a config


sl0thentr0py:
Alright, this is possible, but we'll make this an optional config that still defaults to 101, I'll ship a PR later this week for that. I'll need to verify that normal payloads aren't way too oversized otherwise the event will just be dropped by the server.

_Originally posted by @sl0thentr0py in https://github.com/getsentry/sentry-ruby/issues/2742#issuecomment-3386234596_
            
* also seems vernier is not at 101, so we need to fix that

PR #2745, opened 2025-10-09
Add new config.profiles_sample_interval

sl0thentr0py:
### Issues

* resolves: #2744
* resolves: RUBY-104


nateberkopec:
This is a major change to the (unconfigured) sampling interval for Vernier, right?

sl0thentr0py:
correct, this will go on the 6.0 major, will call it out more explicitly in the changelog

PR #2752, opened 2025-10-16
Always set up the StructuredLogger and no-op capture_log_event if logs are disabled

sl0thentr0py:


sentry-ruby/lib/sentry/hub.rb
```diff
@@ -218,7 +218,7 @@ def capture_check_in(slug, status, **options)
     end
 
     def capture_log_event(message, **options)
-      return unless current_client
+      return unless current_client && current_client.configuration.enable_logs
```
solnic, line 221: Why is this needed now?

sl0thentr0py: because `Sentry.logger` is always set up to structured logger now and if we don't have this people have to do nil checks on `Sentry.logger` before using it otherwise.

`config.enable_logs` should be a top level on/off switch without people having to do 

```
Sentry.logger.info('bla') if Sentry.logger
```

sl0thentr0py: as a design principle, our top level APIs should always work with whether they do something or not decided on the hub/configuration level.

sentry-ruby/lib/sentry-ruby.rb
```diff
@@ -622,24 +622,9 @@ def continue_trace(env, **options)
...
-          warn <<~STR
-            [sentry] `Sentry.logger` will no longer be used as internal SDK logger when `enable_logs` feature is turned on.
-                    Use Sentry.configuration.sdk_logger for SDK-specific logging needs."
-
-                    Caller: #{caller.first}
-          STR
-
-          configuration.sdk_logger
-        end
+      @logger ||= configuration.structured_logging.logger_class.new(configuration)
```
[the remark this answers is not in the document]
sl0thentr0py: the whole SDK relies on Sentry.init being called before doing anything else

sentry-ruby/lib/sentry/hub.rb
solnic: But we always set the logger now, so is the intention here to suppress the logger via `enable_logs = false`?

sl0thentr0py: yes, otherwise logs would always get sent

solnic (approved):

PR #2775, opened 2025-11-14
fix(rails): properly skip silenced logs from AR

solnic:
This ensures that if `silenced` logging was used, the ActiveRecord log subscriber won't log silenced events.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - Optimize getting query source location in ActiveRecord tracing - this makes tracing up to roughly 40-60% faster depending on the use cases ([#2769](https://github.com/getsentry/sentry-ruby/pull/2769))
 
+### Bug fixes
+
+- Properly skip silenced `ActiveRecord::Base.logger`'s log entries in the ActiveRecord log subscriber ([#2775](https://github.com/getsentry/sentry-ruby/pull/2775))
+
 ## 6.1.0
 
 ### Features


sentry-rails/lib/sentry/rails/log_subscribers/active_record_subscriber.rb
```diff
@@ -30,6 +30,7 @@ class ActiveRecordSubscriber < Sentry::Rails::LogSubscriber
         # @param event [ActiveSupport::Notifications::Event] The SQL event
         def sql(event)
           return unless Sentry.initialized?
+          return if logger && !logger.info?
```
sl0thentr0py, line 33: this won't work if the user passes in a different severity to `silence`?
https://github.com/rails/rails/blob/14c115b120ed089331ff3dc13f36bd9129ced33d/activesupport/lib/active_support/logger_silence.rb#L16-L19

solnic: So, this PR is targeting default which is ERROR, and it should solve the solid_queue silencing being ignored issue. Handling levels is actually a bigger task, because we just default to `info` right now.

Rocela-DV:
Hi @solnic, thanks for this fix, I still see this error coming from sentry after updating to 6.1.1 
wonder if this still an issue 
image

solnic:
@Rocela-DV thanks for reporting this, I created #2794 to track it and will be fixing it today, hopefully for good this time 😄

solnic:
> Hi @solnic, thanks for this fix, I still see this error coming from sentry after updating to 6.1.1 wonder if this still an issue image

wait, we fixed it in 6.1.2 - please upgrade and let me know if it goes away :)

Rocela-DV:
@solnic you are correct sorry for the mislead, issue resolve in my end. Thanks!

PR #2777, opened 2025-11-17
fix(logs): skip malformed utf-8 messages

### LINKED ISSUE #2773: Invalid UTF-8 for logs
### Issue Description

Sending invalid UTF-8 results in logs not being sent.  Same situation only for logs; previously covered in https://github.com/getsentry/sentry-ruby/issues/2376.

### Reproduction Steps

Send invalid UTF-8 logs.

### Expected Behavior

Should be sent or have better handling.

### Actual Behavior

Instead it fails hard.

### Ruby Version

3.3.0

### SDK Version

6.0

_No response_

### Sentry Config

_No response_

solnic:
Fixes #2773

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Bug fixes
 
 - Properly skip silenced `ActiveRecord::Base.logger`'s log entries in the ActiveRecord log subscriber ([#2775](https://github.com/getsentry/sentry-ruby/pull/2775))
+- Skip malformed utf-8 messages ([#2777](https://github.com/getsentry/sentry-ruby/pull/2777))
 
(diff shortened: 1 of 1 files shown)


sentry-ruby/lib/sentry/utils/encoding_helper.rb
```diff
@@ -3,6 +3,8 @@
 module Sentry
   module Utils
     module EncodingHelper
+      EMPTY_STRING = ""
```
chaadow, line 6: hi, just wondering where this constant is used?

solnic: right we no longer need this, thanks I'll remove it!

PR #2780, opened 2025-11-18
fix(logs): don't choke on malformed attributes

solnic:
Follow up after #2777 

#skip-changelog

### DIFF
--- a/sentry-ruby/spec/sentry/structured_logger_spec.rb
+++ b/sentry-ruby/spec/sentry/structured_logger_spec.rb
@@ -113,6 +113,22 @@ RSpec.describe Sentry::StructuredLogger do
             expect(log_event[:body]).to eql(malformed_string_default)
           end
 
+          it "doesn't choke on malformed UTF-8 in body and attributes" do
+            malformed_string = "Hello World %{user_agent} \x92".dup.force_encoding("UTF-8")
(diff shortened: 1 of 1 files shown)


sentry-ruby/lib/sentry/client.rb
```diff
@@ -199,7 +199,17 @@ def event_from_log(message, level:, **options)
 
       return unless body
 
-      LogEvent.new(level: level, body: body, attributes: attributes, origin: origin)
+      sanitized_attributes = attributes.transform_values do |value|
+        if value.is_a?(String)
+          return unless (sanitized_string = Utils::EncodingHelper.safe_utf_8_string(value))
```
solnic: @sl0thentr0py WDYT about replacing malformed values with something like `<malformed utf string>` or something like this instead of silently dropping? Feels like a reasonable thing to do. Same re dropping if the body is malformed.

sl0thentr0py: fine with me

sentry-ruby/lib/sentry/client.rb
```diff
@@ -197,9 +197,15 @@ def event_from_log(message, level:, **options)
...
+      sanitized_attributes = attributes.transform_values do |value|
+        if value.is_a?(String)
+          Utils::EncodingHelper.safe_utf_8_string(value)
+        else
+          value
+        end
+      end
 
-      LogEvent.new(level: level, body: body, attributes: attributes, origin: origin)
+      LogEvent.new(level: level, body: body, attributes: sanitized_attributes, origin: origin)
```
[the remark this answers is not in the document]
solnic: no, I changed it to return a default value which is a string

PR #2789, opened 2025-11-24
fix(rails): fix logging of cached queries with binds

### LINKED ISSUE #2785: `ActiveRecordSubscriber` fails to log most queries on Rails 8.0.3, everywhere except JRUBY
### Issue Description

Since upgrading to sentry-ruby 6.1, instead of query logs we mostly get these:

Image

At a glance, this is because this line accesses `type_casted_binds`, but only in JRUBY does `type_casted_binds` return a Hash.

Image

Everywhere else (unless one is using an unusual database adapter that overloads `cache_notification_info`), `type_casted_binds` returns a lambda, and since `Proc#[]` is an alias for `Proc#call`, the attempt to access a value under a hash key on :57 actually calls the proc with the key as the unexpected argument.

Image

### Reproduction Steps

1. Install sentry-rails 6.1 on a Rails 8.0.3 app on MRI (with postgres, or probably almost anything)
2. Configure `Sentry.configuration.send_default_pii = true`
3. Attempt to run any normal query with at least one bound parameter, e.g., `Account.where(email: "foo@example.com").first`

### Expected Behavior

Logs query

### Actual Behavior

Logs its own failure to log the query

### Ruby Version

3.3.2

### SDK Version

sentry-rails 6.1.0 (not sure what this is asking)

Rails 8.0.3

### Sentry Config

```ruby
Sentry.init do |config|
  config.dsn = ENV["SENTRY_EU_DSN"]

  if ENV["SENTRY_LOGGING_ENABLED"]
    config.enable_logs = true
    config.enabled_patches = [:logger]
    config.rails.structured_logging.enabled = true

    config.rails.structured_logging.subscribers = {
      action_controller: Sentry::Rails::LogSubscribers::ActionControllerSubscriber,
      active_record: Class.new(Sentry::Rails::LogSubscribers::ActiveRecordSubscriber) do
        def sql(event)
          # SolidQueue database queries are too noisy
          return if event.payload[:name]&.starts_with?("SolidQueue::")
          super
        end
      end
    }
  end

  # TODO: If/when `sentry-rails` adds solid_queue or ActiveJob support for not
  # reporting exceptions from retried jobs (as it does for Sidekiq and DelayedJob),
  # use that and stop excluding `CompletedDocument::MissingFolder`
  config.excluded_exceptions += ["CompletedDocument::MissingFolder"]

    true
  end

  config.profiles_sample_rate = 1.0

  config.release = ENV.fetch("HEROKU_RELEASE_VERSION", "")
end
```

solnic:
Fixes #2785

### DIFF
--- a/sentry-rails/spec/sentry/rails/log_subscribers/active_record_subscriber_spec.rb
+++ b/sentry-rails/spec/sentry/rails/log_subscribers/active_record_subscriber_spec.rb
@@ -147,7 +147,7 @@ RSpec.describe Sentry::Rails::LogSubscribers::ActiveRecordSubscriber do
           expect(log_event).not_to be_nil
         end
 
-        it "includes cached flag when query is cached", skip: Rails.version.to_f < 5.1 ? "Rails 5.0.0 doesn't include cached flag in sql.active_record events" : false do
+        it "includes cached flag when query is cached" do
           post = Post.create!(title: "test")
 
           Sentry.get_current_client.flush


sentry-rails/lib/sentry/rails/log_subscribers/active_record_subscriber.rb
```diff
@@ -50,7 +50,7 @@ def sql(event)
 
           binds = event.payload[:binds]
 
-          if Sentry.configuration.send_default_pii && binds&.any?
+          if Sentry.configuration.send_default_pii && (binds && !binds.empty?)
```
solnic, line 53: More verbose but actually faster

sentry-rails/lib/sentry/rails/log_subscribers/active_record_subscriber.rb
```diff
@@ -75,13 +75,15 @@ def sql(event)
...
-        else
-          def type_casted_binds(event)
-            event.payload[:type_casted_binds]
+        def type_casted_binds(event)
+          binds = event.payload[:type_casted_binds]
+
+          # When a query is cached, binds are a callable,
+          # and under JRuby they're always a callable.
+          if binds.respond_to?(:call)
+            binds.call
```
sl0thentr0py, line 84: does that mean in the cached case, we will incur extra overhead to fetch the binds?

solnic: @sl0thentr0py a bit, but it's hard to tell what the impact is w/o measuring it. Want me to check?

sl0thentr0py: would be great!

solnic: @sl0thentr0py we good amigo 😄 
image

sl0thentr0py: thanks for checking!

sl0thentr0py: thanks for checking!

Issue #2806, opened 2025-12-05
Sentry capture_exception blows up on `[]` backtrace


RafaelPAndrade:
### Issue Description

Observed that an exception thrown by an unrelated SDK was blowing up on our Sentry reporting. Narrowed down to this line exploding: https://github.com/getsentry/sentry-ruby/blob/6c0ac11ea294f50ace2bb2576ea74724dba5a46f/sentry-ruby/lib/sentry/interfaces/single_exception.rb#L63

when the backtrace is `[]` (but not on `nil`)

### Reproduction Steps

```.ruby
begin
  raise StandardError, 'Stuff', []
rescue => e
   Sentry.capture_exception(e)
end
```

### Expected Behavior

It does not blow up.

### Actual Behavior

```
(XXXX):13:in '<main>': undefined method 'vars=' for nil (NoMethodError)

        stacktrace.frames.last.vars = locals
                              ^^^^^^^
/app/vendor/bundle/ruby/3.4.0/gems/irb-1.15.1/lib/irb.rb:406:in 'full_message': Stuff (StandardError)
```

backtrace (trimmed for relevance): 

```
["/app/vendor/bundle/ruby/3.4.0/gems/sentry-ruby-6.1.0/lib/sentry/interfaces/single_exception.rb:63:in 'Sentry::SingleExceptionInterface.build_with_stacktrace'",
 "/app/vendor/bundle/ruby/3.4.0/gems/sentry-ruby-6.1.0/lib/sentry/interfaces/exception.rb:36:in 'block in Sentry::ExceptionInterface.build'",
 "/app/vendor/bundle/ruby/3.4.0/gems/sentry-ruby-6.1.0/lib/sentry/interfaces/exception.rb:33:in 'Array#map'",
 "/app/vendor/bundle/ruby/3.4.0/gems/sentry-ruby-6.1.0/lib/sentry/interfaces/exception.rb:33:in 'Sentry::ExceptionInterface.build'",
 "/app/vendor/bundle/ruby/3.4.0/gems/sentry-ruby-6.1.0/lib/sentry/error_event.rb:35:in 'Sentry::ErrorEvent#add_exception_interface'",
 "/app/vendor/bundle/ruby/3.4.0/gems/sentry-ruby-6.1.0/lib/sentry/client.rb:133:in 'block in Sentry::Client#event_from_exception'",
 "<internal:kernel>:91:in 'Kernel#tap'",
[...]

PR #2807, opened 2025-12-05
Handle empty frames case gracefully with local vars

sl0thentr0py:
* resolves: #2806 
* resolves: RUBY-124


sl0thentr0py:
sidekiq is failing because of https://github.com/mperham/connection_pool/issues/210

Issue #2805, opened 2025-12-02
exception happened in background worker: source sequence is illegal/malformed utf-8


grosser:
### Issue Description

Something fails in production and all I see in the logs is this message,
which is not helpful since I need the original backtrace and error message to begin debugging

### Reproduction Steps

raise an error that includes invalid utf8

### Expected Behavior

`<?>` icons replace invalid utf8

### Actual Behavior

error not sent to sentry and log shows message

### Ruby Version

3.4.6

### SDK Version

5.28.0

### Integration and Its Version

none

### Sentry Config

nothing special, except a custom logger that does `{message: message}.to_json`

solnic:
@grosser thanks for the report. Any chance you could give me an example message that causes this? We do have handling of malformed utf-8 strings in place.

grosser:
sadly did not find more info and don't have a way to reliably reproduce it
feel free to close if you don't see any obvious spot where this could have slipped through

solnic:
@grosser alright, I'll dig into it - could you tell me what exact custom logger do you use that does `{message: message}.to_json`?

grosser:
standard logger with a formatter that does:
```
def call(level, _, _, m)
  message = {message: m.message, backtrace: m.backtrace.first(5).join("\n")} if m.is_a?(Exception)
  message = {message: m} if m.is_a?(String)
  {severity: level, worker: @worker}.merge(message).to_json << "\n"
end
```

solnic:
@grosser how exactly do you configure this logger in Sentry config?

grosser:
```
class JsonLogger
  def call(level, _, _, m)
    message = m
    message = {message: m.message, backtrace: m.backtrace.first(5).join("\n")} if m.is_a?(Exception)
    message = {message: m} if m.is_a?(String)
    {severity: level}.merge(message).to_json << "\n"
  end

  def self.build(out)
    logger = Logger.new(out, level: Logger::INFO)
    logger.formatter = new
    logger
  end
end

Sentry.init do |config|
  config.sdk_logger = JsonLogger.build($stdout)
  ...
end
```

solnic:
@grosser alright! So I suspect that is your culprit, because Sentry is crashing trying to use its own logger. This is because:

```ruby
irb(main):001> { message: "Hello World\x92".dup.force_encoding("UTF-8") }.to_json
(irb):1:in '<main>': undefined method 'to_json' for an instance of Hash (NoMethodError)
[...]

PR #2817, opened 2026-01-02
fix: prevent SDK crash when SDK logging fails 

dingsdax:
Add error handling to LoggingHelper methods to fallback to stderr when SDK loggers fail (e.g., JSON formatters with invalid UTF-8 in custom SDK logger). Prevents SDK from crashing while maintaining error visibility.

* resolves: #2805 
* resolves: RUBY-123

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,21 @@
 
 ### Features
 
+- Implement new `Sentry.metrics` functionality ([#2818](https://github.com/getsentry/sentry-ruby/pull/2818))
+
+  The SDK now supports Sentry's new [Trace Connected Metrics](https://docs.sentry.io/product/explore/metrics/) product.
+
+  ```ruby
+   Sentry.init do |config|
+     # ...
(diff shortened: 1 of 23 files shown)


sentry-ruby/lib/sentry/utils/logging_helper.rb
```diff
@@ -6,26 +6,37 @@ module LoggingHelper
...
     end
 
     # @!visibility private
     def sdk_logger
       @sdk_logger ||= Sentry.sdk_logger
     end
+
+    # @!visibility private
+    def log_to_stderr(error, message)
+      $stderr.puts("Sentry SDK logging failed (#{error.class}: #{error.message}): #{message}".scrub(%q(<?>)))
```
dingsdax: 🤔 output stacktrace maybe too?

solnic: @dingsdax yes I think that would be very helpful

sl0thentr0py (approved):

Issue #2820, opened 2026-01-05
Ensure maximum cap on logs and metrics buffers


sl0thentr0py:
Currently we flush after 100, but also have a 10k upper bound after which we drop completely

PR #2831, opened 2026-01-13
Max drop limit for TelemetryEventBuffer after which we completely drop events

sl0thentr0py:
## Description

* This is mainly to protect from memory blowup in high throughput scenarios
* We record `queue_overflow` client reports in this case
* Also raised the Metric buffer limits to 1k/10k as compared to 100/1k for logs (consistent with Python)

## Issues

* resolves: #2820
* resolves: RUBY-132

### DIFF
--- a/sentry-ruby/spec/sentry/configuration_spec.rb
+++ b/sentry-ruby/spec/sentry/configuration_spec.rb
@@ -770,7 +770,7 @@ RSpec.describe Sentry::Configuration do
   end
 
   describe "#max_metric_events" do
-    it "returns 100 by default" do
+    it "returns 1000 by default" do
       expect(subject.max_metric_events).to eq(1000)
     end
 


sentry-ruby/lib/sentry/telemetry_event_buffer.rb
```diff
@@ -53,10 +55,16 @@ def flush
...
+      log_debug("[#{self.class}] expected a #{@event_class}, got #{item.class}") unless item.is_a?(@event_class)
 
       @mutex.synchronize do
-        @pending_items << item
+        if size >= @max_items_before_drop
+          log_debug("[#{self.class}] exceeded max capacity, dropping event")
+          @client.transport.record_lost_event(:queue_overflow, @data_category)
+        else
+          @pending_items << item
+        end
```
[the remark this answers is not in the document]
sl0thentr0py: intentional, we should not raise in our apis, just silently fail with debugging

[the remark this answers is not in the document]
sl0thentr0py: ah wait i need to return, nvm

sentry-ruby/spec/support/shared_examples_for_telemetry_event_buffers.rb
```diff
@@ -67,6 +67,49 @@
     end
   end
 
+  describe "max capacity and dropping events" do
+    let(:max_items) { 3 }
+    let(:max_items_before_drop) { 10 }
+
+    before do
+      subject.instance_variable_set(:@max_items_before_drop, max_items_before_drop)
```
solnic, line 75: It would be cleaner to use `event_factory` to set up the subject instead of reaching for internals.

sentry-ruby/spec/sentry/configuration_spec.rb
```diff
@@ -771,7 +771,7 @@ class SentryConfigurationSample < Sentry::Configuration
 
   describe "#max_metric_events" do
     it "returns 100 by default" do
```
solnic:
```suggestion
    it "returns 1000 by default" do
```

sentry-ruby/spec/support/shared_examples_for_telemetry_event_buffers.rb
sl0thentr0py: event factory is for the events, not sure how you mean?

solnic: @sl0thentr0py ah sorry, of course, this should use buffer constructor

PR #2832, opened 2026-01-13
Lazily start TelemetryEventBuffer threads

sl0thentr0py:
* resolves: #2821
* resolves: RUBY-133

### DIFF
--- a/sentry-ruby/lib/sentry/telemetry_event_buffer.rb
+++ b/sentry-ruby/lib/sentry/telemetry_event_buffer.rb
@@ -50,9 +50,9 @@ module Sentry
     alias_method :run, :flush
 
     def add_item(item)
-      return unless ensure_thread
-
       @mutex.synchronize do
+        return unless ensure_thread
+
         if size >= @max_items_before_drop
           log_debug("[#{self.class}] exceeded max capacity, dropping event")
           @client.transport.record_lost_event(:queue_overflow, @data_category)


sentry-ruby/lib/sentry/telemetry_event_buffer.rb
```diff
@@ -55,6 +50,8 @@ def flush
     alias_method :run, :flush
 
     def add_item(item)
+      return unless ensure_thread
```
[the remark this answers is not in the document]
sl0thentr0py: that's fine

sentry-ruby/spec/support/shared_examples_for_telemetry_event_buffers.rb
```diff
@@ -24,6 +24,26 @@
...
+
+      expect(subject.thread).to be_a(Thread)
+    end
+
+    it "spawns only one thread" do
+      expect do
+        subject.add_item(event)
+      end.to change { Thread.list.count }.by(1)
+
+      expect(subject.thread).to receive(:alive?).and_return(true)
```
solnic, line 32: @sl0thentr0py is this really needed?

sentry-ruby/spec/support/shared_examples_for_telemetry_event_buffers.rb
```diff
@@ -24,6 +24,26 @@
   describe "#add_item" do
     let(:max_items) { 3 }
 
+    it "spawns new thread" do
+      expect do
+        subject.add_item(event)
+      end.to change { Thread.list.count }.by(1)
+
+      expect(subject.thread).to be_a(Thread)
+    end
```
solnic: This one seems redundant, the next one tests the same thing

sentry-ruby/lib/sentry/telemetry_event_buffer.rb
solnic: Super nit but maybe move this below the item class check so that in this (very unlikely case) that we get an invalid item type, at least we won't be spawning a thread.

sentry-ruby/spec/support/shared_examples_for_telemetry_event_buffers.rb
sl0thentr0py: just copied over from the session flusher, it's fine

sentry-ruby/spec/support/shared_examples_for_telemetry_event_buffers.rb
sl0thentr0py: removed

sentry-ruby/lib/sentry/telemetry_event_buffer.rb
sl0thentr0py: will remove but in general nit about stuff that never happens is not necessary as a review principle

sl0thentr0py: that check itself is not necessary since we control all the call sites, so I'm going to just remove that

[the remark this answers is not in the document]
sl0thentr0py: moved inside mutex

Issue #2840, opened 2026-01-20
Add OTLP support


sl0thentr0py:
[OTLP Integration](https://develop.sentry.dev/sdk/telemetry/traces/otlp/)

PR #2841, opened 2026-01-21
Add external_propagation_context support

sl0thentr0py:
## Description

Needed for OTLP Integration.

* part of: #2840 
* part of: RUBY-140

### DIFF
--- a/sentry-ruby/lib/sentry-ruby.rb
+++ b/sentry-ruby/lib/sentry-ruby.rb
@@ -666,6 +666,38 @@ module Sentry
       META
     end
 
+    # Registers a callback function that retrieves the current external propagation context.
+    # This is used by OpenTelemetry integration to provide trace_id and span_id from OTel context.
+    #
+    # @param callback [Proc, nil] A callable that returns [trace_id, span_id] or nil
+    # @return [void]
+    #
+    # @example
+    #   Sentry.register_external_propagation_context do
+    #     span_context = OpenTelemetry::Trace.current_span.context
+    #     return nil unless span_context.valid?
+    #     [span_context.hex_trace_id, span_context.hex_span_id]
+    #   end
+    def register_external_propagation_context(&callback)
+      @external_propagation_context_callback = callback
+    end
+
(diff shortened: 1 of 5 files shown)


sentry-ruby/lib/sentry/scope.rb
```diff
@@ -305,6 +294,33 @@ def get_span
...
+    def get_dynamic_sampling_context
+      if span
+        span.get_dynamic_sampling_context
+      elsif Sentry.get_external_propagation_context
+        # can't do dynamic sampling for otel/otlp yet
+        {}
+      else
+        propagation_context.get_dynamic_sampling_context
+      end
+    end
```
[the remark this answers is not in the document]
sl0thentr0py: fixed by fetching both together

sentry-ruby/lib/sentry-ruby.rb
```diff
@@ -666,6 +666,38 @@ def sdk_meta
       META
     end
 
+    # Registers a callback function that retrieves the current external propagation context.
```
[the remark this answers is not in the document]
sl0thentr0py: fine since it would just overwrite later, this is mainly for internal use

sentry-ruby/lib/sentry/scope.rb
```diff
@@ -94,7 +85,7 @@ def apply_to_event(event, hint = nil)
     # @return [MetricEvent, LogEvent] the telemetry event with scope context applied
     def apply_to_telemetry(telemetry)
       # TODO-neel when new scope set_attribute api is added: add them here
-      trace_context = span ? span.get_trace_context : propagation_context.get_trace_context
+      trace_context = get_trace_context
```
solnic, line 88: This seems to be an overkill because we're building up a bunch of hashes just to get trace_id and span_id - it would be better to have something more dedicated to get these two values.

sl0thentr0py: its one hash, its fine because all these values belong together, they are not separate

sl0thentr0py: its one hash, its fine because all these values belong together, they are not separate

PR #2848, opened 2026-02-02
WIP: better NotFound error handling in rails tracing

### LINKED ISSUE #2825: Transaction status incorrectly reports 500 (internal_error) instead of actual HTTP status code (eg. 404) when Rails handles exceptions
### Issue Description

When Rails handles an exception internally and converts it to a specific HTTP status code (e.g., `ActiveRecord::RecordNotFound` -> 404), the Sentry transaction is incorrectly reported with status `internal_error` and `http.response.status_code: 500` instead of the actual HTTP response status code.

This creates inaccurate reports, as all Rails-rescued exceptions appear as 500 errors in Sentry, even though the client receives the correct HTTP status code (404, 422, etc.).

The breadcrumbs in the transaction correctly show the actual status code (404), but the transaction-level status and span-level status codes are incorrect (500).

Even though version 6.0.0+ includes `trace_ignore_status_codes` which by default excludes 404s from being traced, this is a workaround rather than a fix. The underlying issue of incorrect status code reporting remains for any status codes that are not ignored.

### Reproduction Steps

1. Configure a Rails application with `sentry-rails` version `5.27.0`
2. Enable tracing with `config.traces_sample_rate = 1.0`
3. Create a route that raises `ActiveRecord::RecordNotFound` or similar exception that Rails rescues and converts to a 404 response
4. Make a request to that route
5. Examine the transaction in Sentry

### Expected Behavior

- Span status should be "not_found"
- Span `data.http.response.status_code` should be 404

### Actual Behavior

- Span status is "internal_error"
- Span `data.http.response.status_code` is 500

### Ruby Version

Ruby 3.4.4

### SDK Version

sentry-ruby: 5.27.0

_No response_

### Sentry Config

_No response_

solnic:
WIP

Closes #2825

### DIFF
--- a/sentry-rails/spec/sentry/rails/tracing_spec.rb
+++ b/sentry-rails/spec/sentry/rails/tracing_spec.rb
@@ -59,6 +59,23 @@ RSpec.describe Sentry::Rails::Tracing, type: :request do
       expect(second_span[:timestamp] - second_span[:start_timestamp]).to be_between(min_duration, max_duration)
     end
 
+    it "records correct status when Rails rescues exception to non-500 status" do
+      get "/posts/999999"
+
+      expect(response).to have_http_status(:not_found)
+      expect(transport.events.count).to eq(1)
+
+      transaction = transport.events.last.to_h
+
+      expect(transaction[:type]).to eq("transaction")
+
+      first_span = transaction[:spans][0]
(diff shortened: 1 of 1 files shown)


solnic:
Closing in fav of https://github.com/getsentry/sentry-ruby/pull/2850

PR #2855, opened 2026-02-17
docs: update README to reflect Ruby 4.0 testing support

mjq:
Update the Requirements section to indicate that we test up to Ruby 4.0 instead of Ruby 3.4, reflecting the current state of our test matrix.

### DIFF
--- a/sentry-ruby/README.md
+++ b/sentry-ruby/README.md
@@ -33,7 +33,7 @@ If you're using `sentry-raven`, we recommend you to migrate to this new SDK. You
 
 ## Requirements
 
-We test from Ruby 2.4 to Ruby 3.4 at the latest patchlevel/teeny version. We also support JRuby 9.0.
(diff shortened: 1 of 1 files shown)


sentry-ruby/README.md
```diff
@@ -33,7 +33,7 @@ If you're using `sentry-raven`, we recommend you to migrate to this new SDK. You
 
 ## Requirements
 
-We test from Ruby 2.4 to Ruby 3.4 at the latest patchlevel/teeny version. We also support JRuby 9.0.
+We test from Ruby 2.4 to Ruby 4.0 at the latest patchlevel/teeny version. We also support JRuby 9.0.
```
mjq, line 36: Given that our test matrix is dynamic and we always test up to the latest released Ruby, I'm tempted to reword to say that explicitly rather than state the version + need an annual update.

Issue #2857, opened 2026-02-19
Trace metrics are always dropped by Relay with "No Data" reason in admin


Kobby-Bawuah:
### Issue Description

Metrics sent via `Sentry::Metrics.distribution` (or any metric method) never show up in the Sentry UI. In admin, they all appear as Dropped (Server) / No Data.

The timestamp in the metric payload is being sent as a string like `"2026-02-18 21:54:48 UTC"` instead of a numeric value like `1771453701.042`. Relay can't parse that format, so it rejects the metric entirely.

`LogEvent` already handles this correctly by converting the timestamp with `.to_f`.

### Reproduction Steps

1. Initialize Sentry with defaults
2. Send any metric:
`
Sentry::Metrics.distribution("test_metric", 20.0, unit: "kilobyte", attributes: { foo: "bar" })
`
3. Check the Metrics Explore page, nothing appears
4. Admin stats show all metrics as Dropped (Server) / No Data

### Expected Behavior

Metrics are accepted and visible in the Sentry UI.

### Actual Behavior

All metrics are dropped. The timestamp goes over the wire in a format Relay can't parse. The Python and JavaScript SDKs send timestamps as numbers and work fine.

### Ruby Version

3.4.0

### SDK Version

6.3.1

### Sentry Config

Sentry.init do |config|
config.dsn = 'https://...'
config.sample_rate = 1
config.traces_sampler = -> (c) do ... end
end

tnir:
Thanks @Kobby-Bawuah for filing this.

sl0thentr0py:
I've created a reproduction for this issue: https://github.com/getsentry/repro/pull/24

The reproduction confirms the bug: `MetricEvent#to_h` serializes the timestamp as a raw `Time` object, which becomes a string like `"2026-02-23 11:56:29 UTC"` in JSON. Relay expects a numeric Unix timestamp. `LogEvent#to_h` already handles this correctly with `timestamp.to_f`.

You can run it by cloning the repo and following the instructions in the README.

sl0thentr0py:
the older serialization also seems to work but `to_f` is indeed safer so will make the change anyway.

PR #2862, opened 2026-02-23
Fix MetricEvent timestamp serialization to float

sl0thentr0py:
* resolves: #2857 
* resolves: RUBY-147


sl0thentr0py:
vernier problem is unrelated and is in progress upstream

Issue #2873, opened 2026-02-27
Request queue time not tracked in rails


davekaro:
### Issue Description

In https://github.com/getsentry/sentry-ruby/pull/2838, queue time is captured for Rack apps. I am using the newest gem version, but noticed I couldn't find any data for the request queue time. Looking at the code, I think it's because the capture is done in https://github.com/getsentry/sentry-ruby/blob/faa2853052ba3c7ddbf9ab095b9a7162fe8e1225/sentry-ruby/lib/sentry/rack/capture_exceptions.rb#L66-L83, but since I'm running a Rails app, my rack app uses the subclass that overrides that method here https://github.com/getsentry/sentry-ruby/blob/faa2853052ba3c7ddbf9ab095b9a7162fe8e1225/sentry-rails/lib/sentry/rails/capture_exceptions.rb#L38-L52 and never captures the queue time.

### Reproduction Steps

Install the latest sentry-ruby gem, configured in a Rails app, look for `http.server.request.time_in_queue` in the tracing.

### Expected Behavior

I should see `http.server.request.time_in_queue` in the tracing.

### Actual Behavior

I can't find `http.server.request.time_in_queue` in the tracing.

### Ruby Version

4.0.1

### SDK Version

6.4.0

dingsdax:
thx for reporting 🙏,, we could independently [reproduce the issue](https://github.com/getsentry/repro/pull/27) & will release a fixed version soon.

PR #2874, opened 2026-02-27
Track request queue time in sentry-rails

davekaro:
Ensure the default behavior of start_transaction from the parent class `Sentry::Rack::CaptureExceptions` is executed when using Rails.

Resolves https://github.com/getsentry/sentry-ruby/issues/2873

However, I have not tested this at all. I opened this more as a starting point and possible solution. Note that prior to https://github.com/getsentry/sentry-ruby/pull/1405, the `start_transaction` method did call `super`. I don't have strong enough familiarity with the code to know whether this change makes sense or if another solution is better.

### DIFF
--- a/sentry-rails/lib/sentry/rails/capture_exceptions.rb
+++ b/sentry-rails/lib/sentry/rails/capture_exceptions.rb
@@ -36,19 +36,11 @@ module Sentry
       end
 
       def start_transaction(env, scope)
-        options = {
-          name: scope.transaction_name,
-          source: scope.transaction_source,
-          op: transaction_op,
-          origin: SPAN_ORIGIN
-        }
-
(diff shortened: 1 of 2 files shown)


dingsdax:
Thanks for digging into this @davekaro! One issue with the super approach though: SPAN_ORIGIN in the parent method resolves lexically to "auto.http.rack" regardless of the subclass calling it, so Rails transactions would silently get the wrong origin.        

We opened [#2877](https://github.com/getsentry/sentry-ruby/pull/2877) with a slightly different take, extracting `attach_queue_time` as a private method so the Rails subclass can keep its own explicit start_transaction and correct SPAN_ORIGIN. Cheers!

davekaro:
Sounds great, thanks for the quick fix!

PR #2885, opened 2026-03-05
chore(transport): Handle HTTP 413 response for oversized envelopes

sl0thentr0py:
## Description

Handle HTTP 413 (Content Too Large) responses from Relay. Previously, oversized envelopes returned HTTP 400 from Relay. Now that Relay returns 413, the SDK can distinguish size-related rejections from other errors.

**Changes:**
- When the transport receives an HTTP 413 response, log a warning: "HTTP 413: Envelope dropped due to exceeded size limit" (with response body if available)
- Record client reports with reason `send_error` for each dropped item (so Sentry can track data loss)
- Do NOT retry on 413 — the data is definitively too large
- The 413 check comes before the existing 429 rate-limit check in the response handling flow
- Added `SizeExceededError` exception class (subclass of `ExternalError`) for internal flow control
- Added `:send_error` to the `CLIENT_REPORT_REASONS` list

Closes getsentry/sentry-ruby#2844

## References
- Python: https://github.com/getsentry/sentry-python/pull/5380
- Go: https://github.com/getsentry/sentry-go/pull/1185

## Test plan
- [x] Added tests for 413 response handling with body present
- [x] Added tests for 413 response handling with empty body
- [x] Added tests verifying client reports are recorded with `send_error` reason
- [x] Added tests verifying `send_envelope` does not re-raise the error
- [x] All existing transport tests continue to pass (65 examples, 0 failures)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sentry-ruby/lib/sentry/exceptions.rb
+++ b/sentry-ruby/lib/sentry/exceptions.rb
@@ -6,4 +6,7 @@ module Sentry
 
   class ExternalError < Error
   end
+
+  class SizeExceededError < ExternalError
+  end
 end
--- a/sentry-ruby/lib/sentry/transport.rb
+++ b/sentry-ruby/lib/sentry/transport.rb
@@ -19,7 +19,8 @@ module Sentry
       :before_send,
       :event_processor,
       :insufficient_data,
-      :backpressure
+      :backpressure,
+      :send_error
     ]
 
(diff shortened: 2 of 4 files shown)


sl0thentr0py:
reviewed and approved from me since this is a claude PR

Issue #2894, opened 2026-03-16
set_propagation_headers overwrites existing baggage header instead of merging


jakubsomonday:
### Issue Description

`Sentry::Utils::HttpTracing#set_propagation_headers` unconditionally overwrites all propagation headers on outgoing HTTP requests using `req[k] = v`. While this is fine for the Sentry-specific `sentry-trace` header, the `baggage` header is a [W3C standard](https://www.w3.org/TR/baggage/) shared across multiple systems. Any pre-existing baggage entries (e.g. set by OpenTelemetry, application code, or other instrumentation) are silently discarded.

The affected code is in [`sentry-ruby/lib/sentry/utils/http_tracing.rb#L14-L16`](https://github.com/getsentry/sentry-ruby/blob/master/sentry-ruby/lib/sentry/utils/http_tracing.rb#L14-L16):

```ruby
def set_propagation_headers(req)
  Sentry.get_trace_propagation_headers&.each { |k, v| req[k] = v }
end
```

### Reproduction Steps

1. Set a custom `baggage` header on an outgoing HTTP request (e.g. `routingKey=myvalue`)
2. Ensure `propagate_traces` is enabled in Sentry config and the target URL matches `trace_propagation_targets`
3. Make the HTTP request using `Net::HTTP`, `Faraday`, or `Excon`
4. Inspect the outgoing request headers

```ruby
require "net/http"

Sentry.init do |config|
  config.dsn = "https://key@sentry.io/1"
  config.traces_sample_rate = 1.0
  config.propagate_traces = true
end

uri = URI("https://example.com/api")
req = Net::HTTP::Get.new(uri)
req["baggage"] = "routingKey=myvalue,tenantId=123"

# After Sentry instruments this request, the baggage header
# will only contain sentry-* entries -- the original values are lost
Net::HTTP.start(uri.hostname, uri.port, use_ssl: true) { |http| http.request(req) }
```

### Expected Behavior

When a `baggage` header already exists on the outgoing request, Sentry should **merge** its entries with the existing value by joining them with a comma (`,`), as per the W3C Bagga
ge specification. The resulting header should look like:

```
sentry-trace_id=abc123,sentry-environment=production,routingKey=myvalue,tenantId=123
```

### Actual Behavior

Sentry **replaces** the entire `baggage` header with only Sentry's entries. The resulting header contains:

```
sentry-trace_id=abc123,sentry-environment=production,sentry-release=xyz,sentry-public_key=key123
```

[...]

PR #2896, opened 2026-03-17
fix: Do not overwrite baggage header contents if it already exists

jakubsomonday:
When making outgoing HTTP requests, the SDK was overwriting any pre-existing baggage header on the request with Sentry's own baggage values. This caused loss of third-party baggage entries (e.g., routingKey, tenantId) that were already set on the request.

This fix updates set_propagation_headers to merge Sentry's baggage with any existing baggage on the request instead of replacing it, following the W3C Baggage specification which allows comma-separated list entries from multiple providers.

A similar issue was previously identified and fixed in the Python SDK: https://github.com/getsentry/sentry-python/pull/2191

Changes

- Modified Sentry::Utils::HttpTracing#set_propagation_headers to check for an existing baggage header and merge values instead of overwriting
- Added test cases covering both the merge scenario and the default (no pre-existing baggage) scenario

Fixes https://github.com/getsentry/sentry-ruby/issues/2894

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Bug Fixes 🐛
 
+- Do not overwrite baggage header contents if it already exists by @jakubsomonday in [#2894](https://github.com/getsentry/sentry-ruby/pull/2894)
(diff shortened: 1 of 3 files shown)


solnic:
Thank you! I'm gonna merge this in and follow-up with additional features that will ensure we're not exceeding W3C limits with the baggage content 😄

PR #2901, opened 2026-03-18
perf: Reduce memory allocations by ~53% during exception capture

HazAT:
## Summary

SCR-20260318-ohzp-2

This PR reduces memory allocations during Sentry exception capture in Rails applications by **~53%** (442k → 206k bytes, 3305 → 1538 objects).

### Benchmark: Rails exception capture (MemoryProfiler, 5-iteration average)

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| Total allocated memory | 442,558 bytes | 205,918 bytes | **-53.5%** |
| Total allocated objects | 3,305 | 1,538 | **-53.5%** |
| Retained memory | 100,613 bytes | 49,309 bytes | **-51.0%** |
| Retained objects | 924 | 432 | **-53.2%** |

### Approach

All changes are internal optimizations — **zero behavior changes**. Same inputs produce identical outputs. All existing tests pass.

This PR is composed of 7 sub-PRs for easier review, grouped by risk level:

#### ✅ Low risk (safe to merge with quick review)
- #2902 — Avoid unnecessary allocations in hot paths (indexed captures, `match?`, `byteslice`, single-pass iteration)
- #2903 — LineCache optimization (`Hash#fetch`, direct context setting, context caching)

#### ⚡ Medium risk (caching with invalidation — review the invalidation logic)
- #2904 — Cache `longest_load_path` and `compute_filename` results (class-level, $LOAD_PATH invalidation)
- #2908 — Request interface optimizations (avoid `env.dup`, header name cache, `hint: nil`)

#### ⚠️ Needs closer look (new caching layers — review bounds and safety)
- #2905 — Backtrace line parsing and Line object caching (bounded at 2048 entries)
- #2906 — Frame object caching in StacktraceBuilder (by Line identity)
- #2907 — Scope and BreadcrumbBuffer shallow dup (dup vs deep_dup semantics)

Each sub-PR can be reviewed and merged independently into master. This base branch merges all of them together for integration testing.

PR #2903, opened 2026-03-18
perf: optimize LineCache to reduce allocations

HazAT:
## ✅ Low risk — internal refactor of LineCache

### Changes

- Use `Hash#fetch` instead of `||=` to avoid double hash lookup in `getlines`
- Remove `valid_path?`/`getline` indirection — inline bounds checking via `line_at` helper
- Build pre/post context arrays directly with `Array.new` instead of creating a single large array and slicing it
- Add `set_frame_context` method that sets frame attributes directly, avoiding the intermediate `[pre, context_line, post]` array allocation
- Cache context results per `(filename, lineno)` since the same frames repeat across exceptions — avoids recreating identical arrays

The public `get_file_context` API is preserved for custom LineCache implementations.


sl0thentr0py:
`set_frame_context` is not used so removing, plus the `context_cache` would've grown unboundedly, not a good design anyway.

PR #2906, opened 2026-03-18
perf: cache Frame objects in StacktraceBuilder by Line identity

HazAT:
## ⚠️ Needs closer review — caches Frame objects by Line object_id

### Changes

Cache Frame objects in `StacktraceBuilder#convert_parsed_line_into_frame` keyed by `Line#object_id`. When combined with Line object caching (#2905), the same cached Line instances produce the same Frame instances, eliminating ~72 duplicate Frame creations per exception.

- Cache is per-`StacktraceBuilder` instance (not class-level)
- Bounded to 2048 entries
- Frame objects are effectively read-only after context is set — attributes are only read during `to_h` serialization

### Review focus

- Keying by `object_id` — works correctly when Line objects are cached (same object = same id). For non-cached Lines, unique ids mean no false cache hits. Is this assumption documented clearly enough?
- Most effective when paired with #2905 (Line caching), but provides modest benefit independently within a single exception's backtrace.

### DIFF
--- a/sentry-ruby/lib/sentry/interfaces/stacktrace_builder.rb
+++ b/sentry-ruby/lib/sentry/interfaces/stacktrace_builder.rb
@@ -78,8 +78,16 @@ module Sentry
     private
 
     def convert_parsed_line_into_frame(line)
+      # Cache frames by Line object identity — same Line produces same Frame
+      cache_key = line.object_id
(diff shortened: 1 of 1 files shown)


sl0thentr0py:
i don't think change is safe, closing, this cache is too high level

PR #2911, opened 2026-03-24
fix: Add block syntax support to StructuredLogger for compatibility

jeromepl:
## Description

Ruby's Logger supports [passing the message as a block](https://ruby.github.io/logger/Logger.html#class-Logger-label-How+to+log+a+message). For Sentry's structured logger to be used interchangeably with Ruby's logger, it should also support this syntax.

For example, I wanted to pass `Sentry.logger` in our Rails app configured with `config.rails.structured_logging.subscribers = { ... }` to Faraday's logging middleware but that did not work because that middleware uses the block syntax:
https://github.com/lostisland/faraday/blob/8b4d1fd06fd47dd33f3720794d4df38498c240ec/lib/faraday/logging/formatter.rb#L26:L28

### DIFF
--- a/sentry-ruby/lib/sentry/structured_logger.rb
+++ b/sentry-ruby/lib/sentry/structured_logger.rb
@@ -59,8 +59,8 @@ module Sentry
     # @param attributes [Hash] Additional attributes to include with the log
     #
     # @return [LogEvent, nil] The created log event or nil if logging is disabled
-    def trace(message, parameters = [], **attributes)
-      log(__method__, message, parameters: parameters, **attributes)
+    def trace(message = nil, parameters = [], **attributes, &block)
+      log(__method__, message, parameters: parameters, **attributes, &block)
     end
 
     # Logs a message at DEBUG level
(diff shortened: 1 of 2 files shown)


afurm:
In `log()`, when both `message` and `block_given?` are true, the block is silently discarded since `message = yield` only runs when `message.nil?`. Is this the intended precedence, or should the block take priority over an already-supplied message (consistent with Ruby's Logger behavior where the block result replaces the message)?

jeromepl:
> In `log()`, when both `message` and `block_given?` are true, the block is silently discarded since `message = yield` only runs when `message.nil?`. Is this the intended precedence, or should the block take priority over an already-supplied message (consistent with Ruby's Logger behavior where the block result replaces the message)?

In Ruby's Logger, the block result replaces the message **only if the message is `nil`**. So the behaviour proposed in this PR should be consistent with the Ruby Logger one.
See the relevant Ruby Logger source here: https://github.com/ruby/logger/blob/50fedc46062381de31bb18ea8478f4df32b43ccc/lib/logger.rb#L696:L699

PR #2925, opened 2026-03-31
feat(yabeda): Add sentry-yabeda adapter gem

### LINKED ISSUE #2899: Yabeda adapter for Sentry Metrics
### Description

[Yabeda](https://github.com/yabeda-rb/yabeda) is a popular Ruby metrics framework that uses an adapter pattern to decouple metric declaration from the backend. There are existing adapters for Prometheus, Datadog, NewRelic, and StatsD :arrow_right: but currently nothing for Sentry.

`yabeda-sentry` gem that implements `Yabeda::BaseAdapter` and forwards to `Sentry.metrics`. 

Setup would look like this:

``` ruby
# Gemfile
gem "yabeda-sentry"
gem "yabeda-rails"   # optional plugins as usual
gem "yabeda-sidekiq"

Sentry.init do |config|
  config.dsn = ENV["SENTRY_DSN"]
  config.enable_metrics = true
end

# config/initializers/yabeda.rb
require "yabeda-sentry"

Yabeda.configure do
  group :myapp do
    counter   :orders_created, comment: "Orders placed",      tags: %i[region payment_method]
    gauge     :queue_depth,    comment: "Jobs waiting",       tags: %i[queue_name]
    histogram :response_time,  comment: "HTTP response time", unit: :milliseconds, tags: %i[controller action]
  end
end
```

Application code stays untouched, metrics flow to Sentry automatically.

**Mapping:**
- `counter` => `Sentry.metrics.count` (1:1)
- `gauge` => `Sentry.metrics.gauge` (1:1)
- `histogram` / `summary` => `Sentry.metrics.distribution` (bucket config dropped; Sentry aggregates server-side)

Yabeda tags map directly to `attributes:`. No pre-registration needed, so the registration hooks are no-ops.

Because `Sentry.metrics` are trace-connected, if `sentry-rails` or `sentry-sidekiq` is tracing the current request or job, metrics emitted via the adapter should automatically carry that trace context, letting you pivot from a metric spike directly into the relevant trace.

:+1: if you'd use this

dingsdax:
Sentry Ruby gem that connects [**Yabeda metrics**](https://github.com/yabeda-rb/yabeda) to [**Sentry Metrics**](https://docs.sentry.io/product/explore/metrics/).

## What it does
* Translates Yabeda metrics into Sentry:
  - `counter` ➡️ `Sentry.metrics.count`
  - `gauge` ➡️ `Sentry.metrics.gauge`
  - `histogram` / `summary` ➡️ `Sentry.metrics.distribution`
* Adds a background worker that periodically calls `Yabeda.collect!`  
  (needed because Yabeda is pull-based, while Sentry is push-based).
* Provides `start_collector!` and `stop_collector!` to control this worker (must run after `Sentry.init`).

- Histograms and summaries are both treated as distributions.
- The adapter is automatically registered when required.
- Unit handling is basic for now, with plans to improve formatting later (we don't show units in Sentry yet)

Fixes RUBY-161
Fixes #2899


spec/apps/yabeda-mini/Gemfile
sl0thentr0py, line 1: what is the purpose of this new app, we're not doing anything with it?

sentry-yabeda/spec/spec_helper.rb
sl0thentr0py, line 1: this is a new gem so it needs a new github workflow for actually testing in CI

sentry-yabeda/README.md
sl0thentr0py, line 1: we don't have per gem README's for any of the others, they just symlink back to the root with a small entry in the table

sl0thentr0py (changes requested):
some cosmetics first

sl0thentr0py (changes requested):
in the current `start_/stop_collector` model this is not very useful. Sentry generally does things automagically so the ideal design is that users just enable/add this gem and then something happens automatically, not that they have to start and stop it wherever in their code.

spec/apps/yabeda-mini/Gemfile
dingsdax: use it for manual testing, kept it in the draft PR, but will remove now

dingsdax:
@sl0thentr0py 
* CI workflow added
* hooked in `after(:configured)` instead of explicitly needed starting
* removed app for manual testing
* minimal README.md linking back to root README.md

sl0thentr0py (changes requested):
for completeness and ensuring no data loss on shutdown, we need:

* a final flush/collect during [`close`](https://github.com/getsentry/sentry-ruby/blob/8635dbc8b041e478e63b9a1ecb512ecf81bec892/sentry-ruby/lib/sentry-ruby.rb#L258)
* for this, since this is coming from another gem, we need another [callback called `:closed`](https://github.com/getsentry/sentry-ruby/blob/8635dbc8b041e478e63b9a1ecb512ecf81bec892/sentry-ruby/lib/sentry/configuration.rb#L461-L466)  registered from the gem and call `flush` there
* move the `kill` in the `:configured` callback to this new `:closed` callback

sl0thentr0py (approved):
:shipit:

dingsdax:
need to do docs still

PR #2940, opened 2026-04-24
fix(sidekiq): Report error when retry limit is below attempt_threshold

### LINKED ISSUE #2939: The `attempt_threshold` option in `sentry-sidekiq` will silently drop jobs with low retry count
### Problem Statement

The attempt_threshold (added in https://github.com/getsentry/sentry-ruby/pull/2503) has a bug when a job's retry option is lower than its attempt_threshold (e.g. retry: 1 with attempt_threshold: 3).

In those cases the error handler skipped every attempt and the job was silently dropped.

### Solution Brainstorm

https://github.com/getsentry/sentry-ruby/pull/2940

marcboquet:
## Description

Fix https://github.com/getsentry/sentry-ruby/issues/2939

The `attempt_threshold` (added in https://github.com/getsentry/sentry-ruby/pull/2503) has a bug when a job's `retry` option is lower than its `attempt_threshold` (e.g. `retry: 1` with `attempt_threshold: 3`).

In those cases the error handler skipped every attempt and the job was silently dropped.

This PR caps the threshold at the job's final attempt so these jobs still report before entering the dead set. Jobs whose retry count meets or exceeds the threshold behave exactly as before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Bug Fixes 🐛
+
+- (sidekiq) Report final attempt when retry limit is below `attempt_threshold` by @marcboquet in [#XXXX](https://github.com/getsentry/sentry-ruby/pull/XXXX)
+
 ## 6.5.0
 
 ### New Features ✨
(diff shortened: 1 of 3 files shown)


sl0thentr0py (approved):
thx @marcboquet and apologies for the bot auto-close!

PR #2966, opened 2026-06-10
ci: Weekly lockfile generation CI job

sl0thentr0py:
#### Issues

* part of: #2961
* part of: RUBY-180


dingsdax (approved):
lgtm

I wonder though if this could by DRY-ed; steps are always the same (checkout, setup, bundlce, lock, upload) only gem name, matrix and lockfile name change, extracting the gen-* steps into a reusable workflow?

or separate workflow files per gem?

sl0thentr0py:
will do later

PR #2976, opened 2026-06-16
Support Render X-Request-Start nanosecond timestamps

nateberkopec:
Render sends X-Request-Start as a whole-nanosecond epoch timestamp.


nateberkopec:
Hang on, I think I'm wrong about the premise.

PR #3017, opened 2026-07-10
fix: Fix double counting of Faraday spans when using net_http_persistent adapter

robertjlooby:
When using the `net_http_persistent` Faraday adapter (https://github.com/lostisland/faraday-net_http_persistent) with the `faraday` patch, the spans/breadcrumbs are double counted.

### DIFF
--- a/sentry-ruby/Gemfile
+++ b/sentry-ruby/Gemfile
@@ -33,6 +33,7 @@ gem "benchmark-memory"
 gem "yard"
 gem "webrick"
 gem "faraday"
+gem "faraday-net_http_persistent"
 gem "excon"
 gem "webmock"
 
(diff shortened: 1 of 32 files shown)


sentry-ruby/Gemfile
```diff
@@ -33,6 +33,7 @@ gem "benchmark-memory"
 gem "yard"
 gem "webrick"
 gem "faraday"
+gem "faraday-net_http_persistent"
```
solnic, line 36: Can we not introduce this dep and use a test adapter that would reproduce the issue instead?

solnic (commented):
Thanks for the PR! Would you mind resolving conflicts (probably it's going to be easier to just reapply commits on a freshly reset branch and f-push) and also see my question re test adapter 😄

Issue #1495, opened 2021-07-02
Review state management implementation


st0012:
Related discussions:
- https://twitter.com/_solnic_/status/1410539769495527432
- https://github.com/dry-rb/dry-effects/issues/82#issuecomment-865162374

singpolyma:
Found this because I was reviewing the code and surprised to see a thread local variable used.  I would expect each Fiber, as a separate logical context and stack, to have a different hub, but I see this was done to solve some bug related to the way Rails uses Fibers? So maybe it needs to be an option...

st0012:
@singpolyma I'd like to give each fiber its own hub as well. But as long as users still use thread as a minimum concurrency unit instead of Fiber, it'll cause issues like #1374.

As for adding a fiber-based option, I don't plan to do that right now. But I'll revisit this issue when we start the implementation for version 5.0.

sl0thentr0py:
we should revisit this when we implement the new hub-free scope system.
fwiw, opentelemtry itself uses a fiber local, if I understand correctly

https://github.com/open-telemetry/opentelemetry-ruby/blob/ff9a7d3af31ad83ed6e07445a9e1609bec53d779/api/lib/opentelemetry/context.rb#L126-L128

jakeonfire:
any movement on this? in one app we are using the fiber-based falcon webserver (https://github.com/socketry/falcon) and all our sentry metadata is uselessly scrambled.

jakeonfire:
currently using this hacky monkey-patch and it seems to be working:
```ruby
module Sentry
  class ThreadStorageOverride
    def thread_variable_get(key)
      Fiber[key]
    end

    def thread_variable_set(key, value)
      Fiber[key] = value
    end

    def method_missing(name, *args, &block)
      return super unless ::Thread.current.respond_to?(name)
      ::Thread.current.send(name, *args, &block)
    end

    def respond_to_missing?(name, include_private = false)
      ::Thread.current.respond_to?(name)
    end
  end

  class Thread < ::Thread
    def self.current
      @_current ||= ThreadStorageOverride.new
    end
  end
end
```

Issue #1374, opened 2021-03-31
ActiveRecord breadcrumbs aren't emitted inside graphql-ruby queries (Transactions lost when switching to different fibres)


bubba:
**Describe the bug**
I'm trying to integrate performance monitoring for a GraphQL API server that uses graphql-ruby on rails. Breadcrumbs for ActiveRecord queries are emitted fine up until `Schema.execute`, at which point no more breadcrumbs are logged:

```ruby
class GraphQLController < ApplicationController
  def execute
    scope = Sentry.get_current_scope
    p scope.get_transaction # returns a transaction
    result = Schema.execute(params[:query], variables: prepare_variables(params[:variables]), operation_name: params[:operation_name])
    p scope.get_transaction # returns the same transaction
    render json: result
  rescue => e
    raise e unless Rails.env.development?
    handle_error_in_development e
  end
# more stuff...
end
```

The weird thing is is that inside my GraphQL schema class, if I log the scope and transaction, the scope is different and the transaction is nil:

```ruby
# somewhere where Schema.execute calls
def foo
  p Sentry.get_current_scope # a different scope from the one inside my controller
  p Sentry.get_current_scope.get_transaction # nil!
end
```

This is why I believe the activerecord subscriber isn't emitting any breadcrumbs. And I believe this is happening because graphql-ruby does this execution on different fibres (same thread though). Thus the hub ends up getting cloned:

https://github.com/getsentry/sentry-ruby/blob/c7be1f3fb367a60246ff893eb17421cd6295180b/sentry-ruby/lib/sentry-ruby.rb#L89-L96

But once it's cloned the transaction is lost. 

**To Reproduce**

Create a new rails project, install graphql-ruby, try getting the current transaction from inside your schema/types. I think it might be possible to make a smaller test case by doing something with fibres yourself.

**Expected behavior**

`get_current_scope` and `get_current_transaction` return the same scope and transaction as the originating fibre/inside the action controller, and any activerecord queries are added as breadcrumbs.

**Actual behavior**
`get_current_scope` returns a different scope and `get_current_transaction` returns nil, and no activerecord queries are added to sentry as breadcrumbs

**Environment**
 - Ruby Version: 3
 - SDK Version:
   - sentry-ruby 4.3.1
   - sentry-rails 4.3.3
 - Integration Versions (if any):
  - Rails 6.1.3.1
  - graphql-ruby 1.12.6
  - 

**Raven Config**

This is not necessary but could be helpful.

[...]

PR #3018, opened 2026-07-11
feat: add config.hub_isolation_level for fiber-safe hub storage (Falcon/async)

SeanLF:
## What

Adds an opt-in `config.hub_isolation_level = :thread | :fiber` (default `:thread`) that controls where the SDK stores the current `Hub`.

- `:thread` (default) — thread-local storage, **exactly today's behaviour**. Nothing changes for anyone who doesn't opt in.
- `:fiber` — Ruby 3.2+ Fiber Storage (`Fiber[]`). Each fiber gets its own hub, and child fibers inherit it, so concurrent requests on a fiber-based server (Falcon, `async`) are isolated instead of sharing and corrupting one another's scope.

This is the fiber-based option discussed and deferred in #1495.

## Why

On a fiber server, concurrent requests run as sibling **fibers on one thread**, so they share the single thread-local hub. When a request holds a scope across a reactor yield (streaming, IO), a concurrent request's `with_scope` mutates the same scope stack, and scope/breadcrumbs/**user** leak between requests.

Reproduced against a real `Async::HTTP::Server` + the real `Sentry::Rack::CaptureExceptions` middleware, firing concurrent HTTP requests that each set a user, yield mid-scope, then capture an event:

```
[:thread] 6 error events | 5 attributed to the WRONG user
[:fiber]  6 error events | 0 attributed to the WRONG user
```

Under thread-local, 5 of 6 events were reported under the wrong user (cross-user PII bleed). `:fiber` fixes it.

## Why this is safe now, when #1380 moved the other way

#1380 (2021) moved **off Ruby's old fiber-local variables** (`Thread.current[]`) because they don't inherit into child fibers, which lost graphql-ruby's context (#1374). Ruby 3.2 added **Fiber Storage** ([`Fiber#storage`](https://bugs.ruby-lang.org/issues/19078)) — a different primitive that **copies the parent's storage into a newly created fiber**. It gives per-request isolation **and** child-fiber inheritance at once, so it fixes fiber servers without re-breaking #1374. The added regression tests prove both hold.

This is also how the other SDKs already work (sentry-python `contextvars`, sentry-javascript `AsyncLocalStorage` — both isolate per task and inherit into children), and it follows opentelemetry-ruby's 2025 move off thread-local storage (#1807).

## Design

- New `Sentry::HubStorage` owns get/set/clear of the current hub and switches on the isolation level. `THREAD_LOCAL = :sentry_hub` is kept as the storage key (user monkey-patches reference the constant).
- `config.isolation_level` is applied at `Sentry.init`. Requesting `:fiber` on Ruby < 3.2 logs a warning and falls back to `:thread` (feature-detected, so nothing breaks on 2.7–3.1).
- Every former `thread_variable_*` site now routes through `HubStorage` (core `sentry-ruby.rb`, plus `sentry-rails` `active_job.rb` and core `test_helper.rb`, which touched `THREAD_LOCAL` directly and would be wrong under `:fiber`).

Mirrors Rails' `config.active_support.isolation_level` ergonomics; uses `Fiber[]` (not Rails' non-inheriting `Fiber.attr_accessor`) because the SDK needs the inheritance property.

```ruby
Sentry.init do |config|
  config.dsn = "..."
  config.hub_isolation_level = :fiber # opt in on Falcon / async
end
```

## Compatibility

- Default `:thread` is byte-for-byte the previous behaviour. Puma/Sidekiq/Unicorn/Resque are unaffected.
- Ruby `>= 2.7` still supported; `:fiber` requires 3.2+ and degrades to `:thread` with a warning below that.
- `:fiber` is safe on threaded servers too (each request re-clones the hub); the difference only matters when child fibers are spawned per request.

## Caveat (documented)

Under `:fiber`, a raw `Thread.new` spawned mid-request inherits the request's hub object (documented Ruby fiber-storage thread inheritance). Call `Sentry.clone_hub_to_current_thread` in the spawned thread to isolate it — same as today.

## Tests

- New `spec/sentry/hub_storage_spec.rb`, `config.isolation_level` specs, and a fiber-isolation regression in `spec/sentry_spec.rb` (sibling fibers keep isolated user attribution across a yield; child fiber inherits the parent hub). `:fiber` specs guard on Fiber-storage availability.
- Full core suite green; the #1374 regression test passes under both levels.

## Performance

benchmark-ips (Ruby 4.0.5): `Fiber[]` get is **1.24x faster** than `thread_variable_get`; at the level of a full `Sentry.add_breadcrumb` the two isolation levels are within measurement error. Not a hot-path regression.

## Notes for reviewers

- Two commits: core feature, then a 2-line `sentry-rails` follow-on so ActiveJob is correct under `:fiber`. Happy to split the rails commit into a separate PR if you'd prefer to keep this core-only.
- No `CHANGELOG.md` edit (per the repo's auto-generated-at-release convention). Suggested entry:
  > - Add `config.isolation_level = :thread | :fiber` for fiber-safe hub storage on fiber-based servers (Falcon/async), fixing cross-request scope contamination. Default `:thread` keeps existing behaviour.

Refs #1495. /cc @sl0thentr0py @solnic

### Changelog Entry

Add Fiber support new `config.hub_isolation_level` which can be set to `:fiber` or `:thread`. Use this in case you use Fiber based servers like [Falcon](https://github.com/socketry/falcon)


SeanLF:
Both findings addressed in 3ce928c:

1. **Post-init isolation sync (medium):** `Configuration#isolation_level=` now applies the level to `HubStorage` when the SDK is already initialized, so a post-`init` change to `Sentry.configuration.isolation_level` actually takes effect instead of leaving the SDK on the old backend. The `initialize` default is assigned to the ivar directly so a throwaway `Configuration.new` can't clobber the active isolation level (only `Sentry.init` builds the live config).

2. **Fiber specs vs Ruby < 3.2 fallback (high):** the `:fiber`-asserting specs now stub `fiber_storage_available?` so they're deterministic across the Ruby matrix, plus an explicit downgrade-to-`:thread` spec. They previously asserted `:fiber` unconditionally and would have failed on the < 3.2 cells, where `:fiber` correctly falls back to `:thread`. Added a regression test for the post-init sync too.

sl0thentr0py:
thanks a lot @SeanLF, if you don't mind, I will reduce the code a bit and make it more minimal.

sl0thentr0py (approved):
* renamed config to `hub_isolation_level` for more clarity
* moved logic all inline to `Configuration` instead of `HubStorage`

sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -610,6 +628,23 @@ def release=(value)
...
+      unless ISOLATION_LEVELS.include?(level)
+        raise ArgumentError, "hub_isolation_level must be one of #{ISOLATION_LEVELS.inspect}, got #{level.inspect}"
+      end
+
+      # :fiber relies on Fiber Storage (Ruby 3.2+); downgrade so the hub is never
+      # asked to call Fiber[] on a Ruby that lacks it.
+      if level == :fiber && !fiber_storage_available?
+        log_warn("hub_isolation_level :fiber requires Ruby 3.2+ Fiber Storage; falling back to :thread on Ruby #{RUBY_VERSION}.")
+        level = :thread
+      end
```
solnic, line 643: We could completely avoid that and define writer method dynamically depending on the availability, which would simplify runtime production code path.

sentry-rails/lib/sentry/rails/active_job.rb
```diff
@@ -117,13 +117,7 @@ def log_producer_span_error(e)
           end
 
           def record(job, trace_headers: nil, user: nil, &block)
-            # Always give this thread a fresh hub cloned from the main hub so
-            # the job's events are fully isolated.  Save and restore whatever
-            # hub was on the thread before (e.g. the Rack request hub set by
-            # CaptureExceptions, or a stale hub left by a recycled thread-pool
-            # thread) so the outer context continues working correctly after
-            # the job finishes.
-            original_hub = Thread.current.thread_variable_get(Sentry::THREAD_LOCAL)
```
solnic, line 126: Was this comment inaccurate and that's why it got removed?

sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -610,6 +628,23 @@ def release=(value)
       @release = value
     end
 
+    def hub_isolation_level=(level)
+      level = level.to_sym if level.respond_to?(:to_sym)
```
solnic, line 632: What's the reason for allowing non-symbols over here?

sentry-ruby/lib/sentry/configuration.rb
```diff
@@ -389,6 +389,23 @@ class Configuration
...
+    #   own hub and child fibers inherit it, so concurrent requests on a
+    #   fiber-based server (Falcon/async) are isolated instead of sharing and
+    #   corrupting one another's scope. Requested on a Ruby without Fiber
+    #   Storage (< 3.2), the SDK logs a warning and falls back to +:thread+.
+    #
+    # @return [Symbol]
+    attr_reader :hub_isolation_level
+
+    # Isolation levels the SDK understands for hub storage.
+    ISOLATION_LEVELS = %i[thread fiber].freeze
```
solnic, line 407: Nit and could be a follow-up mini-refactor, but: we have a `validate` macro - we could extend it to support `included_in: [...]` and then overriding the writer would only be needed for the situation where the Fiber storage API is not available.

solnic (commented):
Thank you for working on this 🙏🏻 Left some nits for your consideration 😄

sentry-rails/lib/sentry/rails/active_job.rb
sl0thentr0py: we should generally not have this clanker generated comment bloat, so I will remove it wherever I see it unless it really adds any value.

sentry-ruby/lib/sentry/configuration.rb
sl0thentr0py: just syntax sugar

sentry-ruby/lib/sentry/configuration.rb
sl0thentr0py: we still want to have the warning message if they try to set it, so this needs to stay

trevorturk:
Hello! I ran into this PR because I was having an issue in an app I'm running with Falcon that was leaking Sentry tags across requests where it should have been impossible. I ran this all through Claude Fable, so please accept my apologies, I'll try to keep the AI slop to a minimum, but I think we might have room to improve. I think the tl;dr is a bit strange because of my currently-mixed workload where I'm running Falcon for web, but SolidQueue for jobs, and SolidQueue doesn't ship with an Async/Fiber mode yet (but hopefully will soon: https://github.com/rails/solid_queue/pull/728)

So... Claude slop follows, edited a bit to keep it short, I hope this is helpful, but please accept my apologies for not entirely working through this myself in detail -- this should be taken with a grain of salt, but I thought it worth sharing:

---

A caveat worth considering before this ships: Fiber storage is inherited by `Thread.new`, sharing one hub across threads. 

While evaluating the `:fiber` approach (we prototyped equivalent Fiber-storage hub accessors on 6.6.2 while waiting for the release), we hit a caveat worth raising before `hub_isolation_level` ships: **Ruby's fiber storage is inherited by `Thread.new`, not just `Fiber.new`, and the inherited values are the same objects** — so a hub stored via `Fiber[]` in one fiber is *shared* (not copied) into any thread that fiber spawns.

## Why this bites in practice

In a typical Rails app, hubs get seeded early and implicitly — e.g. `Sentry::Rails::Breadcrumb::ActiveSupportLogger` calls `Sentry.add_breadcrumb → get_current_hub` on every ActiveSupport notification, so the first SQL query in a process seeds the current fiber's hub. Any thread pool created after that point (a job processor's workers, an app's own worker threads) inherits that one hub into every thread. In `:fiber` mode:

- Concurrent jobs in a threaded processor (we reproduced the shape with a Solid Queue-style pool) all read and write **one shared scope**: 4 workers → 1 hub, each seeing another job's tags from inside its own `Sentry.with_scope`.
- The shared `Hub#@stack` Array is then pushed/popped from multiple OS threads without synchronization.

So a mixed web+jobs app that enables `:fiber` globally (the natural reading of a config option) trades cross-*request* contamination on the web side for cross-*job* contamination on the worker side.

## Repro sketch (Ruby 3.4, semantics-level)

```ruby
Fiber[:hub] = Object.new
threads = Array.new(4) { Thread.new { Fiber[:hub] } }
threads.map(&:value).map(&:object_id).uniq.size # => 1 — all threads share the fiber's object
```

With hub accessors reading `Fiber[KEY] || clone`, the `|| clone` fallback never fires in those threads — they inherit a non-nil hub — so the per-context isolation the fallback was designed to provide is silently bypassed.

## Possible mitigation

Tagging the stored hub with its owning thread restores per-thread cloning at thread boundaries while keeping fiber inheritance within a thread:

```ruby
def get_current_hub
  thread, hub = Fiber[KEY]
  thread.equal?(Thread.current) ? hub : clone_hub_to_current_thread
end

def clone_hub_to_current_thread
  hub = get_main_hub.clone
  Fiber[KEY] = [Thread.current, hub]
  hub
end
```

SeanLF:
@trevorturk thanks for digging into this! My own claude managed to reproduce it on master

| | expected | actual |
|---|---|---|
| 4 worker threads | 4 hubs | 1 hub |
| 4 concurrent `with_scope` jobs | own tags | each read another job's tags |
| 3 sibling request fibers | 3 hubs | 1 hub |

`Fiber.new` inherits storage the same way `Thread.new` does, so `:fiber` never fixed the Falcon case it was built for either. The shared `@stack` is also pushed and popped from several threads with no synchronization.

The existing regression spec missed it because it calls `Sentry.clone_hub_to_current_thread` inside each fiber, which overwrites the inherited entry. Real request fibers never call that.

Two notes on your mitigation. Tagging with `Thread.current` fixes the worker-thread case but leaves sibling fibers sharing one hub, since they run on the same thread; tagging with the owning fiber covers both. And `clone_hub_to_current_thread` returns whatever the setter returns, so `Fiber[k] = [thread, hub]` hands callers the Array.

Cloning the hub is also not enough on its own. `Hub#clone` deep-dups the scope's span, which builds a fresh `SpanRecorder`, so child-fiber spans land on a detached transaction copy and are never sent. The fork has to keep the live span.

PR incoming. This is unreleased, so it can land before the next release

trevorturk:
Excellent! Thank you for digging into that. Please link the PR here so we can keep track and I'll QA IRL when everything is settled.

SeanLF:
PR is up: #3036

It tags each stored entry with its owning fiber, so a context that inherited one takes its own copy. The copy keeps the live span, otherwise child spans land on a detached transaction copy and are never sent.

One thing that turned up while testing across the matrix, relevant to your Solid Queue setup: the `fiber-storage` gem polyfills `Fiber[]` on Ruby < 3.2 and on JRuby, so `:fiber` is live there too. That polyfill inherits into child fibers but not into `Thread.new`, so on those Rubies only the sibling-fiber half of the bug exists; worker threads already start from the main hub. On 3.2+ with native fiber storage you get both halves, which matches what you saw.

Happy to have it QA'd IRL once it lands.

sl0thentr0py:
> The existing regression spec missed it because it calls Sentry.clone_hub_to_current_thread inside each fiber, which overwrites the inherited entry. Real request fibers never call that.

So there are some architectural concerns missing in this conversation. How the Hub is currently designed is that the user of a concurrency boundary is **supposed** to clone the hub via `clone_hub_to_current_thread`. I know there are some holes in this implementation but that's how we've done it in practice because it's hard to generalize over all thread/fiber/goroutine usage patterns and how they want their concurrency to behave and map over to Sentry telemetry.

For instance, in `Rack`, we [explicitly isolate every request/response lifecycle](https://github.com/getsentry/sentry-ruby/blob/395aca9a7f5527ed7846cf8bc3b5a3b7d637b8ed/sentry-ruby/lib/sentry/rack/capture_exceptions.rb#L18) since across HTTP server implementations, some reuse a thread while others don't, so isolating automatically on the thread level is not sufficient.

> Any thread pool created after that point (a job processor's workers, an app's own worker threads) inherits that one hub into every thread.

In this case, I would rather isolate on the worker / task level by again explicitly calling `clone_hub_to_current_thread` like in [`Sidekiq`](https://github.com/getsentry/sentry-ruby/blob/395aca9a7f5527ed7846cf8bc3b5a3b7d637b8ed/sentry-sidekiq/lib/sentry/sidekiq/sentry_context_middleware.rb#L43).

The base feature configured by `hub_isolation_level` should be mainly about where the hub is stored, forking should be handed off to specific integrations or monkeypatches.

sl0thentr0py:
@solnic just to confirm, with the active job changes, [this isolation](https://github.com/getsentry/sentry-ruby/blob/395aca9a7f5527ed7846cf8bc3b5a3b7d637b8ed/sentry-rails/lib/sentry/rails/active_job.rb#L132) will also work for `SolidQueue` right?

SeanLF:
Correction to what I said above about `fiber-storage`: it isn't a sentry-ruby dependency, it arrives transitively (graphql-ruby depends on it). [`fiber_storage_available?`](https://github.com/getsentry/sentry-ruby/blob/e89d3bc0bed366f7201af6eca30fecef861be844/sentry-ruby/lib/sentry/configuration.rb#L845-L847) is `Fiber.respond_to?(:[])`, which can't tell the polyfill from native storage, and they differ: the polyfill doesn't inherit into `Thread.new`. So on Ruby 3.1, an app that sets `:fiber` and happens to have graphql-ruby in its bundle gets it, with no [warning or fallback](https://github.com/getsentry/sentry-ruby/blob/e89d3bc0bed366f7201af6eca30fecef861be844/sentry-ruby/lib/sentry/configuration.rb#L640-L644). The semantics it gets are the polyfill's, not native storage's.

While testing that, I mapped what each level does when you spawn a thread or fiber mid-request:

| | `:thread` | `:fiber` |
| --- | --- | --- |
| `Thread.new` | own scope, starts from main | own scope, copied from caller |
| `Fiber.new` | shares the caller's scope | own scope, copied from caller |

[`clone_hub_to_current_thread`](https://github.com/getsentry/sentry-ruby/blob/e89d3bc0bed366f7201af6eca30fecef861be844/sentry-ruby/lib/sentry-ruby.rb#L389-L392) is always `get_main_hub.clone`, so it resets to main rather than isolating the current context. That is what you want at a request or job boundary. Inside one it usually isn't: under `:fiber` it discards the user and tags the spawned context would otherwise inherit. Under `:thread`, calling it inside a fiber also swaps the hub out from under the calling thread, so the caller loses its own scope data. That last one predates #3036.

sentry-docs doesn't mention the method at all. Want a PR?

<details>
<summary>Repro</summary>

```ruby
# spawn_probe.rb
require "sentry-ruby"

Sentry.init do |c|
  c.dsn = "http://foo@example.com/1"
  c.hub_isolation_level = ARGV[0].to_sym
  c.background_worker_threads = 0
  c.sdk_logger = Logger.new(IO::NULL)
end

Sentry.get_current_scope.set_user(id: "req-user")
seen = -> { Sentry.get_current_scope.user[:id].inspect }

puts "level=#{Sentry.instance_variable_get(:@hub_isolation_level).inspect}"
puts "  Thread.new sees: #{Thread.new { seen.call }.value}"
puts "  Fiber.new  sees: #{Fiber.new { seen.call }.resume}"
Fiber.new { Sentry.clone_hub_to_current_thread }.resume
puts "  caller after a fiber cloned: #{seen.call}"
```

```console
$ ruby -Ilib spawn_probe.rb thread
level=:thread
  Thread.new sees: nil
  Fiber.new  sees: "req-user"
  caller after a fiber cloned: nil

$ ruby -Ilib spawn_probe.rb fiber
level=:fiber
  Thread.new sees: "req-user"
  Fiber.new  sees: "req-user"
  caller after a fiber cloned: "req-user"
```

`Fiber.new` reads the same under both levels; the last line is what separates them.

```console
$ docker run --rm ruby:3.1-slim sh -c 'gem install fiber-storage -N -q && ruby -rfiber/storage -e "
  o = Object.new; Fiber[:h] = o
  puts %{child thread: #{Thread.new { Fiber[:h].equal?(o) }.value}}"'
child thread: false     # true on ruby:3.2-slim with native storage
```
</details>

sl0thentr0py:
> sentry-docs doesn't mention the method at all. Want a PR?

we consider this api internal in the sense that integration authors need to know about it but not SDK users, so I'd rather not expose that.

solnic:
> @solnic just to confirm, with the active job changes, [this isolation](https://github.com/getsentry/sentry-ruby/blob/395aca9a7f5527ed7846cf8bc3b5a3b7d637b8ed/sentry-rails/lib/sentry/rails/active_job.rb#L132) will also work for `SolidQueue` right?

@sl0thentr0py correct - `SolidQueue` has zero customization right now, it's based 100% on the Active Job integration.

PR #3077, opened 2026-09-09
feat: Gracefully handle SDK callback exceptions

### LINKED ISSUE #3072: Wrap all SDK user callbacks in a try catch
| Callback | On exception |
| -- | -- |
| `before_send` | Logs internal error, drops event |
| `before_send_transaction` | Logs internal error, drops event |
| `before_send_log` | Logs internal error, drops log |
| `before_send_metric` | Logs internal error, drops metric |
| `before_send_span` | Logs internal error, emits span |
| `before_breadcrumb` | Logs internal error, drops breadcrumb |
| `error_sampler` | Logs internal error, falls back to sample_rate |
| `traces_sampler` | Logs internal error, falls back to traces_sample_rate |
| `profiles_sampler` | Logs internal error, falls back to profiles_sample_rate |
| `event_processor` | Logs internal error, drops event |
| *and potentially more* 🙈 |  |

sl0thentr0py:
### Changelog Entry

The SDK now wraps user callbacks so that they don't unintentionally break your application if they raise inside the callback.

- `before_send`: return nil and drop the event
- `before_send_transaction`: return nil and drop the transaction
- `before_send_check_in`: return nil and drop the transaction
- `before_send_check_in`: return nil and drop the check-in
- `event_processor`: return nil and drop the event
- `before_send_log`: return nil and drop the log item
- `before_send_metric`: return nil and drop the metric item
- `before_breadcrumb`: return nil and drop the breadcrumb
- `traces_sampler`: fall back to `traces_sample_rate`
- `backtrace_cleanup_callback`: fall back to the original backtrace
- `std_lib_logger_filter`: return nil and skip the log

#### Issues

* Resolves: #3072
* Resolves: RUBY-211


solnic (approved):
LGTM

PR #3084, opened 2026-09-14
fix(metrics, logs): Ensure TelemetryBuffer uses own thread for both size based and periodic flushing

sl0thentr0py:
Before this change, the `add_item` triggered size based flush on the caller thread which is absolutely incorrect behavior.

Changes to `ThreadedPeriodicWorker`
* a dedicated `@thread_mutex` for syncing its own internal state
* a condition variable `@wake_condition` that is used for `wake`ing the thread from outside (like size based flushing) - otherwise defaults to the `interval` which makes it backwards compatible
* another condition variable `@idle_condition` used for the final `flush` that needs to wait for any running task to finish
* two new methods `wake` and `wait_for_idle`

These are then used in `TelemetryEventBuffer`:
* `run` and `flush` are now two different methods, `run` runs on the buffer thread whereas `flush` on the main one


sentry-ruby/lib/sentry/threaded_periodic_worker.rb
sl0thentr0py, line 1: reviewing this with whitespace off recommended

sentry-ruby/spec/isolated/std_lib_logger_spec.rb
```diff
@@ -224,5 +228,23 @@
         end
       end
     end
+
+    it "does not conflict with the Sentry logger patch" do
```
sl0thentr0py, line 232: moved this here to avoid logger patch leaking into the other test

solnic (commented):
I honestly think that at this point we'd be better off using `concurrent-ruby`'s primitives for this stuff, we already depend on that gem anyway.

sentry-ruby/lib/sentry/threaded_periodic_worker.rb
```diff
@@ -4,41 +4,82 @@ module Sentry
...
+        @condition.signal
+      end
+
+      true
+    end
+
+    def wait_until_idle
+      @thread_mutex.synchronize do
+        @idle_condition.wait(@thread_mutex) while @running
+      end
```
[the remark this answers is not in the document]
sl0thentr0py: This is not a valid issue under normal Ruby thread semantics.

 Thread#kill raises a termination exception in the target thread, and Ruby still executes the thread’s ensure
 blocks. Therefore, if run is interrupted, this cleanup still runs:

 ```ruby
   ensure
     @thread_mutex.synchronize do
       @running = false
       @idle_condition.broadcast
     end
   end
 ```

 Also, kill releases @thread_mutex before calling thread.kill, so the worker cannot be permanently blocked
 trying to enter the ensure block because of the caller.

 Additionally, Sentry.close does not kill the telemetry event-buffer workers. It kills the session flusher and
 backpressure monitor, then calls client.flush on the telemetry buffers. The telemetry worker remains alive,
 so the described “kill this worker, then flush this same worker” sequence does not occur during normal
 shutdown.

 wait_until_idle is also guarded by the @running predicate:

 ```ruby
   @idle_condition.wait(@thread_mutex) while @running
 ```

 So a broadcast is only used to wake it for a re-check; it is not relying on a one-time signal.

 A hang would require an independent failure inside the ensure block or abnormal process termination. No fix
 is needed for the reported scenario.

sentry-ruby/spec/support/shared_examples_for_telemetry_event_buffers.rb
```diff
@@ -95,7 +100,7 @@
     let(:max_items) { 30 }
 
     it "thread-safely handles concurrent access" do
-      expect(client).to receive(:send_envelope).exactly(3).times
+      expect(client).to receive(:send_envelope).at_least(:once)
```
solnic, line 103: @sl0thentr0py I assume that was needed due to jruby, amrite?

sl0thentr0py:
ran sol pro on this for deadlock analysis
https://gist.github.com/sl0thentr0py/5be6b2ff7adee8508d86f5315d3e142e

sentry-ruby/lib/sentry/threaded_periodic_worker.rb
```diff
@@ -4,41 +4,84 @@ module Sentry
...
     rescue ThreadError
+      @thread_mutex.synchronize { @exited = true }
       log_debug("[#{self.class.name}] thread creation failed")
-      @exited = true
       false
     end
 
+    def wake
+      @thread_mutex.synchronize do
+        @force_run = true
```
solnic: @sl0thentr0py what does this `force_run` mean? does it cause it to flush immediately without waiting? (just trying to understand the mechanics here).

sentry-ruby/spec/support/shared_examples_for_telemetry_event_buffers.rb
sl0thentr0py: no, not just jruby, it does not always trigger 3 times deterministically now, just whatever `wake` picks up as pending.

sentry-ruby/lib/sentry/threaded_periodic_worker.rb
sl0thentr0py: yes exactly, it either runs on `timeout` or when `wake_condition` triggers, this is an extra boolean that goes together with the `wake_condition` signal. Can rename to make it clearer, like `is_woken`.

sentry-ruby/lib/sentry/telemetry_event_buffer.rb
```diff
@@ -41,30 +41,23 @@ def initialize(configuration, client, event_class:, max_items:, max_items_before
...
-      pending_items = @mutex.synchronize do
-        next if @pending_items.empty?
-
-        items = @pending_items
-        @pending_items = []
-        items
-      end
-
-      return unless pending_items
+      wait_until_idle
```
adinauer, line 44: Ran it through my review team and found this:


#### F-001: Reentrant telemetry flush deadlocks telemetry workers and shutdown

- **Location:** `sentry-ruby/lib/sentry/telemetry_event_buffer.rb:43`; `sentry-ruby/lib/sentry/threaded_periodic_worker.rb:64`; interaction with `sentry-ruby/lib/sentry/client.rb:130-134`
- **Reporters:** Correctness, Guardian, Guardian: Security, Android, Backend, Competition, Customer, Ecosystem
- **Description:** A threshold-triggered send now executes on a telemetry worker with `@running = true`. `send_items` invokes the supported `before_send_log` or `before_send_metric` callback. If that callback calls `Sentry.get_current_client.flush` or `Sentry.close`, `Client#flush` reaches the same buffer's new `wait_until_idle`. The worker waits for itself to finish, but `@running` is cleared only after the callback returns, so neither operation can complete.
- **Impact:** The telemetry worker becomes permanently blocked. Later external flushes and normal SDK/process shutdown can also hang. Backend and Guardian: Security independently reproduced the actual telemetry callback path against the PR implementation, including a subsequent external flush timing out. Before this PR, the outer send had already swapped out its pending batch and a nested flush returned without waiting.
- **Agent perspectives:**
  - **Correctness/Backend:** Confirmed the public callback-to-flush route and P1 severity.
  - **Guardian/Security:** Confirmed this is an availability regression, not a confidentiality or supply-chain issue.
  - **Customer/Ecosystem:** Confirmed before-send callbacks and flush/close are supported customer-facing routes.
  - **Android:** No Java/Android code is affected, but the deadlock violates standard SDK flush/lifecycle expectations.
- **Suggested fix:** Make client-level telemetry flushing reentrancy-aware. When `Client#flush` runs on either the log or metric telemetry worker, it must not wait on either telemetry buffer. A guard only in the current buffer is insufficient: if both workers concurrently execute callbacks that call `Client#flush`, the log worker can wait on the metric worker while the metric worker waits on the log worker. Preserve any safe non-buffer transport flushing as appropriate.
- **Required regression coverage:**
  1. A threshold-triggered log or metric before-send callback calls `client.flush`; the callback completes and a later external flush also completes.
  2. Log and metric workers concurrently invoke callbacks that call `client.flush`; neither worker cross-waits.

adinauer: Apologies, I only just now read through this finding and this seems irrelevant since we don't expect anyone to flush from within the `beforeSend` callbacks. Ignore this.

solnic (approved):

